PR #2, opened 2015-02-20
Change 'utils' library to 'http.utils'

kaendfinger:
It is important to do this because using a generic 'utils' library can conflict with other packages and even the user of the library.


munificent:
Good catch! Can you also change the version number in the pubspec to `0.11.2-dev`? That ensures we know there are unpublished changes in the repo.

kaendfinger:
Sure

nex3:
Oops! I had made the pubspec change but forgotten to push it.

Issue #5, opened 2015-03-08
Add utility methods for Basic Auth, Digest Auth, etc.


wstrange:
dart:io HttpClient has utility methods for adding client credentials to the request:

https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:io.HttpClientCredentials

It would be convenient to have similar methods. Not sure where it fits in this library's API.

donny-dont:
This will be done through middleware going forward. Can close.

PR #7, opened 2015-04-03
add withCredentials to base methods

johnryan-wf:
This adds withCredentials as an optional parameter for BaseClient requests.  When not using `BrowserClient`, this flag is ignored.  Defaults to false.

@nex3 Please let me know if you would prefer a different approach

### DIFF
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -44,6 +44,7 @@ class BrowserClient extends BaseClient {
       _xhrs.add(xhr);
       xhr.open(request.method, request.url.toString(), async: true);
       xhr.responseType = 'blob';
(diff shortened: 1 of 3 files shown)


johnryan-wf:
this addresses https://github.com/dart-lang/http/issues/5 I think

nex3:
I don't want to add anything platform-specific to the shared `Request` class. Ultimately, my goal is to move `http` to a more `shelf`-like API with a `context` getter, which would work well for this, but that's a ways away. For now, it's probably best to put this property on `BrowserClient`.

johnryan-wf:
@nex3 I agree that's a cleaner solution.  Done.

PR #34, opened 2016-01-13
Update from package:unittest to package:test

jirkadanek:
After doing this, all tests pass except the ones that rely on the /echo server. I am unsure as to what it was and where I can get it, so I made a new echo server in a separate pull request. With that pull requests, all tests pass.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,6 +8,6 @@ dependencies:
   path: ">=0.9.0 <2.0.0"
   stack_trace: ">=0.9.1 <2.0.0"
 dev_dependencies:
-  unittest: ">=0.9.0 <0.12.0"
+  test: ">=0.12.0 <0.13.0"
 environment:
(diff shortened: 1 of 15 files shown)


kevmoo:
@jirkadanek I was wondering about this, too. Talked to @nex3 

Turns out this is blocked by a feature in pkg/test – https://github.com/dart-lang/test/issues/330 – so we can run browser tests w/ a custom server.

donny-dont:
@nex3 @kevmoo after I land the updated browser client I plan on doing this which will obsolete this pull request.

One issue is that is going to crop up is that `test` depends on `http` due to [package_resolver](https://github.com/dart-lang/package_resolver/blob/master/pubspec.yaml#L12). Might need to make a branch on it that would accept `0.12.0-dev` to get any results.

donny-dont:
@nex3 i believe this is obsoleted and we can close it out. Got some more tests incoming for you.

PR #35, opened 2016-01-13
Add an echo server for testing

jirkadanek:
This extends my previous pull requests with an echo server. AFAIK package:test testrunner does not contain an echo server and this is my proposal what to do about it.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,6 +8,6 @@ dependencies:
   path: ">=0.9.0 <2.0.0"
   stack_trace: ">=0.9.1 <2.0.0"
 dev_dependencies:
-  unittest: ">=0.9.0 <0.12.0"
+  test: ">=0.12.0 <0.13.0"
 environment:
   sdk: ">=1.9.0 <2.0.0"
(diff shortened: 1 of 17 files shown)


nex3:
I'm going to review these in reverse order, because the echo server is the main thing blocking the move to the `test` package.

The tricky thing here is that we need the echo server *in HTML tests*, and those tests (being confined to the browser) have no way of spinning up such a server. As far as I can tell, the only way to solve this will be to write a plugin for the test package that starts such a server, but that's blocked on dart-lang/test#109. If you can come up with a way around that restriction, though, I'm all ears.

jirkadanek:
What I am doing is starting the echo server manually before running `pub test`. I am setting a CORS header so that the website can access it, even though it is on a different origin (different port).

I am not sure if running two commands instead of one is a problem. It should be possible to do a script to automate that, if you think it is a problem. Regarding the CORS thing, the main problem is IMO to decide on a port number that is guaranteed to be unoccupied. I picked 395 with a random number generator and added 8 to it to make it > 1024.

nex3:
We really need the tests to be able to run automatically on the bots. I think it's probably best to just leave this un-upgraded until we can add an echo server properly.

PR #37, opened 2016-06-16
Fix http.testing doc example

chalin:
Fixed:
- Syntax errors (two closing parentheses were missing)
- Return type of function literal

The sample code was formatted with `dartfmt`.

cc @kwalrath

### DIFF
--- a/lib/testing.dart
+++ b/lib/testing.dart
@@ -11,14 +11,15 @@
 ///     import 'dart:convert';
 ///     import 'package:http/testing.dart';
 ///
-///     var client = new MockClient((request) {
+///     var client = new MockClient((request) async {
 ///       if (request.url.path != "/data.json") {
 ///         return new Response("", 404);
 ///       }
-///       return new Response(JSON.encode({
-///         'numbers': [1, 4, 15, 19, 214]
-///       }, 200, headers: {
-///         'content-type': 'application/json'
-///       });
-///     };
+///       return new Response(
+///           JSON.encode({
+///             'numbers': [1, 4, 15, 19, 214]
+///           }),
+///           200,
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks for fixing this, Patrice! I'll leave the review to @nex3...

lib/testing.dart
```diff
@@ -11,14 +11,15 @@
 ///     import 'dart:convert';
 ///     import 'package:http/testing.dart';
 ///
-///     var client = new MockClient((request) {
+///     var client = new MockClient((request) async {
```
munificent: I don't think it needs to be `async`, but @nex3 would know better.

chalin: It doesn't type check otherwise. The expected type of the constructor argument is
```dart
typedef Future<Response> MockClientHandler(Request request);
```

lib/testing.dart
```diff
@@ -11,14 +11,15 @@
...
-///       }, 200, headers: {
-///         'content-type': 'application/json'
-///       });
-///     };
+///       return new Response(
+///           JSON.encode({
+///             'numbers': [1, 4, 15, 19, 214]
+///           }),
+///           200,
+///           headers: {'content-type': 'application/json'});
```
nex3: I think this new formatting reads less well than the original, `dartfmt` notwithstanding.

munificent: For what it's worth, I actually like the new output. I think the 200 gets a little lost in the before code.

chalin: So ... is this ready to be merged?

(IMHO, the raison d'etre of source code formatters is that they relieve us, as designers, from the tedium and trivialities surrounding code layout :)

Issue #47, opened 2016-12-14
Request set bodyFields, need to be able to have List<String> for a map value


vsmenon:
_From @rwrozelle on December 14, 2016 14:5_

I need to be able to set bodyFields with a map that has multiple values for a single key, so map would be Map<String, Object> where Object would be either a String or List<String> at runtime.  Here is an example map:  {q: *:*, start: 0, rows: 0, facet: true, stats: true, stats.field: [{!countDistinct=true }vin, {!countDistinct=true }veh_year]}

Here is sample code of changes I would like, this is example code only, implementation could be a lot better:

```dart
//http 0.11.3+9/lib/src/request.dart - change Map input from <String, String> to <String, Object>
  set bodyFields(Map<String, Object> fields) {
    var contentType = _contentType;
    if (contentType == null) {
      _contentType = new MediaType("application", "x-www-form-urlencoded");
    } else if (contentType.mimeType != "application/x-www-form-urlencoded") {
      throw new StateError('Cannot set the body fields of a Request with '
          'content-type "${contentType.mimeType}".');
    }

    this.body = mapToQuery(fields, encoding: encoding);
  }

//http 0.11.3+9/lib/src/utils.dart - rewrite to put key in multiple times when value is a list.
  /// Converts a [Map] from parameter names to values to a URL query string.
  ///
  ///     mapToQuery({"foo": "bar", "baz": "bang"});
  ///     //=> "foo=bar&baz=bang"
  String mapToQuery(Map<String, Object> map, {Encoding encoding}) {
    String retStr = '';
    String delimiter = '';
    var keys = map.keys;
    for (String key in keys) {
      var value = map[key];
      if (value is List) {
        for (String subvalue in value) {
          retStr +="${delimiter}${key}=${subvalue}";
          delimiter='&';
        }
      }
      else {
        retStr +="${delimiter}${key}=${value}";
        delimiter='&';
      }
    }
    return retStr;
  }
```

_Copied from original issue: dart-lang/sdk#28098_

vsmenon:
_From @lrhn on December 14, 2016 14:18_

I think this is referring to the `http` package, so I've filed an issue over there pointing back to this.

vsmenon:
_From @rwrozelle on December 14, 2016 14:21_

Thank you,

 I followed the link called out at the bottom of https://pub.dartlang.org/packages/http

http://dartbug.com/new

vsmenon:
_From @lrhn on December 14, 2016 14:50_

[...]

PR #49, opened 2016-12-20
Fix for Request set bodyFields, need to be able to have List<String> …

rwrozelle:
…for a map value: https://github.com/dart-lang/http/issues/47

### DIFF
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -11,6 +11,8 @@ import 'base_request.dart';
 import 'byte_stream.dart';
 import 'utils.dart';
 
+import 'package:collection/collection.dart';
+
 /// An HTTP request where the entire request body is known in advance.
 class Request extends BaseRequest {
   /// The size of the request body, in bytes. This is calculated from
@@ -105,7 +107,7 @@ class Request extends BaseRequest {
   /// `application/x-www-form-urlencoded`.
   ///
   /// This map should only be set, not modified in place.
-  Map<String, String> get bodyFields {
+  MultiMap<String, String> get bodyFields {
     var contentType = _contentType;
     if (contentType == null ||
         contentType.mimeType != "application/x-www-form-urlencoded") {
@@ -113,7 +115,7 @@ class Request extends BaseRequest {
           'content-type "application/x-www-form-urlencoded".');
     }
 
-    return splitQueryString(body, encoding: encoding);
+    return MultiMap.splitQueryString(body, encoding: encoding);
   }
 
   set bodyFields(Map<String, Object> fields) {
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -113,10 +113,10 @@ class Request extends BaseRequest {
           'content-type "application/x-www-form-urlencoded".');
     }
 
-    return Uri.splitQueryString(body, encoding: encoding);
+    return splitQueryString(body, encoding: encoding);
   }
 
-  set bodyFields(Map<String, String> fields) {
+  set bodyFields(Map<String, Object> fields) {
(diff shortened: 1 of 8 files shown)


rwrozelle:
Fix for #47

lib/src/request.dart
```diff
@@ -113,10 +113,10 @@ class Request extends BaseRequest {
           'content-type "application/x-www-form-urlencoded".');
     }
 
-    return Uri.splitQueryString(body, encoding: encoding);
+    return splitQueryString(body, encoding: encoding);
```
nex3: Hmm, returning a `List<String>` value here will break backwards-compatibility, as will changing the return type to `Map<String, Object>`. I'm also not sure that returning a map that may have list or string keys is the right API in the abstract; it would make it easy for code written assuming strings to crash or, worse, behave unpredictably if the browser caused a list to be returned instead.

I think the ideal solution here would be to have a sort of multimap class that implements `Map<String, String>` but allows `[]=` to take a list and has a `.getMultiple()` method so users can opt in to getting multiple values. That sort of thing would also be really useful for headers. If you'd like to implement that (maybe in the `collection` package), it would be awesome; if you don't want to, maybe we should just make this feature write-only for now.

rwrozelle: Agree with your assessment, want evolved behavior rather than breaking behavior.  Multi-map would solve problems in other areas as well.  I'll need to bring myself up to speed on lower level details of Dart language and see if I can tackle this.  I do want to try, and will let you know if I'm up to the challenge.  Thanks

rwrozelle: Here is example code of what a MultiMap would look like (if I coded it)
https://github.com/rwrozelle/multimap

Is this what you are looking for?  I've played around with using it in local copies of the http and http_server package and it definitely plays nicely with only minimal updates to existing code and backward compatible.  I decided on using a parent/child component template because it was the least amount of work and I didn't heavily research the existing Mixin classes, so might need to add some "with" stuff on the class definition?

I'm having a heck of a time compiling the SDK, I need to build a new VM on my laptop to get everything lined up and don't have the time until next year, so I can't put together example of code integrated into Collections package.

nex3: > Here is example code of what a MultiMap would look like (if I coded it)
> https://github.com/rwrozelle/multimap
> 
> Is this what you are looking for?  I've played around with using it in local copies of the http and http_server package and it definitely plays nicely with only minimal updates to existing code and backward compatible.  I decided on using a parent/child component template because it was the least amount of work and I didn't heavily research the existing Mixin classes, so might need to add some "with" stuff on the class definition?

The behavior here looks more or less right. We can iterate on the details of the design in a separate CL.

> I'm having a heck of a time compiling the SDK, I need to build a new VM on my laptop to get everything lined up and don't have the time until next year, so I can't put together example of code integrated into Collections package.

Good news: you don't have to add it to the SDK! The [`collection` pub package](https://github.com/dart-lang/collection) is the place for new collection types these days.

rwrozelle: Not familiar with term: CL

I've forked the collection package and added my code, do I perform a pull request and reference the original issues?

nex3: > Not familiar with term: CL

Sorry, that's a Google-ism for "pull request".

> I've forked the collection package and added my code, do I perform a pull request and reference the original issues?

Yes please!

rwrozelle:
This pull request is dependent upon https://github.com/dart-lang/collection/pull/47 being accepted.

natebosch:
https://github.com/dart-lang/collection/pull/47 stalled out. Closing this for now.

PR #54, opened 2017-01-14
Emulate shelf messaging

donny-dont:
Aligns `http`'s Request and Response with what is present in `shelf`


donny-dont:
In `headers.dart` only the `getHeader` is currently being used. It was present in `shelf` so its there.

lib/src/request.dart
```diff
@@ -3,159 +3,94 @@
...
+class Request extends Message {
+  /// The HTTP method of the request. Most commonly "GET" or "POST", less
+  /// commonly "HEAD", "PUT", or "DELETE". Non-standard method names are also
+  /// supported.
+  final String method;
+
+  /// The URL to which the request will be sent.
+  final Uri url;
+
+  /// Creates a new [Request] for [url] using [method].
```
nex3: Document all the optional parameters.

These constructors inherently involve a lot of duplicated documentation, but that makes it a lot easier for users to figure out how to do what they want at a glance.

lib/src/request.dart
```diff
@@ -3,159 +3,94 @@
...
+  Request(this.method, this.url,
+      {dynamic body,
+      Encoding encoding,
+      Map<String, String> headers,
+      Map<String, Object> context})
+      : super(body, encoding: encoding, headers: headers, context: context);
+
+  /// Creates a new [Request] for [url] using a "HEAD" method.
```
nex3: I'd phrase these as e.g. `Creates a new HEAD [Request] to [url].`

lib/src/request.dart
```diff
@@ -3,159 +3,94 @@
...
+      {Map<String, String> headers, Map<String, Object> context})
+      : this('HEAD', url, headers: headers, context: context);
+
+  /// Creates a new [Request] for [url] using a "GET" method.
+  Request.get(Uri url,
+      {Map<String, String> headers, Map<String, Object> context})
+      : this('GET', url, headers: headers, context: context);
+
+  /// Creates a new [Request] for [url] using a "POST" method.
+  Request.post(Uri url,
```
nex3: Constructors for methods that typically have entities should make the body a required parameter.

lib/src/response.dart
```diff
@@ -2,94 +2,49 @@
...
 /// An HTTP response where the entire response body is known in advance.
-class Response extends BaseResponse {
-  /// The bytes comprising the body of this response.
-  final Uint8List bodyBytes;
-
-  /// The body of the response as a string. This is converted from [bodyBytes]
-  /// using the `charset` parameter of the `Content-Type` header field, if
-  /// available. If it's unavailable or if the encoding name is unknown,
-  /// [LATIN1] is used by default, as per [RFC 2616][].
+class Response extends Message {
```
nex3: At some point this should utility getters like `shelf.Response.expires` *et al*. That doesn't need to happen in this PR though.

lib/src/request.dart
```diff
@@ -3,159 +3,94 @@
...
-    if (!finalized) return;
-    throw new StateError("Can't modify a finalized Request.");
+  /// [body] is the request body. It may be either a [String], a [List<int>], a
+  /// [Stream<List<int>>], or `null` to indicate no body.
+  Request change(
+      {Map<String, String> headers,
+      Map<String, Object> context,
+      dynamic body}) {
+    final updatedHeaders = updateMap/*<String, String>*/(this.headers, headers);
+    final updatedContext = updateMap/*<String, Object>*/(this.context, context);
```
nex3: Are these explicit type annotations necessary? Can't it infer the type from the parameters?

lib/src/utils.dart
```diff
@@ -8,15 +8,30 @@ import 'dart:typed_data';
...
 ///     //=> "foo=bar&baz=bang"
 String mapToQuery(Map<String, String> map, {Encoding encoding}) {
   var pairs = <List<String>>[];
-  map.forEach((key, value) =>
-      pairs.add([Uri.encodeQueryComponent(key, encoding: encoding),
-                 Uri.encodeQueryComponent(value, encoding: encoding)]));
+  map.forEach((key, value) => pairs.add([
+        Uri.encodeQueryComponent(key, encoding: encoding),
+        Uri.encodeQueryComponent(value, encoding: encoding)
+      ]));
```
nex3: Please don't include unrelated formatting changes.

lib/src/message.dart
```diff
@@ -0,0 +1,199 @@
...
+import 'dart:convert';
+
+import 'package:collection/collection.dart';
+import 'package:http_parser/http_parser.dart';
+
+import 'body.dart';
+import 'headers.dart';
+import 'http_unmodifiable_map.dart';
+
+Body getBody(Message message) => message._body;
```
donny-dont: Can do. It wasn't documented in `shelf` so I had assumed it was more of a utility function.

lib/src/body.dart
```diff
@@ -44,14 +42,14 @@ class Body {
       stream = new Stream.fromIterable([]);
     } else if (body is String) {
       if (encoding == null) {
-        var encoded = UTF8.encode(body);
+        final encoded = UTF8.encode(body);
```
donny-dont: K I can walk those changes back.

lib/src/headers.dart
```diff
@@ -0,0 +1,47 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+import 'http_unmodifiable_map.dart';
+
+/// Adds a header with [name] and [value] to [headers], which may be null.
+///
+/// Returns a new map without modifying [headers].
+Map<String, String> addHeader(
```
donny-dont: Just `getHeader` is used so far. I wasn't completely positive if the others would be of use as we moved further. I'll remove and add back if needed.

lib/src/http_unmodifiable_map.dart
```diff
@@ -53,6 +53,7 @@ class HttpUnmodifiableMap<V> extends UnmodifiableMapView<String, V> {
 /// A const implementation of an empty [HttpUnmodifiableMap].
 class _EmptyHttpUnmodifiableMap<V> extends MapView<String, V>
     implements HttpUnmodifiableMap<V> {
-  bool get _ignoreKeyCase => true;
   const _EmptyHttpUnmodifiableMap() : super(const {});
+
+  bool get _ignoreKeyCase => true;
```
donny-dont: There's a lint rule there that I had enabled, http://dart-lang.github.io/linter/lints/sort_constructors_first.html

lib/src/message.dart
```diff
@@ -0,0 +1,199 @@
...
+  /// Creates a new [Message] by copying existing values and applying specified
+  /// changes.
+  Message change(
+      {Map<String, String> headers, Map<String, Object> context, dynamic body});
+}
+
+/// Adds information about encoding to [headers].
+///
+/// Returns a new map without modifying [headers].
+Map<String, String> _adjustHeaders(Map<String, String> headers, Body body) {
```
donny-dont: I'll modify `_defaultHeaders` to be completely empty.

lib/src/request.dart
donny-dont: Before I start copy pasting it everywhere how does this sound?

```
  /// Creates a new [Request] for [url] using [method].
  ///
  /// [body] is the request body. It may be either a [String], a [List<int>], a
  /// [Stream<List<int>>], or `null` to indicate no body. If it's a [String],
  /// [encoding] is used to encode it to a [Stream<List<int>>]. The default
  /// encoding is UTF-8.
  ///
  /// [headers] are the HTTP headers for the request.
  ///
  /// Extra [context] can be used by middleware and handler to pass information
  /// between inner middleware and handlers.
```

lib/src/message.dart
```diff
@@ -0,0 +1,199 @@
...
+
+import 'body.dart';
+import 'headers.dart';
+import 'http_unmodifiable_map.dart';
+
+Body getBody(Message message) => message._body;
+
+/// The default set of headers for a message created with no body and no
+/// explicit headers.
+final _defaultHeaders = new HttpUnmodifiableMap<String>({},
```
donny-dont: Modified to not include content length by default

lib/src/request.dart
donny-dont: Fixed

lib/src/message.dart
donny-dont: Fixed

lib/src/http_unmodifiable_map.dart
donny-dont: I can change it back if you desire. I like the rule because it keeps things consistent.

donny-dont:
Okay @nex3 I think I have everything resolved.

lib/src/http_unmodifiable_map.dart
nex3: Yes. In general, please don't change things here just because lints say to—this package doesn't use the linter and it causes churn.

lib/src/request.dart
nex3: :+1:

lib/src/message.dart
```diff
@@ -12,11 +12,12 @@ import 'body.dart';
 import 'headers.dart';
 import 'http_unmodifiable_map.dart';
 
+/// Retrieves the [Body] contained in the [message].
```
nex3: Explain that it's a separate function so the message body is accessible within shelf but hidden elsewhere.

lib/src/message.dart
```diff
@@ -12,11 +12,12 @@ import 'body.dart';
...
 Body getBody(Message message) => message._body;
 
 /// The default set of headers for a message created with no body and no
 /// explicit headers.
-final _defaultHeaders = new HttpUnmodifiableMap<String>({"content-length": "0"},
+final _defaultHeaders = new HttpUnmodifiableMap<String>({},
```
nex3: I don't think just making this empty is enough. We still need to send `content-length: 0` for `null`-bodied POSTs and other requests that have content.

lib/src/message.dart
```diff
@@ -26,13 +27,13 @@ abstract class Message {
...
 
-  /// Extra context that can be used by for middleware and handlers.
+  /// Extra context that can be used by middleware and handlers.
   ///
   /// For requests, this is used to pass data to inner middleware and handlers;
   /// for responses, it's used to pass data to outer middleware and handlers.
   ///
   /// Context properties that are used by a particular package should begin with
-  /// that package's name followed by a period. For example, if [logRequests]
+  /// that package's name followed by a period. For example, if `logRequests`
```
nex3: Referring to `logRequests` here seems strange since that's not middleware that exists—and probably not something this package will ever provide. Maybe rephrase this as describing a hypothetical middleware?

lib/src/message.dart
```diff
@@ -26,13 +27,13 @@ abstract class Message {
...
   /// wanted to take a prefix, its property name would be `"http.prefix"`,
```
nex3: "take a prefix" -> "take a context property"

lib/src/request.dart
```diff
@@ -7,62 +7,124 @@ import 'dart:convert';
 import 'message.dart';
 import 'utils.dart';
 
-/// Represents an HTTP request to be processed by a `http` application.
+/// Represents a HTTP request to be sent to a server.
```
nex3: "an" is correct here.

lib/src/message.dart
donny-dont: Ok so I'm not totally sure the best way to get all this working as you would expect. So how about I create unit tests on what your expectation is and we go from there?

nex3: Actually, re-reading [the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html), it says

> The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers.

So it's probably fine to just not include a `content-length` header for requests that have no entity body, regardless of their method. We'll need to make sure that `dart:io` doesn't force a chunked transfer encoding in those cases, though.

PR #55, opened 2017-02-15
Stop using dart:mirrors.

### LINKED ISSUE #1: Remove dart:mirror usage from package:http
A change to the dart2js compiler introduced a warning on mirror usage in dart code -- see [Emit warning on import of dart:mirrors.](https://codereview.chromium.org//588183002). According to the warning, users will be forced to pass an additional flag to the dart2js compiler when their code is using dart:mirrors. This affects all users of `package:http`.

When compiling the following example

``` dart
main()=> print("hello");
```

dart2js will print out the following warning:

```
Warning: 
****************************************************************
* WARNING: dart:mirrors support in dart2js is experimental,
*          and not recommended.
*          This implementation of mirrors is incomplete,
*          and often greatly increases the size of the generated
*          JavaScript code.
*
* Your app imports dart:mirrors via:
*   x.dart => package:http => dart:mirrors
*
* Starting with Dart 1.9, you must use the
* --enable-experimental-mirrors command-line flag to opt-in.
* You can begin using this flag now if mirrors support is critical.
*
* To learn what to do next, please visit:
*    http://dartlang.org/dart2js-reflection
****************************************************************
```

That flag does not seem to be enforced as of SDK version 1.9.0-dev.1.0+google3-trunk, but that seems to be only a matter of time.

nex3:
Closes #1


lib/src/client.dart
```diff
@@ -28,10 +27,7 @@ abstract class Client {
   /// Currently this will create an [IOClient] if `dart:io` is available and
   /// throw an [UnsupportedError] otherwise. In the future, it will create a
   /// [BrowserClient] if `dart:html` is available.
```
munificent: This comment isn't relevant anymore, since you can't import this library without "dart:io" being available, can you?

lib/src/io_client.dart
```diff
@@ -3,37 +3,24 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
```
munificent: Does this mean "package:http/http.dart" can only be imported on the command line VM? If so, isn't that a breaking change?

lib/src/client.dart
nex3: It can now be importable while also being unavailable in the sense that its API just throws `UnsupportedError`s.

lib/src/io_client.dart
nex3: As of 1.22, `dart:io` is importable on all platforms and will just throw `UnsupportedError`s at runtime if it's not supported.

munificent: What?! Where was that announced? It's not in the CHANGELOG.

nex3: That's a @dgrove question.

skybrian: Is the change to dart:io for 1.22 or 1.23? (I heard it was 1.23.)

matanlurey: I have a fix out as https://github.com/dart-lang/http/pull/58

PR #56, opened 2017-02-17
Update Client interface

donny-dont:
This updates the `Client` interface exposed and removes the extensions of `Request` and `Response` that are no longer applicable.


lib/http.dart
```diff
@@ -32,7 +32,7 @@ export 'src/streamed_response.dart';
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> head(url, {Map<String, String> headers}) =>
+FutureOr<Response> head(url, {Map<String, String> headers}) =>
```
kevmoo: Uh...these are *always* async, right?

Why the move to `FutureOr`?

donny-dont: @kevmoo the intent is to move to middleware like shelf.

kevmoo: Okay...but do we imagine a future where these calls may be sync?

donny-dont: Are we just talking http.dart or throughout? With the http helpers no.

kevmoo: Exactly. So this method should always return `Future`. `FutureOr` implies that it may synchronously return `Response`.

donny-dont:
@kevmoo fixed the interface.

lib/http.dart
```diff
@@ -161,7 +157,7 @@ Future<String> read(url, {Map<String, String> headers}) =>
 Future<Uint8List> readBytes(url, {Map<String, String> headers}) =>
   _withClient((client) => client.readBytes(url, headers: headers));
 
-Future/*<T>*/ _withClient/*<T>*/(Future/*<T>*/ fn(Client client)) async {
+Future<T> _withClient<T>(FutureOr<T> fn(Client client)) async {
```
nex3: Why is this `FutureOr`? I'm pretty sure all the functions return `Future`s.

lib/src/base_client.dart
```diff
@@ -23,15 +21,15 @@ abstract class BaseClient implements Client {
   /// can be a [Uri] or a [String].
   ///
   /// For more fine-grained control over the request, use [send] instead.
-  Future<Response> head(url, {Map<String, String> headers}) =>
-    _sendUnstreamed("HEAD", url, headers);
+  FutureOr<Response> head(url, {Map<String, String> headers}) =>
```
nex3: Same question here: why `FutureOr`? This is never going to return a `Response` synchronously.

lib/src/base_client.dart
```diff
@@ -23,15 +21,15 @@ abstract class BaseClient implements Client {
...
+    send(new Request.head(_uri(url), headers: headers));
```
nex3: The request constructors should take dynamic `url`s, not `Uri`s.

donny-dont: How about the named constructors take dynamic but the default takes a Uri? If all of them take dynamic then there's a check on each call to `change`.

nex3: I'd rather have a cleaner public API and add a private constructor for `change` to call, I think.

donny-dont: K works for me 👍

donny-dont:
@nex3 @kevmoo think we're ready to go know.

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: "Dart Team <misc@dartlang.org>"
 homepage: https://github.com/dart-lang/http
```
kevmoo: We should make sure to update the pubspec version to make a major change, too.

kevmoo: We should make sure to update the pubspec version to make a major change, too.

donny-dont: Updated to 0.12.0

kevmoo: 👍

lib/src/request.dart
```diff
@@ -147,10 +151,22 @@ class Request extends Message {
...
+        this.method,
+        this.url,
+        body ?? getBody(this),
+        this.encoding,
+        updatedHeaders,
+        updatedContext);
+  }
+}
+
+Uri _uri(url) {
```
nex3: Either make this a static method of `Request`, or put it in `utils`.

lib/src/request.dart
```diff
@@ -30,12 +30,12 @@ class Request extends Message {
   ///
   /// Extra [context] can be used to pass information between inner middleware
   /// and handlers.
-  Request(this.method, this.url,
+  Request(String method, url,
```
nex3: Document the acceptable types for `url`. You can probably just copy the docs from the top-level methods.

pubspec.yaml
nex3: This should be 0.12.0-dev. We use the "-dev" suffix to clearly distinguish between released versions and those that are still in progress.

donny-dont:
🎊

Issue #57, opened 2017-02-17
Dependency on dart:io


parnham:
The latest update `0.11.3+10` which removes the mirrors dependency, adds a dependency to "dart:io" which the BrowserClient can see (via BaseClient -> Client -> IOClient). Whilst it still works when compiled to javascript, when running in Dartium it throws the following error:

```
The requested built-in library is not available on Dartium.'package:http/src/io_client.dart': error: line 6 pos 1: library handler failed
import 'dart:io';
^: package:http/src/io_client.dart
```

matanlurey:
Have you updated your version of Dartium? I believe you need a version that has the `1.22.0` VM.

parnham:
That was my first thought also :) 
Have just double checked and I am running `dartium-linux-x64-stable-1.22.0.0`  and still getting the error I'm afraid.

parnham:
I just pulled `1.23.0` instead and our application is working again.

matanlurey:
Thanks! Looks like we need to up the pubspec, it's not accurate.

PR #58, opened 2017-02-17
Correct the pubspec.yaml, we require 1.23.0

matanlurey:
... for importing `dart:io` into Dartium, specifically it looks like.

Closes https://github.com/dart-lang/http/issues/57.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: http
-version: 0.11.3+10
+version: 0.11.3+11
 author: "Dart Team <misc@dartlang.org>"
 homepage: https://github.com/dart-lang/http
 description: A composable, Future-based API for making HTTP requests.
@@ -12,4 +12,4 @@ dependencies:
 dev_dependencies:
   unittest: ">=0.9.0 <0.12.0"
 environment:
-  sdk: ">=1.22.0 <2.0.0"
+  sdk: ">=1.23.0-dev.0.0 <2.0.0"


dustyholmes-wf:
It seems that this solution will leave people on 1.22.0 pulling version 0.11.3+10 incorrectly. Should there be a version in between this fix and the current that corrects the library for the stable release of dart?

matanlurey:
I think we're going to try and pull the previous version... @kevmoo  any update?

nex3:
Can you add a CHANGELOG entry for this?

matanlurey:
I don't want to merge this until we figure out how to fix the previous release (i.e. either roll it back first or publish a new 1.22.0 release with the last commit reverted). @kevmoo WDUT?

kevmoo:
I vote for rollback. My understanding is that natalie is running w/ that -
as of an hour ago...

dustyholmes-wf:
I no longer see 0.11.3+10 on pub. Looks like it has been rolled back.

nex3:
Yep, I yanked 0.11.3+10. Given that, I think it might make sense to wait on landing/releasing 0.11.3+11 until 1.23 is stable. @kevmoo, how urgent is it that we have a mirrorless version of this released?

kevmoo:
Natalie: will follow-up w/ you offline.

kevmoo:
@nex3 Let's get this out w/ the `-dev` constraint. It'll let our flutter folks roll forward.

tvolkert:
Even once SDK 1.23 is stable, some users won't have it, so do we need to release a `0.11.3+11` version cut off a branch that doesn't import `dart:io` anyway?

nex3:
@tvolkert No, we yanked the version with the invalid SDK constraint.

PR #63, opened 2017-03-06
Don't quote the multipart boundary header.

### LINKED ISSUE #61: Consider sending unquoted boundary in multipart requests
Some naive back-ends are unable to handle quoted boundary values, specifically those implemented based on [RFC 1867](https://www.ietf.org/rfc/rfc1867.txt), which, while refers to [RFC 1521](https://www.ietf.org/rfc/rfc1521.txt), does not itself call out the rules for encoding the boundary property.

nex3:
This is be more broadly compatible than the quoted version, although the
quoted version is what's described by the spec. curl and browsers send
unquoted boundaries, so we can safely assume that any server will
support them.

Closes #61

### DIFF
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -117,14 +117,19 @@ class MultipartRequest extends BaseRequest {
     return new ByteStream(controller.stream);
   }
 
-  /// All character codes that are valid in multipart boundaries. From
-  /// http://tools.ietf.org/html/rfc2046#section-5.1.1.
+  /// All character codes that are valid in multipart boundaries. This is the
+  /// intersection of the characters allowed in the `bcharsnospace` production
+  /// defined in [RFC 2046][] and those allowed in the `token` production
+  /// defined in [RFC 1521][].
+  ///
+  /// [RFC 2046]: http://tools.ietf.org/html/rfc2046#section-5.1.1.
+  /// [RFC 1521]: https://tools.ietf.org/html/rfc1521#section-4
   static const List<int> _BOUNDARY_CHARACTERS = const <int>[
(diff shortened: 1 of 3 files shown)


lib/src/multipart_request.dart
```diff
@@ -83,7 +83,7 @@ class MultipartRequest extends BaseRequest {
   ByteStream finalize() {
     // TODO(nweiz): freeze fields and files
     var boundary = _boundaryString();
-    headers['content-type'] = 'multipart/form-data; boundary="$boundary"';
+    headers['content-type'] = 'multipart/form-data; boundary=$boundary';
```
yjbanov: I think we'll have to remove 39 (`'`) and 61 (`=`) in `_BOUNDARY_CHARACTERS` for this to really work. I'd actually leave just alpha-numerics, 95 (underscore) and 45 (hyphen). I think all others play some special role in URL encoding, so I'd leave them out just in case.

yjbanov: Here are the suspicious ones:

```
39: '
40: (
41: )
43: +
44: ,
46: .
47: /
58: :
61: =
63: ?
```

nex3: I don't want to go alphanumeric-only unless we absolutely have to, but I've removed all the characters that aren't valid in MIME tokens.

tvolkert:
FYI, this change causes the following tests to fail:

https://github.com/dart-lang/http/blob/0.11.3%2B12/test/multipart_test.dart
https://github.com/dart-lang/http/blob/0.11.3%2B12/test/io/multipart_test.dart

FYI, I've disabled the tests internally when bumping the version of `package:http` within Google.

kevmoo:
@yjbanov any chance you could look at the failing tests?

yjbanov:
I'll take a look.

PR #72, opened 2017-03-29
remove boundary characters that pkg:http_parser cannot parse

### LINKED ISSUE #71: Investigate/fix test failures after boundary change
See https://github.com/dart-lang/http/pull/63#issuecomment-289933929

yjbanov:
Fixes #71 

FYI, I'm not familiar with this project's setup. It seems the relevant code is on branch 0.11.x?

/cc @tvolkert @nex3

### DIFF
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 import 'dart:math';
 
 import 'base_request.dart';
+import 'boundary_characters.dart';
 import 'byte_stream.dart';
 import 'multipart_file.dart';
 import 'utils.dart';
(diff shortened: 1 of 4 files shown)


test/io/utils.dart
```diff
@@ -85,7 +85,7 @@ Future startServer() {
           requestBody = requestBodyBytes;
         }
 
-        var content = {
+        var content = <String, dynamic>{
```
yjbanov: FYI, the reason for this change was strong mode inferred the map as `Map<String, Object>` and then complained about expression `content['headers'][name]` on line 98, saying that `Object` lacks `operator[]`.

lib/src/multipart_request.dart
```diff
@@ -124,8 +126,9 @@ class MultipartRequest extends BaseRequest {
   ///
   /// [RFC 2046]: http://tools.ietf.org/html/rfc2046#section-5.1.1.
   /// [RFC 1521]: https://tools.ietf.org/html/rfc1521#section-4
-  static const List<int> _BOUNDARY_CHARACTERS = const <int>[
-    39, 43, 95, 44, 45, 46, 58, 61, 63, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57,
+  @visibleForTesting
+  static const List<int> BOUNDARY_CHARACTERS = const <int>[
```
kevmoo: Better idea

1) Put this in another file in src
2) import here
3) import via `lib/src/...` in the test
4) Don't expose another top-level field

I was going to say exclude from 'lib/http.dart' but that's more error prone

kevmoo: ...then you don't need `pkg/meta` either 😄

yjbanov: Done.

PR #73, opened 2017-03-30
version bump; changelog about boundary characters

yjbanov:


yjbanov:
> Just noticed the array is out-of-order.

It [always has been](https://github.com/dart-lang/http/blob/9076857998d27fca722bc48f7bc9ad17789cf91f/lib/src/multipart_request.dart#L127)

Issue #75, opened 2017-04-06
Null request payload/body values not handled properly


KOGI:
It appears that `http/src/utils.dart`'s `mapToQuery()` method doesn't properly handle objects with null values causing unhandled exceptions:

```error
Unhandled exception:
NoSuchMethodError: The getter 'length' was called on null.
Receiver: null
Tried calling: length
...
#3      Uri.encodeQueryComponent (dart:core/uri.dart:1055)
...
```
---
## Repro code:
```dart
import 'package:http/http.dart' as http;

main(List<String> args) async {
  // ======== Works
  try {
    // -- load model
    var foo = loadExistingModel("123");
    // -- modify
    foo.message += " modified";
    // -- save
    var postResult = await http.put('http://www.example.com/myModel', body: foo.asMap);
    print("++update succeeded");

  } catch(e) {
    print("!!update failed");
    rethrow;
  }

  // ======== Broken (due to `null` value for `id` property)
  try {
    // -- create model
    var foo = new MyModel("brand new model");
    // -- save new
    var postResult = await http.post('http://www.example.com/myModel', body: foo.asMap);
    print("++create succeeded");

  } catch(e) {
    print("!!create failed");
    rethrow;
  }
}

class MyModel {
  final String id; // -- only populated once saved
  String message;

  MyModel(this.message, [this.id]);

  Map<String, dynamic> get asMap {
    return {
      "id"     : id,
      "message": message
    };
  }
}

loadExistingModel(String id) {
  // fetch model by id
  // getResult = await http.get('http://www.example.com/myModel/$id');
  return new MyModel("existing model", id);
}
```
---
##### [http-0.11.3+9/lib/src/utils.dart](https://github.com/dart-lang/http/blob/master/lib/src/utils.dart#L28)
```patch
 String mapToQuery(Map<String, String> map, {Encoding encoding}) {
   var pairs = <List<String>>[];
   map.forEach((key, value) =>
+      if (value == null) {
+        return; // -- skip null elements
+      }
+
       pairs.add([Uri.encodeQueryComponent(key, encoding: encoding),
                  Uri.encodeQueryComponent(value, encoding: encoding)]));
+  }
   return pairs.map((pair) => "${pair[0]}=${pair[1]}").join("&");
 }
```

---
##### [http-0.11.3+9/lib/src/utils.dart](https://github.com/dart-lang/http/blob/master/lib/src/utils.dart#L28)
```patch
[...]

PR #77, opened 2017-04-14
Properly handle `null` field values when encoding requests.

KOGI:
This fixes Issue #75

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -13,6 +13,10 @@ import 'byte_stream.dart';
 ///     mapToQuery({"foo": "bar", "baz": "bang"});
 ///     //=> "foo=bar&baz=bang"
 String mapToQuery(Map<String, String> map, {Encoding encoding}) {
+  map.keys
+    .where((k) => (map[k] == null)).toList() // -- keys for null elements
+    .forEach(map.remove);
+
   var pairs = <List<String>>[];
   map.forEach((key, value) =>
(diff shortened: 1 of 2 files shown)


lib/src/utils.dart
```diff
@@ -30,6 +30,10 @@ Map/*<K, V>*/ updateMap/*<K, V>*/(
 ///     mapToQuery({"foo": "bar", "baz": "bang"});
 ///     //=> "foo=bar&baz=bang"
 String mapToQuery(Map<String, String> map, {Encoding encoding}) {
+  map.keys
```
kevmoo: Create a local variable and do this in two steps. It'll be easier to read.

Also: please add a test that exercises this case.

KOGI:
I signed it!

lib/src/utils.dart
KOGI: Both done. The tests were super simple changes, so I don't expect any problems, but, I don't have sdk v1.23 installed due to local requirements for 1.22, so I couldn't *actually* run the test suite.

donny-dont:
@kevmoo I don't see any references to `mapToQuery` in `master`. Shouldn't this target the 0.11 branch?

KOGI:
@donny-dont you're totally right, my bad. I just pushed a rebase and changed the PR base.

Also, not sure what happened to my cla... it *was* there...

nex3:
See https://github.com/dart-lang/http/issues/75#issuecomment-301630690

PR #78, opened 2017-04-14
Adding back missing Request and Response fields

donny-dont:
There were a few fields that were not present in previous pull requests.

Also wondering if `persistentConnection`, `followRedirects`, and `maxRedirects` should be modifiable within the `change` method of `Request`.

### DIFF
--- a/lib/src/response.dart
+++ b/lib/src/response.dart
@@ -15,7 +15,7 @@ class Response extends Message {
   ///
   /// The value takes into account any redirects that occurred during the
   /// request.
-  final Uri location;
+  final Uri finalUrl;
 
   /// The status code of the response.
   final int statusCode;
@@ -23,7 +23,7 @@ class Response extends Message {
   /// The reason phrase associated with the status code.
   final String reasonPhrase;
 
-  /// Creates a new HTTP response for a resource at the [location], which can
+  /// Creates a new HTTP response for a resource at the [finalUrl], which can
   /// be a [Uri] or a [String], with the given [statusCode].
   ///
   /// [body] is the request body. It may be either a [String], a [List<int>], a
@@ -36,16 +36,16 @@ class Response extends Message {
   ///
   /// Extra [context] can be used to pass information between outer middleware
   /// and handlers.
-  Response(location, int statusCode,
+  Response(finalUrl, int statusCode,
       {String reasonPhrase,
       body,
       Encoding encoding,
       Map<String, String> headers,
       Map<String, Object> context})
-      : this._(getUrl(location), statusCode, reasonPhrase ?? '',
+      : this._(getUrl(finalUrl), statusCode, reasonPhrase ?? '',
       body, encoding, headers, context);
 
-  Response._(this.location, this.statusCode, this.reasonPhrase,
+  Response._(this.finalUrl, this.statusCode, this.reasonPhrase,
       body,
       Encoding encoding,
       Map<String, String> headers,
@@ -75,7 +75,7 @@ class Response extends Message {
     var updatedContext = updateMap(this.context, context);
 
     return new Response._(
-        this.location,
+        this.finalUrl,
         this.statusCode,
         this.reasonPhrase,
         body ?? getBody(this),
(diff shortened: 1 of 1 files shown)


nex3:
Following shelf's lead, I'd like to make these part of the `context` map rather than including them on the main objects because they're only meaningful for server clients, and can't be supported on the browser.

donny-dont:
@nex3 works for me! If that's the case then next up is the dart:io client. I have some stuff for that locally.

So with that is there a preference for the name within context for those values?

nex3:
I think `io.persistentConnection`?

donny-dont:
@nex3 do we care about reasonPhrase anymore?

nex3:
I think `reasonPhrase` is still useful as a field—it should be available from all implementations, too.

donny-dont:
K works for me.

I'm also wondering if `url` should get moved into `Message`. Both `dart:html` and `dart:io` apps can get the redirected url. I don't see any other way to get whether a request was redirected.

nex3:
It makes sense to expose that information, but I don't know that it makes sense to use the same name since the meaning is different. Maybe call it `finalUrl` on the response?

donny-dont:
Maybe `location`? https://en.wikipedia.org/wiki/HTTP_location

nex3:
Yeah, I like that.

donny-dont:
@nex3 good to go.

nex3:
Shouldn't `location` come from the headers?

donny-dont:
Its only in the headers when there is a redirect or a newly created resource. I verified it through  https://httpbin.org/ 's methods and only saw it if redirects were off.

Change the name?

nex3:
Maybe we should just have it be `null` if it's not set for those reasons, then.

donny-dont:
Okay so I don't think `location` is the best name anymore because it only seems to apply to actual redirects. Here's some example information using httpbin which has a redirect to URL endpoint.

I have some local patches to enable the `dart:io` client. It uses [RedirectInfo](https://api.dartlang.org/stable/1.23.0/dart-io/RedirectInfo-class.html) to determine where the final request is and puts it at `location` when creating the `Response`.

Request when `io.followRedirects` is `false`.

```
REQUEST
URL: https://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F

RESPONSE
URL: https://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F
Status: 302
Headers: 
x-powered-by: Flask
access-control-allow-credentials: true
connection: close
location: http://example.com/
date: Tue, 23 May 2017 18:20:09 GMT
access-control-allow-origin: *
content-length: 0
via: 1.1 vegur
content-type: text/html; charset=utf-8
x-processed-time: 0.000843048095703
server: meinheld/0.6.1
```

So `location` is there and its pointing to the redirect.

Now if `io.followRedirects` is true the result is the following.

```
REQUEST
URL: https://httpbin.org/redirect-to?url=http%3A%2F%2Fexample.com%2F

RESPONSE
URL: http://example.com/
Status: 200
Headers: 

connection: close
cache-control: max-age=604800
last-modified: Fri, 09 Aug 2013 23:54:35 GMT
date: Tue, 23 May 2017 18:29:26 GMT
content-encoding: gzip
vary: Accept-Encoding
content-type: text/html
server: ECS (cpm/F9D5)
accept-ranges: bytes
content-length: 606
etag: "359670651+gzip"
x-cache: HIT
expires: Tue, 30 May 2017 18:29:26 GMT
```

With the redirect followed `location` is not present.

With this information I'm leaning towards moving `url` to the base `Message` class. They both have a url associated with them. If the `url` is not added to the `Response` then there is no way to determine if there was a redirect.

nex3:
I don't like the idea of using the name `url`, since the semantics are different. I think I still like `finalUrl`.

donny-dont:
And named changed as requested. Can't think of a better name for it 🤷‍♂️

nex3:
Is this ready to land now?

donny-dont:
Yep :shipit:

PR #82, opened 2017-05-24
Adding dart:io client

donny-dont:
Implements a `dart:io` client with the new interface.

### DIFF
--- a/lib/src/io_client.dart
+++ b/lib/src/io_client.dart
@@ -24,8 +24,6 @@ class IOClient extends BaseClient {
 
   /// Sends an HTTP request and asynchronously returns the response.
   Future<Response> send(Request request) async {
-    var stream = await request.read();
-
     try {
       var ioRequest = await _inner.openUrl(request.method, request.url);
       var context = request.context;
@@ -41,9 +39,8 @@ class IOClient extends BaseClient {
         ioRequest.headers.set(name, value);
       });
 
-      var response = await stream.pipe(
-          DelegatingStreamConsumer.typed<List<int>>(ioRequest)
-      ) as HttpClientResponse;
+      request.read().pipe(DelegatingStreamConsumer.typed<List<int>>(ioRequest));
+      var response = await ioRequest.done;
 
       var headers = <String, String>{};
       response.headers.forEach((key, values) {
@@ -57,8 +54,7 @@ class IOClient extends BaseClient {
           body: DelegatingStream.typed<List<int>>(response).handleError(
               (error) => throw new ClientException(error.message, error.uri),
               test: (error) => error is HttpException),
-          headers: headers,
-          context: context);
+          headers: headers);
     } on HttpException catch (error) {
       throw new ClientException(error.message, error.uri);
     } on SocketException catch (error) {
@@ -74,15 +70,17 @@ class IOClient extends BaseClient {
   }
 }
 
+/// Determines the finalUrl retrieved by evaluating any redirects received in
+/// the [response] based on the initial [request].
 Uri _responseUrl(Request request, HttpClientResponse response) {
-  var redirects = response.redirects;
+  var finalUrl = request.url;
 
-  if (redirects.isEmpty) {
-    return request.url;
+  for (var redirect in response.redirects) {
+    var location = redirect.location;
+
+    // Redirects can either be absolute or relative
+    finalUrl = location.isAbsolute ? location : finalUrl.resolveUri(location);
   }
 
-  var location = redirects.last.location;
-
-  // Redirect can be relative or absolute
-  return (location.isAbsolute) ? location : request.url.resolveUri(location);
+  return finalUrl;
 }
--- a/lib/src/io_client.dart
+++ b/lib/src/io_client.dart
@@ -68,19 +68,19 @@ class IOClient extends BaseClient {
     if (_inner != null) _inner.close(force: true);
     _inner = null;
   }
-}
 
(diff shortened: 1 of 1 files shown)


lib/src/io_client.dart
```diff
@@ -72,3 +73,16 @@ class IOClient extends BaseClient {
...
 }
+
+Uri _responseUrl(Request request, HttpClientResponse response) {
+  var redirects = response.redirects;
+
+  if (redirects.isEmpty) {
+    return request.url;
+  }
+
+  var location = redirects.last.location;
```
donny-dont: Actually I might just have to run through the array in case there was some ping ponging between servers. Not likely but certainly possible.

lib/src/io_client.dart
```diff
@@ -23,45 +23,46 @@ class IOClient extends BaseClient {
   IOClient([HttpClient inner]) : _inner = inner ?? new HttpClient();
 
   /// Sends an HTTP request and asynchronously returns the response.
-  Future<StreamedResponse> send(BaseRequest request) async {
-    var stream = request.finalize();
+  Future<Response> send(Request request) async {
+    var stream = await request.read();
```
nex3: `await` doesn't do anything here since this is a `Stream`, right?

lib/src/io_client.dart
```diff
@@ -23,45 +23,46 @@ class IOClient extends BaseClient {
...
-          ..persistentConnection = request.persistentConnection;
+          ..persistentConnection = context['io.persistentConnection'] ?? true;
       request.headers.forEach((name, value) {
         ioRequest.headers.set(name, value);
       });
 
       var response = await stream.pipe(
-          DelegatingStreamConsumer.typed(ioRequest));
+          DelegatingStreamConsumer.typed<List<int>>(ioRequest)
+      ) as HttpClientResponse;
```
nex3: Rather than casting here, just await `ioRequest.done`.

lib/src/io_client.dart
```diff
@@ -23,45 +23,46 @@ class IOClient extends BaseClient {
...
-          request: request,
+          reasonPhrase: response.reasonPhrase,
+          body: DelegatingStream.typed<List<int>>(response).handleError(
+              (error) => throw new ClientException(error.message, error.uri),
+              test: (error) => error is HttpException),
           headers: headers,
-          isRedirect: response.isRedirect,
-          persistentConnection: response.persistentConnection,
-          reasonPhrase: response.reasonPhrase);
+          context: context);
```
nex3: I don't think it's correct to send the request's context back down the chain as part of the response.

lib/src/io_client.dart
```diff
@@ -72,3 +73,16 @@ class IOClient extends BaseClient {
     _inner = null;
   }
 }
+
+Uri _responseUrl(Request request, HttpClientResponse response) {
```
nex3: This should be part of the class, and also documented.

lib/src/io_client.dart
nex3: Why wouldn't the last location be correct?

lib/src/io_client.dart
```diff
@@ -72,3 +73,16 @@ class IOClient extends BaseClient {
...
+  var redirects = response.redirects;
+
+  if (redirects.isEmpty) {
+    return request.url;
+  }
+
+  var location = redirects.last.location;
+
+  // Redirect can be relative or absolute
+  return (location.isAbsolute) ? location : request.url.resolveUri(location);
```
nex3: Nit: unnecessary parens.

lib/src/io_client.dart
donny-dont: Hypothetically if you called to `a.com/foo` that could redirect to another server `b.com/bar` which then does a relative redirect to `b.com/baz`.

I think I just need to run through the array with the `lastLocation` and use that to get the final url.

lib/src/io_client.dart
donny-dont: K you want it static right? Was just following along with what we we're doing with free functions in `message.dart`

lib/src/io_client.dart
nex3: Can you do `p.url.joinAll()` for that?

donny-dont: @nex3 did not know about `p.url.joinAll()` is that still valid with the updated implementation?

lib/src/io_client.dart
donny-dont: fixed

lib/src/io_client.dart
donny-dont: fixed

lib/src/io_client.dart
donny-dont: removed so assuming that it should just be an empty context

nex3: For now, yeah.

lib/src/io_client.dart
nex3: > K you want it static right?

No, as an instance method. I don't like making things static unless they have to be.

> Was just following along with what we we're doing with free functions in `message.dart`

Those are used in constructors, which is pretty different. Still, I might make them static methods.

lib/src/io_client.dart
nex3: The new implementation looks fine as-is.

donny-dont:
@nex3 think everything is resolved now

:shipit: ?

donny-dont:
🎊 thanks @nex3 ! Onto browser_client now

PR #83, opened 2017-05-24
Add a dart:html client

donny-dont:
Adding the browser_client.dart implementation

### DIFF
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -10,7 +10,6 @@ import 'package:async/async.dart';
 import 'package:stack_trace/stack_trace.dart';
 
 import 'src/base_client.dart';
-import 'src/byte_stream.dart';
 import 'src/exception.dart';
 import 'src/request.dart';
 import 'src/response.dart';
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -34,12 +34,6 @@ class BrowserClient extends BaseClient {
   /// Creates a new HTTP client.
   BrowserClient();
 
-  /// Whether to send credentials such as cookies or authorization headers for
-  /// cross-site requests.
-  ///
-  /// Defaults to `false`.
-  bool withCredentials = false;
-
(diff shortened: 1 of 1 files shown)


lib/browser_client.dart
```diff
@@ -41,16 +41,17 @@ class BrowserClient extends BaseClient {
   bool withCredentials = false;
 
   /// Sends an HTTP request and asynchronously returns the response.
-  Future<StreamedResponse> send(BaseRequest request) async {
-    var bytes = await request.finalize().toBytes();
+  Future<Response> send(Request request) async {
+    var stream = new ByteStream(request.read());
```
nex3: I really want to kill `ByteStream`, so we shouldn't use it here. You can pull this functionality into a function in `utils/`, although really it should live in a package somewhere.

donny-dont:
The `responseUrl` does not appear to need any modification. Its the absolute path from what I see in testing.

lib/browser_client.dart
```diff
@@ -59,13 +60,12 @@ class BrowserClient extends BaseClient {
...
-            new ByteStream.fromBytes(body),
+        completer.complete(new Response(
+            xhr.responseUrl,
             xhr.status,
-            contentLength: body.length,
-            request: request,
-            headers: xhr.responseHeaders,
-            reasonPhrase: xhr.statusText));
+            reasonPhrase: xhr.statusText,
+            body: new ByteStream.fromBytes(body),
```
nex3: This also shouldn't be a `ByteStream`. In this case, just `new Stream.fromIterable([body])` should work.

lib/browser_client.dart
donny-dont: It appears it IS in a package 👍

lib/browser_client.dart
donny-dont: Fixed

donny-dont:
@nex3 :shipit: ?

donny-dont:
Nice! Another one down 😄

PR #85, opened 2017-05-25
Adding middleware

donny-dont:
Currently a work in progress to see if this is on the right track. Will add documentation during the review process.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'base_client.dart';
+import 'handler_client.dart';
 import 'io_client.dart';
 import 'request.dart';
 import 'response.dart';
@@ -27,6 +28,9 @@ abstract class Client {
   /// throw an [UnsupportedError] otherwise. In the future, it will create a
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
+
+  factory Client.handler(Handler handler, {CloseHandler onClose})
+      => new HandlerClient(handler, onClose ?? () {});
 
   /// Sends an HTTP HEAD request with the given headers to the given URL, which
   /// can be a [Uri] or a [String].
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'base_client.dart';
+import 'handler.dart';
 import 'handler_client.dart';
 import 'io_client.dart';
 import 'request.dart';
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -30,6 +30,13 @@ abstract class Client {
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
 
+  /// Creates a new client handler.
+  ///
+  /// The [handler] is a function that receives a [Request] and returns a
+  /// [Future<Response>]. It will be called when [Client.send] is invoked. This
+  /// allows for composition of a [Client] within a larger application.
+  ///
+  /// When [Client.close] is called the [onClose] function will be called.
   factory Client.handler(Handler handler, {void onClose()})
       => new HandlerClient(handler, onClose ?? () {});
 
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -29,7 +29,7 @@ abstract class Client {
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
 
-  factory Client.handler(Handler handler, {CloseHandler onClose})
+  factory Client.handler(Handler handler, {void onClose()})
       => new HandlerClient(handler, onClose ?? () {});
 
   /// Sends an HTTP HEAD request with the given headers to the given URL, which
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -30,11 +30,10 @@ abstract class Client {
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
 
-  /// Creates a new client handler.
+  /// Creates a new [Client] from a [handler] callback.
   ///
   /// The [handler] is a function that receives a [Request] and returns a
-  /// [Future<Response>]. It will be called when [Client.send] is invoked. This
-  /// allows for composition of a [Client] within a larger application.
+  /// [Future<Response>]. It will be called when [Client.send] is invoked.
   ///
   /// When [Client.close] is called the [onClose] function will be called.
   factory Client.handler(Handler handler, {void onClose()})
--- a/lib/src/handler.dart
+++ b/lib/src/handler.dart
@@ -10,6 +10,6 @@ import 'response.dart';
 /// The signature of a function which handles a [Request] and returns a
 /// [Future<Response>].
 ///
-/// A [Handler] may receive a request directly for a HTTP resource or it may be
-/// composed as part of a larger application.
+/// A [Handler] may call an underlying HTTP implementation, or it may wrap
+/// another [Handler] or a [Client].
 typedef Future<Response> Handler(Request request);
--- a/lib/src/handler.dart
+++ b/lib/src/handler.dart
@@ -1,0 +1,10 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'request.dart';
+import 'response.dart';
+
+typedef Future<Response> Handler(Request request);
--- a/lib/src/handler.dart
+++ b/lib/src/handler.dart
@@ -7,4 +7,9 @@ import 'dart:async';
 import 'request.dart';
 import 'response.dart';
 
+/// The signature of a function which handles a [Request] and returns a
+/// [Future<Response>].
+///
+/// A [Handler] may receive a request directly for a HTTP resource or it may be
+/// composed as part of a larger application.
 typedef Future<Response> Handler(Request request);
--- a/lib/src/handler_client.dart
+++ b/lib/src/handler_client.dart
@@ -5,10 +5,10 @@
 import 'dart:async';
 
 import 'base_client.dart';
+import 'handler.dart';
 import 'request.dart';
 import 'response.dart';
 
-typedef Future<Response> Handler(Request request);
 typedef void _CloseHandler();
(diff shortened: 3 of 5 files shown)


lib/src/client.dart
```diff
@@ -28,6 +29,9 @@ abstract class Client {
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
 
+  factory Client.handler(Handler handler, {CloseHandler onClose})
```
nex3: Nit: I don't like using function typedefs in parameters, because it's much harder to quickly figure out what they are. `Handler` is okay since it's such a fundamental type, but I'd write `void onClose()`.

lib/src/handler_client.dart
```diff
@@ -0,0 +1,25 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'base_client.dart';
+import 'request.dart';
+import 'response.dart';
+
+typedef Future<Response> Handler(Request request);
+typedef void CloseHandler();
```
nex3: Similarly, I wouldn't make this a public type.

lib/src/middleware.dart
```diff
@@ -0,0 +1,42 @@
...
+
+        final response = await inner.send(request);
+
+        return await responseHandler(response);
+      },
+      () {
+        closeHandler();
+
+        inner.close();
+      },
```
nex3: It's a small thing, but it's probably more efficient to directly pass in `inner.close` if `closeHandler()` isn't passed.

lib/src/middleware.dart
```diff
@@ -0,0 +1,42 @@
...
+
+typedef Client Middleware(Client inner);
+
+typedef Future<Response> ResponseHandler(Response response);
+typedef Future<Request> RequestHandler(Request request);
+
+Middleware createMiddleware({
+  RequestHandler requestHandler,
+  ResponseHandler responseHandler,
+  CloseHandler closeHandler,
```
nex3: I'd call this `onClose`, for consistency with `new Client.handler()` and other Dart APIs. It's also not *really* a handler, since it's not invoked for a request.

lib/src/middleware.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'response.dart';
+
+typedef Client Middleware(Client inner);
+
+typedef Future<Response> ResponseHandler(Response response);
+typedef Future<Request> RequestHandler(Request request);
+
+Middleware createMiddleware({
+  RequestHandler requestHandler,
+  ResponseHandler responseHandler,
```
nex3: It's probably a good idea to add an `errorHandler` here that works the same as shelf's.

lib/src/middleware.dart
```diff
@@ -0,0 +1,42 @@
...
+  CloseHandler closeHandler,
+}) {
+  requestHandler ??= (request) async => request;
+  responseHandler ??= (response) async => response;
+  closeHandler ??= () {};
+
+  return (Client inner) {
```
nex3: Don't type anonymous function parameters.

lib/src/middleware.dart
```diff
@@ -0,0 +1,42 @@
...
+    return new HandlerClient(
+          (Request request) async {
+        request = await requestHandler(request);
+
+        final response = await inner.send(request);
```
nex3: `var`

lib/src/middleware.dart
donny-dont: It would be but this ensures that close goes down through the pipeline.

lib/src/middleware.dart
donny-dont: Do you want this to be

onRequest, onResponse, onClose?

lib/src/middleware.dart
donny-dont: As an aside into performance would it be better to just have

```
Future<Request> _defaultRequest(Request request) async => request;
Future<Response> _defaultResponse(Response response) async => response;
```
etc

lib/src/middleware.dart
donny-dont: Do you want the same `HijackException` functionality?

lib/src/client.dart
donny-dont: Fixed

lib/src/handler_client.dart
donny-dont: Fixed. Would be totally rid of it but no types on the right 😞

lib/src/middleware.dart
donny-dont: Fixed

lib/src/middleware.dart
donny-dont: Fixed

lib/src/handler_client.dart
```diff
@@ -0,0 +1,26 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'base_client.dart';
+import 'request.dart';
+import 'response.dart';
+
+typedef Future<Response> Handler(Request request);
```
nex3: I'd put this in its own library. I know it seems silly when it's just one line of code, but having one file per type makes them a lot easier to track down in the source.

lib/src/middleware.dart
nex3: Not right now. We may find a compelling use-case at some point, but I think it's a lot less crucial on the client since you aren't trying to mount a WebSocket handler as part of an existing app or whatever.

lib/src/middleware.dart
nex3: No, I think `handle` is good for the first two.

lib/src/middleware.dart
nex3: > It would be but this ensures that close goes down through the pipeline.

I'm not sure what you mean by "goes down the pipeline".

> As an aside into performance would it be better to just have
> 
> ```
> Future<Request> _defaultRequest(Request request) async => request;
> Future<Response> _defaultResponse(Response response) async => response;
> ```
> etc

Nah, it would be trivial for implementations to hoist these if it's actually relevant to performance.

donny-dont: `onClose` is a `void ()` function. Within the middleware it will be unable to call `inner.close` because the `Client` is not passed to it. Also this makes things safer because the `inner.close` is always called and someone writing a middleware can't forget to call `close`.

lib/src/handler_client.dart
donny-dont: Fixed

lib/src/middleware.dart
donny-dont: Fixed

donny-dont:
@nex3 let me know if everything other than the error handler bits are looking good. If so I can start adding some documentation to it all as well.

lib/src/handler_client.dart
donny-dont: Actually if I can target 1.24 then we can get rid of the typedef since there's now a syntax for it. Let me know if its ok to do that.

nex3: Yeah, that sounds fine—1.24 will probably be out before this is ready for release.

lib/src/middleware.dart
nex3: I'm suggesting you write:

```dart
onClose: closeHandler == null
    ? inner.close
    : () {
        closeHandler();
        inner.close();
      }
```

donny-dont:
@nex3 I added some error handling any particular issues with it? It follows along with what I saw in `shelf`

lib/src/middleware.dart
donny-dont: Fixed

lib/src/handler_client.dart
donny-dont: Fixed

donny-dont:
@nex3 `HandlerClient` isn't exposed. Its only available through the factory constructor in `Client`. If you would rather expose it directly then we can do that.

nex3:
Oh, I forgot. In that case, it's fine as-is.

donny-dont:
I'll go ahead and add some documentation then this should be good to go 👍

donny-dont:
@nex3 docs look alright?

lib/src/client.dart
```diff
@@ -30,6 +30,13 @@ abstract class Client {
   /// [BrowserClient] if `dart:html` is available.
   factory Client() => new IOClient();
 
+  /// Creates a new client handler.
```
nex3: Maybe "Creates a new `Client` from a [handler] callback." The client itself isn't a handler in the type-system sense, so we should avoid implying that in the docs.

lib/src/client.dart
```diff
@@ -30,6 +30,13 @@ abstract class Client {
...
+  ///
+  /// The [handler] is a function that receives a [Request] and returns a
+  /// [Future<Response>]. It will be called when [Client.send] is invoked. This
+  /// allows for composition of a [Client] within a larger application.
```
nex3: This last sentence is confusing. It's not clear to me why this constructor in particular makes composition easier.

lib/src/handler.dart
```diff
@@ -7,4 +7,9 @@ import 'dart:async';
 import 'request.dart';
 import 'response.dart';
 
+/// The signature of a function which handles a [Request] and returns a
+/// [Future<Response>].
+///
+/// A [Handler] may receive a request directly for a HTTP resource or it may be
+/// composed as part of a larger application.
```
nex3: It's not clear what it means to receive a request "directly". Maybe, "A [Handler] may call an underlying HTTP implementation, or it may wrap another [Handler] or a [Client]."

lib/src/handler_client.dart
```diff
@@ -9,15 +9,26 @@ import 'handler.dart';
 import 'request.dart';
 import 'response.dart';
 
+/// A [Handler] based HTTP client.
```
nex3: nit: "[Handler] based" -> "[Handler]-based"

lib/src/handler_client.dart
```diff
@@ -9,15 +9,26 @@ import 'handler.dart';
...
   final Handler _handler;
   final void Function() _close;
 
+  /// Creates a new client using the [_handler] and [onClose] functions.
   HandlerClient(this._handler, void onClose())
       : _close = onClose;
 
+  /// Sends an HTTP request and asynchronously returns the response.
+  ///
+  /// The [Handler] function will be called when this method is invoked.
```
nex3: I don't think this line is necessary, especially since you can't directly refer to the values. Same for the similar line in `close()`.

nex3:
Thanks!!

PR #87, opened 2017-06-22
Add Pipeline implementation

donny-dont:
Adds the `Pipeline` class and associated tests.

### DIFF
--- a/lib/src/pipeline.dart
+++ b/lib/src/pipeline.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'client.dart';
+import 'middleware.dart';
+
+/// A helper that makes it easy to compose a set of [Middleware] and a
+/// [Client].
+///
+///     var client = const Pipeline()
+///         .addMiddleware(loggingMiddleware)
+///         .addMiddleware(basicAuthMiddleware)
+///         .addHandler(new Client());
+class Pipeline {
+  final Pipeline _parent;
+  final Middleware _middleware;
+
+  const Pipeline()
+      : _parent = null,
+        _middleware = null;
+
+  Pipeline._(this._parent, this._middleware);
+
+  /// Returns a new [Pipeline] with [middleware] added to the existing set of
+  /// [Middleware].
+  ///
+  /// [middleware] will be the last [Middleware] to process a request and
+  /// the first to process a response.
+  Pipeline addMiddleware(Middleware middleware) =>
+      new Pipeline._(this, middleware);
+
+  /// Returns a new [Client] with [client] as the final processor of a
+  /// [Request] if all of the middleware in the pipeline have passed the request
+  /// through.
+  Client addClient(Client client) =>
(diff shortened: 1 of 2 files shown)


donny-dont:
This is dependent on #85 

Enabling tests is dependent on https://github.com/dart-lang/package_resolver/issues/4

donny-dont:
@nex3 trying to get some more parallelism.

lib/src/pipeline.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'client.dart';
+import 'middleware.dart';
+
+/// A helper that makes it easy to compose a set of [Middleware] and a
+/// [Client].
+///
+///     var client = const Pipeline()
+///         .addMiddleware(loggingMiddleware)
+///         .addMiddleware(basicAuthMiddleware)
+///         .addHandler(new Client());
```
nex3: `addClient()`

It may make sense to *also* add an `addHandler()` method, though, which automatically creates a `Client`.

lib/src/pipeline.dart
```diff
@@ -0,0 +1,41 @@
...
+class Pipeline {
+  final Pipeline _parent;
+  final Middleware _middleware;
```
nex3: Please document these.

test/pipeline_test.dart
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:http/http.dart';
+import 'package:test/test.dart';
+// \TODO REMOVE
```
nex3: ??

test/pipeline_test.dart
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:http/http.dart';
+import 'package:test/test.dart';
```
nex3: `package:` imports for other packages should be separate from `package:` imports for the current package.

test/pipeline_test.dart
```diff
@@ -0,0 +1,113 @@
...
+// \TODO REMOVE
+import 'package:http/src/pipeline.dart';
+
+void main() {
+  test('compose middleware with Pipeline', () async {
+    int accessLocation = 0;
```
nex3: Don't type-annotate local variables.

test/pipeline_test.dart
donny-dont: I was doing the pull requests all in tandem so I didn't want to end up with a conflict with the exports. Will fix with a rebase.

test/pipeline_test.dart
donny-dont: Ok so from here on out you want the order to be

* Other package dependencies
* This package dependencies
* Relative imports

Wasn't the case with the previous test files but I can modify accordingly

nex3: Yes; see [the style guide](https://www.dartlang.org/guides/language/effective-dart/style#ordering).

donny-dont: 👍

PR #90, opened 2017-06-22
Add tests for Message

donny-dont:
This adds tests for the `Message` class and begins the migration from `unittest` to `test`.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,6 +10,9 @@ dependencies:
   path: ">=0.9.0 <2.0.0"
   stack_trace: ">=0.9.1 <2.0.0"
 dev_dependencies:
-  unittest: ">=0.9.0 <0.12.0"
+  test: "^0.12.18"
+# Override dependency on package_resolver to enable test
+dependency_overrides:
+  package_resolver: '^1.0.0'
 environment:
   sdk: ">=1.24.0-dev.0.0 <2.0.0"
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -14,8 +14,7 @@ export 'src/base_client.dart';
 export 'src/client.dart';
 export 'src/exception.dart';
 export 'src/io_client.dart';
-export 'src/multipart_file.dart';
-export 'src/multipart_request.dart';
+export 'src/middleware.dart';
 export 'src/request.dart';
 export 'src/response.dart';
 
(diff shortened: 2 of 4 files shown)


donny-dont:
@nex3 the tests were straight up copied from `shelf` with modifications to account for any differences between `http` and `shelf`. Also added expectations for the value of `Message.isEmpty` which weren't present in `shelf`. If any aren't applicable I'll remove.

Only other question was around `contentLength`. So in `shelf` it'll automatically get set to `0` if body is `null`. So wondering if we need to do anything to `Message.isEmpty`.

pubspec.yaml
```diff
@@ -1,15 +1,18 @@
...
+  path: '>=0.9.0 <2.0.0'
+  stack_trace: '>=0.9.1 <2.0.0'
 dev_dependencies:
-  unittest: ">=0.9.0 <0.12.0"
-environment:
-  sdk: ">=1.24.0-dev.0.0 <2.0.0"
+  test: '^0.12.18'
+# Override dependency on package_resolver to enable test
+dependency_overrides:
+  package_resolver: '^1.0.0'
```
nex3: What's with all the formatting changes?

donny-dont:
```
  /// If `true`, the stream returned by [read] won't emit any bytes.
  ///
  /// This may have false negatives, but it won't have false positives.
  bool get isEmpty => _body.contentLength == 0;
```

There isn't currently a check for `null` its just checking for length being 0.

pubspec.yaml
donny-dont: I can just move it all back to using `"`

donny-dont: Reverted

pubspec.yaml
```diff
@@ -10,6 +10,9 @@ dependencies:
   path: ">=0.9.0 <2.0.0"
   stack_trace: ">=0.9.1 <2.0.0"
 dev_dependencies:
-  unittest: ">=0.9.0 <0.12.0"
+  test: "^0.12.18"
+# Override dependency on package_resolver to enable test
+dependency_overrides:
+  package_resolver: '^1.0.0'
```
nex3: It might be nice to add some newlines to make the pubspec a bit more readable.

donny-dont: At this point I was figuring we just fix it up when releasing 0.12.0 into the wild

nex3: Sure.

PR #93, opened 2017-06-23
Remove MockClient

donny-dont:
Mocking a HTTP connection can be done using a `Handler`.


donny-dont:
@nex3 I don't feel this is needed anymore but wanted your opinion.

If we remove I think we should add a section to the README about how to test.

nex3:
> I don't feel this is needed anymore but wanted your opinion.

Yeah this is redundant now that `new Client.handler()` exists.

> If we remove I think we should add a section to the README about how to test.

That's a good idea.

PR #95, opened 2017-06-23
Adding client tests

donny-dont:
This adds a `dart:html` and `dart:io` compatible `Client` test. A server is added for the clients to call.


donny-dont:
@nex3 this adds the tests but some tests fail.

On the `dart:io` side
* Doesn't appear that any bytes are getting transferred over as `body` is empty when retrieved
* The `content-length` header isn't getting set

On the `dart:html` side
* The `content-length` is set automatically by the browser, but it also appends `content-length` to methods other than `GET` and `HEAD`

For both
* `content-type` is not getting set.
* We left out the `x-www-form-urlencoded` implementation

Anyways I'd like to land this and fix the issues individually.

@kevmoo that means that tests will be failing until the above is resolved.

test/client_test.dart
```diff
@@ -0,0 +1,410 @@
...
+  group('client', () {
+    // The server url of the spawned server
+    var serverUrl;
+
+    setUp(() async {
+      var channel = spawnHybridUri('hybrid/server.dart');
+      serverUrl = Uri.parse(await channel.stream.first);
+    });
+
+    // HEAD
```
nex3: These headers seem redundant with the groups.

test/client_test.dart
```diff
@@ -0,0 +1,410 @@
...
+            'headers': {
+              'user-agent': [userAgent()],
+              'x-random-header': ['Value'],
+              'x-other-header': ['Other Value']
+            }
+          })));
+    });
+
+    test('readBytes throws an error for a 4** status code', () {
+      platformClient().readBytes(serverUrl.resolve('/error')).then((_) {
```
nex3: Use async/await.

test/hybrid/client.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
nex3: I don't understand why this is under a "hybrid" directory.

test/hybrid/server.dart
```diff
@@ -0,0 +1,120 @@
...
+  'connection',
+  'origin',
+  'referer',
+
+  // Dart IO headers
+  'cookie',
+  'host',
+];
+
+hybridMain(StreamChannel channel) async {
```
nex3: Document this. In particular, mention which tests it's used in and what the server does.

test/client_test.dart
donny-dont: Is there a specific way I should be using one of the `expect` functions? I wasn't able to get it to work otherwise.

nex3: Looks like you figured it out.

PR #96, opened 2017-06-23
Enable Travis-CI

kevmoo:


kevmoo:
@nex3 – made this green. At least a starting point for @donny-dont

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - stable
+
+dart_task:
+  - test
+  - dartfmt
+  - dartanalyzer
+
+matrix:
+  allow_failures:
+  - dart_task: dartanalyzer
```
nex3: Why?

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+sudo: false
+
+dart:
+  - dev
+  - stable
+
+dart_task:
+  - test
+  - dartfmt
+  - dartanalyzer
```
nex3: Only run these once each. See [the example file](https://github.com/dart-lang/oss_project_templates/blob/master/dart-github-dart-lang/.travis.yml).

.travis.yml
kevmoo: Because there are still files that break analysis – specifically the multi_file things in the lib directory that aren't cleaned up yet

.travis.yml
kevmoo: ack

.travis.yml
donny-dont: Do we want to just allow failures at this point and then when everything is green then protect the branch?

donny-dont:
@kevmoo shoukd we cherry pick some of the 0.11.x commits before doing the dartfmt?

.travis.yml
nex3: I don't think it makes sense to allow failures that are being actively worked on. I'd just let the bots be red until they go green.

kevmoo:
> @kevmoo should we cherry pick some of the 0.11.x commits before doing the dartfmt?

Uh...not sure what needs to be pulled in.

I can just enable tests for now – okay w/ you @nex3 ?

nex3:
I think Don's right... we shouldn't reformat before we merge stuff in from 0.11.x, or the conflicts will get out of control. Kevin, can you remove the formatting stuff for now?

kevmoo:
PTAL

donny-dont:
@kevmoo can we add whatever browser testing the dart team requires? The tests should run in both.

nex3:
Yeah, we should run tests on (at least) the VM and Firefox.

kevmoo:
PTAL

.travis.yml
```diff
@@ -0,0 +1,29 @@
+language: dart
+sudo: false
+
+dart:
+  - dev
+  - stable
+
+dart_task:
+  - test: --platform vm
+  - test: --platform firefox
```
nex3: There aren't very many tests in this package, so it'll probably be faster if we run the VM and Firefox in the same task.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+  - dartfmt
+  - dartanalyzer
```
nex3: Since these are defined in the matrix section, they shouldn't be included here.

kevmoo:
PTAL

PR #102, opened 2017-06-26
Adding Request tests

donny-dont:
This adds back relevant `Request` tests from `http` as well as including some tests from `Request` in `shelf` to match the new implementation.


donny-dont:
@nex3 let me know if there are other tests that should be present or if any of those tests are no longer needed.

After we add these and some for `Response` its time to fix them so everything runs as expected.

donny-dont:
Actually there's only a couple properties that are on `Response` that are not present on `Request`.

Issue #105, opened 2017-07-20
single quote char (ASCII 39) causing failures with GCS


kevmoo:
Seems this is the root cause of the following

https://github.com/dart-lang/pub/issues/1666
https://github.com/dart-lang/pub-dartlang-dart/issues/230

Solution: remove single quote from set of considered boundary characters

kevmoo:
Can't figure out what this change https://github.com/dart-lang/sdk/commit/ebffa49eac88627181233fbb2a29482eb5c72c66

Which included this range of changes https://github.com/dart-lang/http/compare/0.11.3+9...0.11.3+13 would cause this behavior to be exposed

PR #106, opened 2017-07-20
Remove `'` from boundary characters

kevmoo:
Fixes https://github.com/dart-lang/http/issues/105


kevmoo:
CC @devoncarew

devoncarew:
👍 good catch

PR #111, opened 2017-08-11
Add http package prefix for import consistency

perlatus:
In the example, `Client` and `BaseClient` are package qualified, but `BaseRequest` and `StreamedResponse` aren't.


nex3:
Thanks!

PR #113, opened 2017-09-21
Implement multipart forms

### LINKED ISSUE #97: Fix Multipart support
CC @donny-dont

donny-dont:
Implement multipart form handling.

Fixes #97


donny-dont:
@nex3 currently work in progress for the implementation. I have tests running locally but won't push that until we figure out if this is what you want.

Another option rather than implementing `Body` is to move a constructor over there and keep the multipart file handling in there or in a utils sort of file with free functions.

lib/src/multipart_file.dart
```diff
@@ -44,20 +40,20 @@ class MultipartFile {
   /// future may be inferred from [filename].
   MultipartFile(this.field, Stream<List<int>> stream, this.length,
```
nex3: What do you think about consolidating these constructors? I'm thinking:

* `new MultipartFile(String field, body, ...)` which takes either a `List<int>` or a `String`,
* `new MultipartFile.fromStream(String field, stream, int length)` which takes either a `Stream<List<int>>` or a `Stream<String>`,
* `static Future<MultipartFile> loadStream(String field, stream)` which takes either a  `Stream<List<int>>` or a `Stream<String>` and drains it to determine the length.

lib/src/multipart_file.dart
```diff
@@ -92,20 +87,21 @@ class MultipartFile {
...
+        filename: filename, contentType: contentType);
   }
 
-  // Finalizes the file in preparation for it being sent as part of a
-  // [MultipartRequest]. This returns a [ByteStream] that should emit the body
-  // of the file. The stream may be closed to indicate an empty file.
-  ByteStream finalize() {
-    if (isFinalized) {
-      throw new StateError("Can't finalize a finalized MultipartFile.");
+  /// Returns a [Stream] representing the file.
```
nex3: "the file" -> "the contents of the file"

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
```
nex3: This'll need to be renamed to `multipart_body.dart`.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
+  /// Multipart forms do not have an encoding.
+  Encoding get encoding => null;
+
+  /// Creates a [MultipartBody] from the given [fields] and [files].
+  ///
+  /// The [boundary] is used to
+  factory MultipartBody(
+    Map<String, String> fields,
+    List<MultipartFile> files,
+    String boundary,
```
nex3: Nit: remove this trailing comma.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
       writeUtf8(value);
       writeLine();
     });
 
-    Future.forEach(_files, (file) {
-      writeAscii('--$boundary\r\n');
-      writeAscii(_headerForFile(file));
-      return writeStreamToSink(file.finalize(), controller)
-        .then((_) => writeLine());
+    // Iterate over the files to get the length
```
nex3: Nit: trailing period, here and below.

lib/src/request.dart
```diff
@@ -130,6 +133,18 @@ class Request extends Message {
       {Map<String, String> headers, Map<String, Object> context})
       : this('DELETE', url, headers: headers, context: context);
 
+  factory Request.multipart(url, {Map<String, String> headers, Map<String, Object> context, Map<String, String> fields, List<MultipartFile> files,}) {
```
nex3: `files` should be an `Iterable` here.

Nit: no trailing comma.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
+      contentLength += string.length;
+    }
+    void writeUtf8(String string) {
+      var encoded = UTF8.encode(string);
+      controller.add(encoded);
+      contentLength += encoded.length;
+    }
+    void writeLine() {
+      controller.add([13, 10]); // \r\n
+      contentLength += 2;
```
nex3: Since you're only using these for fields now, you might as well just add everything to a single `Uint8List` rather than adding separate chunks to a controller. The `typed_data` package can help with that.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
+    var fileHeaders = new List<List<int>>(fileCount);
+
+    for (var i = 0; i < fileCount; ++i) {
+      var file = files[i];
+      var header = <int>[];
+      header.addAll('--$boundary\r\n'.codeUnits);
+      header.addAll(UTF8.encode(_headerForFile(file)));
+      contentLength += header.length + file.length + 2;
+      fileHeaders[i] = header;
+    }
```
nex3: Add a comment explaining that we compute the headers ahead of time so we can synchronously access the length.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
+    var ending = '--$boundary--\r\n'.codeUnits;
+    contentLength += ending.length;
+
+    // Write the files to the stream
+    //
+    // Future.forEach will ensure that the actions happen in sequence so i is
+    // used to get the fileHeaders
+    var i = 0;
+
+    Future.forEach(files, (file) {
```
nex3: Consider pulling this into a static method so you can make it async and just use a normal `for` loop.

lib/src/multipart_request.dart
```diff
@@ -4,162 +4,143 @@
...
-  /// Encode [value] in the same way browsers do.
-  String _browserEncode(String value) {
-    // http://tools.ietf.org/html/rfc2388 mandates some complex encodings for
-    // field names and file names, but in practice user agents seem not to
-    // follow this at all. Instead, they URL-encode `\r`, `\n`, and `\r\n` as
-    // `\r\n`; URL-encode `"`; and do nothing else (even for `%` or non-ASCII
-    // characters). We follow their behavior.
-    return value.replaceAll(_newlineRegExp, "%0D%0A").replaceAll('"', "%22");
+/// Returns the header string for a field.
+String _headerForField(String name, String value) {
```
nex3: I generally make private methods that are only used by one class members of that class, even if they don't access its state directly.

lib/src/request.dart
```diff
@@ -130,6 +133,18 @@ class Request extends Message {
...
+    fields ??= <String, String>{};
+    files ??= <MultipartFile>[];
+
+    var boundary = boundaryString();
+
+    headers ??= <String, String>{};
+    headers['content-type'] = 'multipart/form-data; boundary=$boundary';
```
nex3: We shouldn't destructively modify the user's `headers` map.

lib/src/multipart_request.dart
donny-dont: These are accessed from a factory constructor so I put them below. Do you want them as static class methods?

nex3: Yes.

lib/src/multipart_file.dart
donny-dont: I've always kinda preferred the explicit constructors rather than using `dynamic`  since there are no union types. I do like the `loadStream` idea rather than `fromPath`.

Would `fromStream` be possible with a `Stream<String>` though as there would need to be some kind of encoding?

donny-dont:
@nex3 https://pub.dartlang.org/packages/mime yay or nay?

lib/src/request.dart
```diff
@@ -130,6 +133,28 @@ class Request extends Message {
       {Map<String, String> headers, Map<String, Object> context})
       : this('DELETE', url, headers: headers, context: context);
 
+  factory Request.multipart(url,
```
nex3: Needs docs

lib/src/request.dart
```diff
@@ -130,6 +133,28 @@ class Request extends Message {
...
+        Map<String, String> fields,
+        Iterable<MultipartFile> files}) {
+    fields ??= <String, String>{};
+    files ??= <MultipartFile>[];
+    headers ??= <String, String>{};
+
+    var boundary = _boundaryString();
+
+    return new Request._(
+        'POST',
```
nex3: Technically, multipart requests don't always have to be POST. They are the vast majority of the time, so it's a good default, but we should probably add an optional `method` parameter.

lib/src/request.dart
```diff
@@ -130,6 +133,28 @@ class Request extends Message {
...
       : this('DELETE', url, headers: headers, context: context);
 
+  factory Request.multipart(url,
+      {Map<String, String> headers,
+        Map<String, Object> context,
+        Map<String, String> fields,
+        Iterable<MultipartFile> files}) {
+    fields ??= <String, String>{};
+    files ??= <MultipartFile>[];
+    headers ??= <String, String>{};
```
nex3: Doesn't context need a default value?

lib/src/request.dart
```diff
@@ -130,6 +133,28 @@ class Request extends Message {
       {Map<String, String> headers, Map<String, Object> context})
       : this('DELETE', url, headers: headers, context: context);
 
+  factory Request.multipart(url,
+      {Map<String, String> headers,
+        Map<String, Object> context,
+        Map<String, String> fields,
+        Iterable<MultipartFile> files}) {
+    fields ??= <String, String>{};
```
nex3: You shouldn't need explicit type parameters here; inference should take care of that.

Also, you might as well make these const to avoid unnecessary allocations.

lib/src/request.dart
```diff
@@ -130,6 +133,28 @@ class Request extends Message {
...
+    headers ??= <String, String>{};
+
+    var boundary = _boundaryString();
+
+    return new Request._(
+        'POST',
+        url,
+        new MultipartBody(fields, files, boundary),
+        null,
+        updateMap(headers, <String, String>{
```
nex3: This type annotation also shouldn't be necessary.

lib/src/request.dart
```diff
@@ -163,4 +188,39 @@ class Request extends Message {
...
+  ///
+  /// This is the intersection of the characters allowed in the `bcharsnospace`
+  /// production defined in [RFC 2046][] and those allowed in the `token`
+  /// production defined in [RFC 1521][].
+  ///
+  /// [RFC 2046]: http://tools.ietf.org/html/rfc2046#section-5.1.1.
+  /// [RFC 1521]: https://tools.ietf.org/html/rfc1521#section-4
+  static const List<int> _boundaryCharacters = const <int>[
+    43, 95, 45, 46, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 65, 66, 67, 68, //
+    69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, //
```
nex3: I think you only need the first `//` here.

lib/src/request.dart
```diff
@@ -163,4 +188,39 @@ class Request extends Message {
...
+
+  /// Returns a randomly-generated multipart boundary string
+  static String _boundaryString() {
+    var prefix = "dart-http-boundary-";
+    var list = new List<int>.generate(_boundaryLength - prefix.length,
+            (index) =>
+        _boundaryCharacters[_random.nextInt(_boundaryCharacters.length)],
+        growable: false);
+    return "$prefix${new String.fromCharCodes(list)}";
+  }
```
nex3: These seem out of place in the main request.dart file. Maybe move them into a boundary.dart file that just exposes `boundaryString()`?

pubspec.yaml
```diff
@@ -8,6 +8,7 @@ dependencies:
   collection: "^1.5.0"
   http_parser: ">=0.0.1 <4.0.0"
   path: ">=0.9.0 <2.0.0"
+  typed_data: "^1.1.5"
```
nex3: You should generally try to make the lower bound of constraints as low as possible while they still provide the APIs you're using, so that the version solver has more flexibility. In this case, `^1.0.0` should be fine.

test/utils.dart
```diff
@@ -66,6 +67,40 @@ class _Parse extends Matcher {
   }
 }
 
+/// A matcher that validates the body of a multipart request after finalization.
```
nex3: Nit: First sentence should be its own paragraph.

test/utils.dart
```diff
@@ -66,6 +67,40 @@ class _Parse extends Matcher {
...
+/// The string "{{boundary}}" in [pattern] will be replaced by the boundary
+/// string for the request, and LF newlines will be replaced with CRLF.
+/// Indentation will be normalized.
+Matcher bodyMatches(String pattern) => new _BodyMatches(pattern);
```
nex3: This name is pretty generic for something that's specifically designed to work with multipart requests. Maybe `multipartBodyMatches()`?

test/utils.dart
```diff
@@ -66,6 +67,40 @@ class _Parse extends Matcher {
...
+
+class _BodyMatches extends Matcher {
+  final String _pattern;
+
+  _BodyMatches(this._pattern);
+
+  bool matches(item, Map matchState) {
+    if (item is! http.Request) return false;
+
+    var future = collectBytes(item.read()).then((bodyBytes) {
```
nex3: Why not just use `Request.readAsString()` here?

test/multipart_test.dart
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
```
nex3: 2017

test/multipart_test.dart
```diff
@@ -0,0 +1,249 @@
...
+    expect(request, bodyMatches('''
+        --{{boundary}}
+        content-type: application/octet-stream
+        content-disposition: form-data; name="file"
+
+        hello
+        --{{boundary}}--
+        '''));
+  });
+}
```
nex3: Are these ported directly from the old API?

test/utils.dart
donny-dont: Because of this check. The string length isn't necessarily the same as the contentLength at this point.

```dart
expect(item.contentLength, equals(bodyBytes.length));
```

test/multipart_test.dart
donny-dont: Yep they're from the 0.11.x branch.

pubspec.yaml
donny-dont: Just for my own reference are we going to want to tighten the restriction for strong mode compliance?

lib/src/request.dart
donny-dont: It will end up getting set when a `HttpUnmodifiableMap` is created within the `Message` constructor so no it doesn't need one.

lib/src/multipart_file.dart
```diff
@@ -4,108 +4,105 @@
...
-
-    return new MultipartFile.fromBytes(field, encoding.encode(value),
-        filename: filename,
-        contentType: contentType);
+    if (encoding != null) {
+      contentType = contentType.change(parameters: {'charset': encoding.name});
+    }
+
+    return new MultipartFile.fromStream(
+        field, new Stream<List<int>>.fromIterable([bytes]), bytes.length,
```
nex3: Nit: you shouldn't need type parameters here.

lib/src/multipart_file.dart
```diff
@@ -4,108 +4,105 @@
...
-    _isFinalized = true;
-    return _stream;
+    var stream = _stream;
+    _stream = null;
+    return stream;
+  }
+
+  /// Looks up the [MediaType] from the [filename]'s extension or from
+  /// magic numbers contained within a file header's [bytes].
+  static MediaType _lookupMediaType(String filename, [List<int> bytes]) {
```
nex3: Nit: `_lookUpMediaType`

"Lookup" is a word, but it's only a noun; the verb is always two words.

lib/src/multipart_file.dart
```diff
@@ -4,108 +4,105 @@
...
+    if ((filename == null) && (bytes == null)) return null;
```
nex3: Nit: unnecessary parens.

lib/src/multipart_file.dart
```diff
@@ -4,108 +4,105 @@
...
+
+    // lookupMimeType expects filename to be non-null but its possible that
+    // this can be called with bytes but no filename.
```
nex3: Can you file an issue against the mime package for this?

lib/src/request.dart
```diff
@@ -130,11 +131,43 @@ class Request extends Message {
       {Map<String, String> headers, Map<String, Object> context})
       : this('DELETE', url, headers: headers, context: context);
 
-  Request._(this.method, this.url,
-      body,
-      Encoding encoding,
+  /// Creates a new `multipart/form-data` [Request] to [url], which can be a
```
nex3: Consider linking `multipart/form-data` to an explanation of the format (maybe Wikipedia).

pubspec.yaml
```diff
@@ -7,7 +7,8 @@ dependencies:
   async: "^1.13.0"
   collection: "^1.5.0"
   http_parser: ">=0.0.1 <4.0.0"
-  path: ">=0.9.0 <2.0.0"
+  mime: ">=0.9.0 < 1.0.0"
```
nex3: `^0.9.0`

With this version constraint, `mime` could release a totally breaking `0.10.0` release and we'd still declare compatibility with it.

nex3:
Looks good! Thanks for all your hard work on this :smile:.

PR #115, opened 2017-09-21
Stop depending on stack_trace

nex3:
At this point, we're only using this for Chain.current which might as
well just be StackTrace.current.


kevmoo:
This might be worth back-porting to 0.11 – good for all of our users...

nex3:
I'd rather avoid porting anything to 0.11 unless it's urgent. 1.0 is almost ready for release.

PR #118, opened 2017-09-22
Fix the content-type header for the Body of a Message

donny-dont:
This corrects the `content-type` header specified by a `Message`.

If its okay to remove the tests that are now being skipped, as they were originally for `shelf` then I'll push that before merging.


donny-dont:
Closing as it modified Body which was not desired.

PR #120, opened 2017-09-22
Fix content type within Message

### LINKED ISSUE #101: Fix url encoded requests
During the migration we lost support for `application/x-www-form-urlencoded`. Wanted to know if there is any preference on how to fix this @nex3 .

I'm thinking it might make sense to have another constructor for Request.

donny-dont:
Attempt to fix `content-type` within `Message`.

Fixes #101

### DIFF
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -48,6 +48,8 @@ class Body {
     Stream<List<int>> stream;
     int contentLength;
 
+    encoding ??= UTF8;
+
     if (body is Map) {
       body = mapToQuery(body, encoding: encoding);
     }
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -29,6 +29,9 @@ class Body {
   /// determined efficiently.
   final int contentLength;
 
+  /// An empty stream for use with empty bodies.
+  static const _emptyStream = const Stream.empty();
+
   Body._(this._stream, this.encoding, this.contentLength);
 
   /// Converts [body] to a byte stream and wraps it in a [Body].
@@ -38,8 +41,9 @@ class Body {
   /// used to convert it to a [Stream<List<int>>].
   factory Body(body, [Encoding encoding]) {
     if (body is Body) return body;
-    if (body == null)
-      return new Body._(new Stream.fromIterable([]), encoding, 0);
+    if (body == null) {
+      return new Body._(_emptyStream, encoding, 0);
+    }
 
     Stream<List<int>> stream;
     int contentLength;
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -8,6 +8,7 @@ import 'dart:convert';
 import 'package:async/async.dart';
 import 'package:collection/collection.dart';
 
+import 'empty_body.dart';
 import 'utils.dart';
 
 /// The body of a request or response.
@@ -42,7 +43,7 @@ class Body {
   factory Body(body, [Encoding encoding]) {
     if (body is Body) return body;
     if (body == null) {
-      return new Body._(_emptyStream, encoding, 0);
+      return const EmptyBody();
     }
 
     Stream<List<int>> stream;
@@ -83,4 +84,13 @@ class Body {
     _stream = null;
     return stream;
   }
+}
+
+class _EmptyBody implements Body {
+  const _EmptyBody();
+
+  Encoding get encoding => UTF8;
+  int get contentLength => 0;
+
+  Stream<List<int>> read() => const Stream<List<int>>.empty();
 }
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -8,6 +8,8 @@ import 'dart:convert';
 import 'package:async/async.dart';
 import 'package:collection/collection.dart';
 
+import 'utils.dart';
+
 /// The body of a request or response.
 ///
 /// This tracks whether the body has been read. It's separate from [Message]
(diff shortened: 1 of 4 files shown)


donny-dont:
@nex3 if this is the approach you want I'll comment up the functions/methods before merging.

donny-dont:
@nex3 so I made `change` actually look at the value of `body` to determine if it was set during the change. This makes it so you can skip the internal headers adjustment which was something I was trying to avoid.

lib/src/message.dart
```diff
@@ -59,16 +58,30 @@ abstract class Message {
...
-      Map<String, String> headers,
-      Map<String, Object> context})
-      : this._(new Body(body, encoding), headers, context);
+        Map<String, String> headers,
+        Map<String, Object> context})
+      : this.__(body, _determineMediaType(body, encoding, headers), headers,
+      context);
 
-  Message._(Body body, Map<String, String> headers, Map<String, Object> context)
+  Message.__(body, MediaType contentType, Map<String, String> headers,
```
nex3: `__` isn't a good name for a constructor--it doesn't communicate what the difference is between this and the other unnamed constructor.

lib/src/message.dart
```diff
@@ -59,16 +58,30 @@ abstract class Message {
...
+      Map<String, Object> context)
+      : this._(new Body(body, encodingForMediaType(contentType, null)), contentType,
```
nex3: This `new Body(...)` invocation looks wrong to me... the `Body`'s encoding shouldn't ever come from the headers, it should only come from the explicit encoding passed by the user. So this should continue to be `new Body(body, encoding)`, which means you don't have to do this many-layered-constructor thing anymore.

lib/src/message.dart
```diff
@@ -59,16 +58,30 @@ abstract class Message {
...
+
+  Message._(Body body, MediaType contentType, Map<String, String> headers,
+      Map<String, Object> context)
+      : this.fromValues(body, _adjustHeaders(headers, body, contentType), context);
+
+  /// Creates a new [Message].
+  ///
+  /// This constructor should be used when no computation is required for the
+  /// [body], [headers] or [context].
+  Message.fromValues(
```
nex3: I understand the performance motivation for having this constructor, but I'd like to focus on API shape and correctness for now and leave optimizations like this for later (ideally driven by real data).

lib/src/body.dart
```diff
@@ -36,25 +38,20 @@ class Body {
...
-        stream = new Stream.fromIterable([encoded]);
-      } else {
-        var encoded = encoding.encode(body);
-        contentLength = encoded.length;
-        stream = new Stream.fromIterable([encoded]);
-      }
+
+    if (body is Map) {
+      body = mapToQuery(body, encoding: encoding);
+    }
```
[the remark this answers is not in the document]
donny-dont: The value of `body` becomes a `String` at this point so it will end up in the `is String` which would then encode it.

lib/src/message.dart
donny-dont: `_determineMediaType` is what determines the encoding for the body. If `encoding` is set directly then it should override what is in `content-type` header so that value can be used when creating the body.

This reflects the behavior of this test https://github.com/dart-lang/http/blob/master/test/request_test.dart#L152-L158

lib/src/message.dart
donny-dont: Ok I'll rework this accordingly.

lib/src/body.dart
```diff
@@ -36,25 +38,20 @@ class Body {
...
+
+    if (body is String) {
```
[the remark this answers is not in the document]
donny-dont: It should never be `null` since the encoding is calculated.

lib/src/message.dart
donny-dont: K I think I need to rename this anyways so it could be used in the multipart implementation.

I also didn't like `fromValues` but didn't have any better ideas last night.

lib/src/message.dart
nex3: The way this is written currently, the encoding passed to `new Body()` can come from `headers`. That's not correct. It should only ever come from the `encoding` parameter, or from the default value (which should be UTF-8).

donny-dont: So just to be clear if someone were to set `content-type: 'application/json; charset=iso-8859-1'` and not set `encoding` to LATIN when creating a `Request` it should not encode the body with LATIN it should be UTF8 encoded?

nex3: Correct, and the header should be updated accordingly.

donny-dont:
@nex3 should be good to go. I think I addressed all your feedback.

lib/src/body.dart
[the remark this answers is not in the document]
nex3: You mean when it's set to `UTF8` above? But then you're marking ASCII-only bodies as UTF8 when a `text/plain` content type would do. The old behavior here was correct, there's no need to change it.

lib/src/message.dart
```diff
@@ -61,12 +60,27 @@ abstract class Message {
       {Encoding encoding,
       Map<String, String> headers,
       Map<String, Object> context})
-      : this._(new Body(body, encoding), headers, context);
+      : this._forwardToContentType(body, encoding,
+            _determineMediaType(body, encoding, headers), headers, context);
 
-  Message._(Body body, Map<String, String> headers, Map<String, Object> context)
+  Message.withContentType(Body body, MediaType contentType,
```
nex3: I don't want this constructor here. The shape of the constructors from before this CL is generally correct; the only change should be adding a `{bool bodyIsMap: false}` parameter to `new Message._()` and `_adjustHeaders()`.

lib/src/body.dart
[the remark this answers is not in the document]
donny-dont: Ok I'll take a look at what 0.11 was doing here.

lib/src/message.dart
donny-dont: I had planned to use it with `http.MultiPart` since the `MediaType` requires a boundary which would be randomly generated.

lib/src/body.dart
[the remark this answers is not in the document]
nex3: Not 0.11, just the previous code in master.

lib/src/message.dart
nex3: I'd rather have `_adjustHeaders()` do some logic like

```dart
if (body is MultipartBody) {
  contentType = "multipart/mixed; boundary=${body.boundary}";
}
```

lib/src/body.dart
```diff
@@ -27,6 +29,9 @@ class Body {
   /// determined efficiently.
   final int contentLength;
 
+  /// An empty stream for use with empty bodies.
+  static const _emptyStream = const Stream.empty();
```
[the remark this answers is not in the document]
donny-dont: Was looking at implementing this but the issue is that calling read removes the stream.

Do you instead want an EmptyBody class that lets you call read() multiple times?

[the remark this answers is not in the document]
nex3: > Do you instead want an EmptyBody class that lets you call read() multiple times?

Yeah, I think this is the way to go. That matches the behavior of `const Stream.empty()`, too.

PR #124, opened 2017-10-19
Add content type library

donny-dont:
Consolidate content-type related code into a single library and start using it within `Message`.

### DIFF
--- a/lib/src/content_type.dart
+++ b/lib/src/content_type.dart
@@ -1,0 +1,40 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:http_parser/http_parser.dart';
+
+/// Returns the [Encoding] that corresponds to [charset].
+///
+/// Returns [fallback] if [charset] is null or if no [Encoding] was found that
+/// corresponds to [charset].
+Encoding encodingForCharset(String charset, [Encoding fallback = LATIN1]) {
+  if (charset == null) return fallback;
+  var encoding = Encoding.getByName(charset);
+  return encoding == null ? fallback : encoding;
+}
+
+/// Returns the name of the charset that corresponds to the [contentType].
+///
+/// If [contentType] is `null` then no charset has been specified and `null`
+/// will be returned. If the [contentType] is present then it will be parsed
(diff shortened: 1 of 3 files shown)


lib/src/content_type.dart
```diff
@@ -0,0 +1,30 @@
...
+
+import 'dart:convert';
+
+import 'package:http_parser/http_parser.dart';
+
+/// Returns the [Encoding] that corresponds to [charset].
+///
+/// Returns [fallback] if [charset] is null or if no [Encoding] was found that
+/// corresponds to [charset].
+Encoding encodingForCharset(String charset, [Encoding fallback = LATIN1]) {
```
nex3: Now that we have `??`, it's probably simpler to just return `null` if the encoding isn't found rather than having `fallback` parameters.

lib/src/content_type.dart
```diff
@@ -0,0 +1,30 @@
...
+///
+/// Returns [fallback] if the charset is not specified in the [type] or if no
+/// [Encoding] was found that corresponds to the `charset`.
+Encoding encodingForMediaType(MediaType type, [Encoding fallback = LATIN1]) {
+  if (type == null) return fallback;
+  return encodingForCharset(type.parameters['charset'], fallback);
+}
+
+/// Modifies the media [type]'s [encoding].
+MediaType modifyEncoding(MediaType type, Encoding encoding) =>
```
nex3: This doesn't seem to be used anywhere, and I'm not sure it would be worth factoring out even if it were...

lib/src/message.dart
```diff
@@ -80,10 +81,11 @@ abstract class Message {
   /// If not set, `null`.
   int get contentLength {
     if (_contentLengthCache != null) return _contentLengthCache;
-    if (!headers.containsKey('content-length')) return null;
-    _contentLengthCache = int.parse(headers['content-length']);
-    return _contentLengthCache;
+    var contentLengthHeader = getHeader(headers, 'content-length');
+    if (contentLengthHeader == null) return null;
+    return _contentLengthCache = int.parse(contentLengthHeader);
```
nex3: We generally avoid using the return value of assignment expressions.

lib/src/content_type.dart
donny-dont: I was going to use it when creating the `content-type` header from a body and also its in the multipart stuff.

nex3: It just doesn't seem like it provides substantial value over just inlining the definition.

PR #127, opened 2017-10-20
Cleanup contents of util

donny-dont:
A number of functions in `util.dart` aren't used now. Those are now removed. Formatted the file accordingly.

`writeStreamToSink` and `isPlainAscii` moved into the multipart implementation at https://github.com/dart-lang/http/pull/113/commits/20971291fa510648c1ef868871d79cb6b92b0554

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -2,9 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
 import 'dart:convert';
-import 'dart:typed_data';
 
 import 'package:collection/collection.dart';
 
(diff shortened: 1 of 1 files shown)


nex3:
I'm not sure I see the logic behind moving `writeStreamToSink()` and `isPlainAscii()` out of the utils directory. They seem like pretty generic utilities to me.

donny-dont:
@nex3 they weren't being used by anything outside of `MultipartBody`. Also wasn't sure if there was something in `async` that might be a replacement for `writeStreamToSink`.

nex3:
> they weren't being used by anything outside of `MultipartBody`.

I think a better criterion for utils is "Stuff whose use is not specific to the place(s) where it's used", rather than "Stuff that's used in multiple places".

> Also wasn't sure if there was something in `async` that might be a replacement for `writeStreamToSink`.

I don't know of anything that specifically writes *without* closing the sink.

PR #129, opened 2017-10-25
Add User-Agent to CORS headers within the test server

donny-dont:
This fixes running `client_test.dart` in Firefox.


nex3:
Nice detective work!

PR #132, opened 2017-10-26
Documentation cleanup

donny-dont:
* Remove docs from overridden methods.
* Remove references to old functionality.
* Use 1 line description.

### DIFF
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -20,10 +20,9 @@ import 'src/response.dart';
 /// XMLHttpRequests.
 ///
 /// This client inherits some of the limitations of XMLHttpRequest. It is
-/// unable to directly set some headers, such as `content-length`. There is no
-/// way to create a persistent connection. Redirects cannot be handled directly.
-/// It is also unable to stream requests or responses; a request will only be
-/// sent and a response will only be returned once all the data is available.
+/// unable to directly set some headers, such as `content-length`. It is also
+/// unable to stream requests or responses; a request will only be sent and a
+/// response will only be returned once all the data is available.
 class BrowserClient extends BaseClient {
   /// The currently active XHRs.
   ///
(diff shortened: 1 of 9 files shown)


lib/browser_client.dart
```diff
@@ -19,11 +19,11 @@ import 'src/response.dart';
...
 /// XMLHttpRequests.
 ///
-/// This client inherits some of the limitations of XMLHttpRequest. It ignores
-/// the [BaseRequest.contentLength], [BaseRequest.persistentConnection],
-/// [BaseRequest.followRedirects], and [BaseRequest.maxRedirects] fields. It is
-/// also unable to stream requests or responses; a request will only be sent and
-/// a response will only be returned once all the data is available.
+/// This client inherits some of the limitations of XMLHttpRequest. It is
+/// unable to directly set some headers, such as `content-length`. There is no
+/// way to create a persistent connection. Redirects cannot be handled directly.
```
nex3: I'd just remove the sentences about persistent connections and redirects.

lib/http.dart
```diff
@@ -28,7 +28,8 @@ export 'src/response.dart';
 /// the request is complete. If you're planning on making multiple requests to
 /// the same server, you should use a single [Client] for all of those requests.
 ///
-/// For more fine-grained control over the request, use [Request] instead.
+/// For more fine-grained control over the request and response, create a
+/// [Client] and use it directly.
```
nex3: Now that these methods support streaming requests and responses, I don't think we need these notes anymore.

lib/src/base_client.dart
```diff
@@ -13,136 +13,46 @@ import 'exception.dart';
 import 'request.dart';
 import 'response.dart';
 
-/// The abstract base class for an HTTP client. This is a mixin-style class;
-/// subclasses only need to implement [send] and maybe [close], and then they
-/// get various convenience methods for free.
+/// The abstract base class for an HTTP client.
+///
+/// This is a mixin-style class; subclasses only need to implement [send] and
```
nex3: "Mixin-style" is confusing now that Dart has actual mixins. I'd just remove that bit.

PR #133, opened 2017-10-27
Modify context variable naming

donny-dont:
Updated naming scheme and wrote some documentation around it.

### DIFF
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -25,10 +25,12 @@ import 'src/response.dart';
 /// response will only be returned once all the data is available.
 ///
 /// [BrowserClient] allows setting values directly on the underlying
-/// [HttpRequest] values through [Request.context].
+/// [HttpRequest] values through [Request.context] :
 ///
-/// * `http.html.with_credentials` is a boolean value, defaulting to `false`
-///   that corresponds to [HttpRequest.withCredentials].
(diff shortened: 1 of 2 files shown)


lib/browser_client.dart
```diff
@@ -23,6 +23,12 @@ import 'src/response.dart';
 /// unable to directly set some headers, such as `content-length`. It is also
 /// unable to stream requests or responses; a request will only be sent and a
 /// response will only be returned once all the data is available.
+///
+/// [BrowserClient] allows setting values directly on the underlying
+/// [HttpRequest] values through [Request.context].
+///
+/// * `http.html.with_credentials` is a boolean value, defaulting to `false`
+///   that corresponds to [HttpRequest.withCredentials].
```
nex3: I'd rather document these in terms of their semantics, rather than in terms of their implementation. Talk about what setting the context variables *do*.

lib/browser_client.dart
```diff
@@ -23,6 +23,12 @@ import 'src/response.dart';
 /// unable to directly set some headers, such as `content-length`. It is also
 /// unable to stream requests or responses; a request will only be sent and a
 /// response will only be returned once all the data is available.
+///
+/// [BrowserClient] allows setting values directly on the underlying
+/// [HttpRequest] values through [Request.context].
```
nex3: Nit: "." -> ":"

PR #134, opened 2017-10-27
Handle onAbort within BrowserClient

donny-dont:
`BrowserClient.close` calls `abort()` on the `HttpRequest`. Make sure that the client receives an error.

### DIFF
--- a/lib/browser_client.dart
+++ b/lib/browser_client.dart
@@ -73,6 +73,7 @@ class BrowserClient extends BaseClient {
           StackTrace.current);
     });
 
+    // Catch the abort event so futures complete when close is called.
     xhr.onAbort.first.then((_) {
       completer.completeError(
           new ClientException('Request cancelled', request.url),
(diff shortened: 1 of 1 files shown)


lib/browser_client.dart
```diff
@@ -73,6 +73,12 @@ class BrowserClient extends BaseClient {
           StackTrace.current);
     });
 
+    xhr.onAbort.first.then((_) {
```
nex3: I'm guessing we throw this error so that in-flight request futures are guaranteed to complete? If so, add a comment indicating that.

What's the behavior of the IO client if you close it while a request is in-flight?

donny-dont: As an example.

```
import 'dart:async';

main() async {
  var client = new Client();
  var future = client.get('https://httpbin.org/delay/1000');

  print('making request');
  future.then((response) {
    print(response.statusCode);
  }).catchError((error) {
    print(error);
  });

  new Timer(new Duration(seconds: 1), () {
    print('closing');
    client.close();
  });
}
```
Output:
```shell
making request
closing
Connection closed before full header was received
```

PR #135, opened 2017-10-27
Cleanup library code

donny-dont:
General cleanups like

* Use `=>` when possible
* Use `const` constructors when possible
* Remove unused imports
* Use method cascades
* Use `var`

Also makes `_checkResponseSuccess` use `getUrl` and check that the response text is not empty rather than null as `reasonPhrase` is never `null`

### DIFF
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -61,10 +61,10 @@ abstract class BaseClient implements Client {
   /// Throws an error if [response] is not successful.
   static void _checkResponseSuccess(url, Response response) {
     if (response.statusCode < 400) return;
-    var message = 'Request to $url failed with status ${response.statusCode}';
+    var message = "Request to $url failed with status ${response.statusCode}";
     if (response.reasonPhrase.isNotEmpty) {
-      message = '$message: ${response.reasonPhrase}';
+      message = "$message: ${response.reasonPhrase}";
     }
-    throw new ClientException('$message.', getUrl(url));
+    throw new ClientException("$message.", getUrl(url));
   }
 }
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -59,7 +59,7 @@ abstract class BaseClient implements Client {
   void close() {}
 
   /// Throws an error if [response] is not successful.
-  static void _checkResponseSuccess(url, Response response) {
+  void _checkResponseSuccess(url, Response response) {
     if (response.statusCode < 400) return;
     var message = "Request to $url failed with status ${response.statusCode}";
     if (response.reasonPhrase.isNotEmpty) {
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -12,6 +12,7 @@ import 'client.dart';
 import 'exception.dart';
 import 'request.dart';
 import 'response.dart';
+import 'utils.dart';
 
 /// The abstract base class for an HTTP client.
 ///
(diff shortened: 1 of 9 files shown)


lib/src/base_client.dart
```diff
@@ -55,16 +56,15 @@ abstract class BaseClient implements Client {
 
   Future<Response> send(Request request);
 
+  void close() {}
+
   /// Throws an error if [response] is not successful.
-  void _checkResponseSuccess(url, Response response) {
+  static void _checkResponseSuccess(url, Response response) {
```
nex3: I don't generally like making helper methods static unless there's a good reason to do so.

lib/src/base_client.dart
```diff
@@ -55,16 +56,15 @@ abstract class BaseClient implements Client {
...
     if (response.statusCode < 400) return;
-    var message = "Request to $url failed with status ${response.statusCode}";
-    if (response.reasonPhrase != null) {
-      message = "$message: ${response.reasonPhrase}";
+    var message = 'Request to $url failed with status ${response.statusCode}';
```
nex3: Please don't change the quotes. The change isn't worth the churn, and trying to keep quotes consistent isn't worth the mental overhead. Worth noting that it's explicitly *not* part of the style guide.

lib/src/middleware.dart
```diff
@@ -48,17 +48,15 @@ Middleware createMiddleware(
...
+  return (inner) => new HandlerClient(
+        (request) => requestHandler(request)
+            .then((req) => inner.send(req))
+            .then((res) => responseHandler(res), onError: errorHandler),
+        onClose == null
+            ? inner.close
+            : () {
+                onClose();
+                inner.close();
+              },
```
nex3: No trailing commas.

lib/src/middleware.dart
```diff
@@ -48,17 +48,15 @@ Middleware createMiddleware(
...
-          .then((res) => responseHandler(res), onError: errorHandler),
-      onClose == null
-          ? inner.close
-          : () {
-              onClose();
-              inner.close();
-            },
-    );
-  };
+  return (inner) => new HandlerClient(
```
nex3: I'm not convinced this `=>` change actually makes this more readable. There's already a *lot* of nesting and a lot of anonymous functions in the `HandlerClient` initialization, so I think having it clearly be a statement of its own makes this read more clearly.

test/utils.dart
```diff
@@ -107,7 +103,7 @@ class _MultipartBodyMatches extends Matcher {
 ///
 /// [message] can be a String or a [Matcher].
 Matcher isClientException([message]) => predicate((error) {
-      expect(error, new isInstanceOf<http.ClientException>());
+      expect(error, const isInstanceOf<http.ClientException>());
```
nex3: We generally use `new isInstanceOf`, might as well keep this consistent.

donny-dont: Any particular rules for when you want `const` used?

lib/src/base_client.dart
donny-dont: Using `static` is a bit more internally consistent with the rest of the code.

nex3: How so? The only other places I see `static` are where it's necessary to use the methods from constructors.

test/utils.dart
nex3: I usually avoid it in general unless it's in a place where allocations may affect performance.

lib/src/base_client.dart
donny-dont: Sorry you're correct I missed where the other ones were being called.

donny-dont:
@nex3 think this is good to go

PR #136, opened 2017-10-31
Handle timeouts in BrowserClient

### LINKED ISSUE #21: add Timeout functionality for http package
_Originally opened as dart-lang/sdk#19831_

*This issue was originally filed by theburnin...&#064;gmail.com*
_____
**What steps will clearly show the issue / need for enhancement?**
1. when making http requests you usually want to timeout the request after a reasonable amount of time (usually in the region of seconds)
2. it would be great to have away to specify timeout as Duration when making a http request via the http/browser client/http client classes.

**What version of the product are you using? On what operating system?**
Dart 1.5.1, http 0.11.1+1

donny-dont:
Adds support for `HttpRequest.timeout` as well as catching the timeout event.

Closes #21 at least for `BrowserClient`.


nex3:
What's the benefit to exposing this as a client-specific context variable rather than providing client-agnostic middleware?

donny-dont:
The functionality is already built-in with the web platform in this case. A client middleware could use configuration specific imports to choose either the `BrowserClient` path or an implementation for `IOClient`.

nex3:
What's the benefit of using the built-in behavior, though?

> A client middleware could use configuration specific imports to choose either the `BrowserClient` path or an implementation for `IOClient`.

I don't understand. How would it know what the implementation of the root-level client is?

PR #143, opened 2018-01-08
Update MultipartRequest sample for 0.11.x

filiph:
Latest stable (`0.11.3+16`) does not have the `http.MultipartFile.fromFile` constructor and instead has the `http.MultipartRequest.fromPath` static method.


nex3:
Thanks!

PR #150, opened 2018-01-31
Clean up request/response/message tests

nex3:


nex3:
@donny-dont Can you review?

test/message_test.dart
```diff
@@ -255,88 +252,206 @@ void main() {
...
+      group('an unknown content-type header and', () {
+        test('no body', () {
+          var message = _createMessage(headers: {
+            'Content-Type': 'text/plain; charset=not-a-real-charset'
+          });
+          expect(message.encoding, isNull);
+          expect(
+              message.headers,
+              containsPair(
+                  'content-type', 'text/plain; charset=not-a-real-charset'));
```
donny-dont: So no errors here when decoding it will just turn into UTF8 decoding correct?

test/message_test.dart
```diff
@@ -255,88 +252,206 @@ void main() {
...
+        });
+
+        test('a non-ASCII body', () {
+          var message = _createMessage(
+              body: _nonAscii,
+              encoding: LATIN1,
+              headers: {'Content-Type': 'text/plain; charset=utf-8'});
+          expect(message.encoding, equals(LATIN1));
+          expect(message.headers,
+              containsPair('content-type', 'text/plain; charset=iso-8859-1'));
```
donny-dont: Do we care that this is the default charset assumed by the HTTP standard?

test/message_test.dart
nex3: Yes, if `encoding` is `null` and the charset is unknown or not set, `readAsString()` will default to using UTF-8.

test/message_test.dart
nex3: No. It's possible we should have `readAsString()` default to Latin-1 instead of UTF-8, but I think matching `shelf`'s behavior is more important and I'm not eager to make a breaking change there.

PR #151, opened 2018-02-01
Update client_test to match the desired semantics

nex3:


nex3:
@donny-dont Please review.

PR #154, opened 2018-03-07
Refactor adjustHeaders

donny-dont:
This PR refactors _adjustHeaders into two functions, one for `content-length` and one for `content-type`. This hopefully makes the logic a bit clearer than it currently is and sets up for handling `Map` types.

I want to do this PR before finishing off the `Map` handling as this is a total redo of `_adjustHeaders`.

Tests are all passing which is a plus.

### DIFF
--- a/lib/src/message.dart
+++ b/lib/src/message.dart
@@ -14,6 +14,15 @@ import 'content_type.dart';
 import 'http_unmodifiable_map.dart';
 import 'utils.dart';
 
+/// The default set of headers for a message created with no body and no
+/// explicit headers.
+final _defaultHeaders = new HttpUnmodifiableMap<String>({'content-length': '0'},
+    ignoreKeyCase: true);
+
+/// The default media type `application/octet-stream` as defined by HTTP.
+final MediaType _defaultMediaType =
+    new MediaType('application', 'octet-stream');
+
 /// Retrieves the [Body] contained in the [message].
 ///
 /// This is meant for internal use by `http` so the message body is accessible
@@ -150,12 +159,6 @@ abstract class Message {
   Message change(
       {Map<String, String> headers, Map<String, Object> context, body});
 
-  /// The default set of headers for a message created with no body and no
-  /// explicit headers.
-  static final _defaultHeaders = new HttpUnmodifiableMap<String>(
-      {'content-length': '0'},
-      ignoreKeyCase: true);
-
   /// Adds information about encoding and content-type to [headers].
   ///
   /// Returns a new map without modifying [headers].
@@ -172,18 +175,9 @@ abstract class Message {
       }
     }
 
-    var newHeaders = headers == null
-        ? new CaseInsensitiveMap<String>()
-        : new CaseInsensitiveMap<String>.from(headers);
-
-    if (contentType != null) {
-      newHeaders['content-type'] = contentType;
-    }
-
-    if (contentLength != null) {
-      newHeaders['content-length'] = contentLength;
-    }
-
+    var newHeaders = new CaseInsensitiveMap<String>.from(headers ?? const {});
+    if (contentType != null) newHeaders['content-type'] = contentType;
+    if (contentLength != null) newHeaders['content-length'] = contentLength;
     return newHeaders;
   }
 
@@ -196,19 +190,15 @@ abstract class Message {
     if (bodyLength == null) return null;
 
     var contentLengthHeader = bodyLength.toString();
-    if (contentLengthHeader == getHeader(headers, 'content-length'))
+    if (contentLengthHeader == getHeader(headers, 'content-length')) {
       return null;
+    }
 
     var coding = getHeader(headers, 'transfer-encoding');
-
     return coding == null || equalsIgnoreAsciiCase(coding, 'identity')
         ? contentLengthHeader
         : null;
   }
-
-  /// The default media type `application/octet-stream` as defined by HTTP.
-  static final MediaType _defaultMediaType =
-      new MediaType('application', 'octet-stream');
 
   /// Determines the `content-type` from the given [headers] and [body].
   ///
--- a/lib/src/message.dart
+++ b/lib/src/message.dart
@@ -149,55 +149,89 @@ abstract class Message {
   /// changes.
   Message change(
       {Map<String, String> headers, Map<String, Object> context, body});
-}
 
-/// Adds information about encoding to [headers].
-///
-/// Returns a new map without modifying [headers].
-Map<String, String> _adjustHeaders(Map<String, String> headers, Body body) {
-  var sameEncoding = _sameEncoding(headers, body);
-  if (sameEncoding) {
-    if (body.contentLength == null ||
-        getHeader(headers, 'content-length') == body.contentLength.toString()) {
-      return headers ?? const HttpUnmodifiableMap.empty();
-    } else if (body.contentLength == 0 &&
-        (headers == null || headers.isEmpty)) {
-      return const HttpUnmodifiableMap.empty();
+  /// Adds information about encoding and content-type to [headers].
+  ///
+  /// Returns a new map without modifying [headers].
+  static Map<String, String> _adjustHeaders(
+      Map<String, String> headers, Body body) {
+    var contentType = _contentTypeHeader(headers, body);
+    var contentLength = _contentLengthHeader(headers, body);
+
+    if (contentType.isEmpty) {
+      if (contentLength.isEmpty) {
+        return headers ?? const HttpUnmodifiableMap.empty();
+      } else if ((contentLength == '0') &&
+          (headers == null || headers.isEmpty)) {
+        return const HttpUnmodifiableMap.empty();
+      }
     }
+
+    var newHeaders = headers == null
+        ? new CaseInsensitiveMap<String>()
+        : new CaseInsensitiveMap<String>.from(headers);
+
+    if (contentType.isNotEmpty) {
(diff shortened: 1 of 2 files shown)


lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+    var contentTypeHeader = getHeader(headers, 'content-type');
+    var changed = false;
+    MediaType mediaType;
+    Encoding mediaEncoding = null;
+
+    if (contentTypeHeader != null) {
+      mediaType = new MediaType.parse(contentTypeHeader);
+      mediaEncoding = Encoding.getByName(mediaType.parameters['charset']);
+    } else {
+      mediaType = _defaultMediaType;
```
donny-dont: This is where there will be a check for `originalBody is Map` which will then give it the proper MediaType.

donny-dont:
@nex3 @kevmoo could I get a review?

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+  /// Returns a new map without modifying [headers].
+  static Map<String, String> _adjustHeaders(
+      Map<String, String> headers, Body body) {
+    var contentType = _contentTypeHeader(headers, body);
+    var contentLength = _contentLengthHeader(headers, body);
+
+    if (contentType.isEmpty) {
+      if (contentLength.isEmpty) {
+        return headers ?? const HttpUnmodifiableMap.empty();
+      } else if ((contentLength == '0') &&
```
nex3: This check seems wrong. The documentation for `_contentLengthHeader()` says it "Returns the value for the `content-length` header *if it should be modified*", but here it's returning a value without modifying the headers.

It seems like `_contentLengthHeader()` should be in charge of checking for an empty body and header map.

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
-      return const HttpUnmodifiableMap.empty();
+  /// Adds information about encoding and content-type to [headers].
+  ///
+  /// Returns a new map without modifying [headers].
+  static Map<String, String> _adjustHeaders(
+      Map<String, String> headers, Body body) {
+    var contentType = _contentTypeHeader(headers, body);
+    var contentLength = _contentLengthHeader(headers, body);
+
+    if (contentType.isEmpty) {
```
nex3: Use `null` rather than an empty string as a sentinel value.

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+  /// within the `content-type` header. The body's encoding will always
+  /// override the value.
+  ///
+  /// Returns the value for the `content-type` header if it should be
+  /// modified, otherwise it returns the empty string.
+  static String _contentTypeHeader(Map<String, String> headers, Body body) {
+    var contentTypeHeader = getHeader(headers, 'content-type');
+    var changed = false;
+    MediaType mediaType;
+    Encoding mediaEncoding = null;
```
nex3: Nit: `= null` is unnecessary here.

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+    if (contentTypeHeader != null) {
+      mediaType = new MediaType.parse(contentTypeHeader);
+      mediaEncoding = Encoding.getByName(mediaType.parameters['charset']);
+    } else {
+      mediaType = _defaultMediaType;
+    }
+
+    var bodyEncoding = body.encoding;
+
+    if ((bodyEncoding != null) && (bodyEncoding != mediaEncoding)) {
```
nex3: Nit: unnecessary parens

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+      changed = true;
```
nex3: Rather than tracking `changed`, I'd just write:

```dart
if (body.encoding == null) return null;
if (body.encoding == mediaEncoding) return null;

return mediaType.change(...);
```

lib/src/message.dart
```diff
@@ -149,55 +149,89 @@ abstract class Message {
...
+    Encoding mediaEncoding = null;
+
+    if (contentTypeHeader != null) {
+      mediaType = new MediaType.parse(contentTypeHeader);
+      mediaEncoding = Encoding.getByName(mediaType.parameters['charset']);
+    } else {
+      mediaType = _defaultMediaType;
+    }
+
+    var bodyEncoding = body.encoding;
```
nex3: I don't think this variable makes anything clearer.

lib/src/message.dart
```diff
@@ -149,55 +149,94 @@ abstract class Message {
...
-        getHeader(headers, 'content-length') == body.contentLength.toString()) {
-      return headers ?? const HttpUnmodifiableMap.empty();
-    } else if (body.contentLength == 0 &&
-        (headers == null || headers.isEmpty)) {
-      return const HttpUnmodifiableMap.empty();
+  /// The default set of headers for a message created with no body and no
+  /// explicit headers.
+  static final _defaultHeaders = new HttpUnmodifiableMap<String>(
+      {'content-length': '0'},
+      ignoreKeyCase: true);
```
donny-dont: This mirrors what is being done in `shelf`.

https://github.com/dart-lang/shelf/blob/master/lib/src/message.dart#L17-L21
```
final _defaultHeaders = new ShelfUnmodifiableMap<String>(
    {"content-length": "0"},
    ignoreKeyCase: true);
```

lib/src/message.dart
```diff
@@ -149,55 +149,94 @@ abstract class Message {
...
+      Map<String, String> headers, Body body) {
+    var contentType = _contentTypeHeader(headers, body);
+    var contentLength = _contentLengthHeader(headers, body);
+
+    if (contentType == null) {
+      if (contentLength == null) {
+        return headers ?? const HttpUnmodifiableMap.empty();
+      } else if ((contentLength == '0') &&
+          (headers == null || headers.isEmpty)) {
+        return _defaultHeaders;
```
donny-dont: This is now fixed.

lib/src/message.dart
```diff
@@ -149,55 +149,94 @@ abstract class Message {
...
+    if (contentTypeHeader != null) {
+      mediaType = new MediaType.parse(contentTypeHeader);
+      mediaEncoding = Encoding.getByName(mediaType.parameters['charset']);
+    } else {
+      mediaType = _defaultMediaType;
+    }
 
-  var charset = new MediaType.parse(contentType).parameters['charset'];
-  return Encoding.getByName(charset) == body.encoding;
+    if ((body.encoding != null) && (body.encoding != mediaEncoding)) {
```
donny-dont: We can't split this into

```
if (body.encoding == null) return null;
if (body.encoding != mediaEncoding) return null;
```

The point of the `changed` value is to allow us to insert a `content-type` for `Map` values. This will be in the next PR.

Since `shelf` has no special handling of `Map` types then there would be no `changed` value and the logic would flow like you suggested.

donny-dont:
@nex3 https://github.com/donny-dont/http/commit/f8ccc46bcdb732edf2802cc4425d5d208fcb67c5 is the next commit assuming this lands as it currently is.

lib/src/message.dart
donny-dont: Just one thought here. https://github.com/dart-lang/shelf/issues/58 if that is still something you want to have happen then perhaps we should not make map a special case.

donny-dont:
@nex3 can we make a decision on this? I have the next patch ready after this lands to add urlencoded form support.

I responded to your comment about the behavior of `content-length`. I'm really not sure why the number of headers present should influence whether `content-length` is populated. Is there something in the HTTP spec that says that should happen?

test/hybrid/server.dart
```diff
@@ -84,6 +84,8 @@ hybridMain(StreamChannel channel) async {
     request.headers.forEach((name, value) {
       // Ignore headers that are generated by the client
       if (_ignoreHeaders.contains(name)) return;
+      // Ignore a zero content-length header generated by the client
+      if (name == 'content-length' && value == '0') return;
```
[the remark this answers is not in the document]
nex3: We definitely shouldn't be re-writing the headers on the server-side. That doesn't make any sense.

If underlying implementations consistently fill in a content length, though, I suppose it's fine for us to default to `0` rather than to not setting it.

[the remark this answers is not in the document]
donny-dont: They don't consistently fill in a content length that's what makes this messy. You can see the test logs on this PR https://travis-ci.org/dart-lang/http/jobs/350628472 or run this branch locally yourself to see.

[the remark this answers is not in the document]
nex3: I'm fine just always explicitly saying `content-length: 0` to provide a strong guarantee about when that will be set.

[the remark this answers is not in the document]
donny-dont: Ok so roll back to https://github.com/dart-lang/http/pull/154/commits/036e4fd6fed9e6e9e222fa22541ed6df677e9146 then we're good to go?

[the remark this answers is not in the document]
nex3: I think you mean 3269f49e94c5742d6e0a1ded3f7699c7b8193e85? If so, yeah.

donny-dont:
@nex3 its rolled back to the state before the content-length null was attempted.

nex3:
I don't think I'll have time to land this today, but I'll prioritize it on Monday.

donny-dont:
🎊

PR #155, opened 2018-03-13
Add urlencoded form support

donny-dont:
This adds urlencoded form support.

All tests are now passing.


donny-dont:
@nex3 please review 😄

PR #187, opened 2018-08-07
Add examples

donny-dont:
This adds a few simple examples of using the `http` library.


donny-dont:
@nex3 still could probably use a middleware example but other than that it should be good to go.

PR #189, opened 2018-08-08
Fix travis

kevmoo:


.test_config
```diff
@@ -0,0 +1,3 @@
+{
+  "test_package": true
```
natebosch: what is this file?

kevmoo: picked up by the package bots

kevmoo:
> Please add more detail to the commit message

See description in the two commits

natebosch:
> See description in the two commits

I see "misc: fix tests" which doesn't tell me much 😛

kevmoo:
@natebosch – PTAL

Issue #22, opened 2015-06-05
Support both the console and browser equally


DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [nex3](https://github.com/nex3)**
_Originally opened as dart-lang/sdk#20068_

---

The http package should support IOClient and BrowserClient on even footing. &quot;new http.Client()&quot; should dynamically choose between them based on which is supported.

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

*Marked this as being blocked by dart-lang/sdk#6943.*

DartBot:
*This comment was originally written by mi...&#064;apaq.dk*
_____
This would be awesome. 

In my particular case I'm building a client library for a Rest service. My first priority is to build a CLI application using the Client and for that purpose I will need to use the dart:io's HttpRequest. But I want to reuse the Client lib for a Web Interface later on and will then need to use dart:html's HttpRequest.

In order to make that work I have to build wrappers for each HttpRequests.

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

It's worth noting that you can use the http package on both the browser and the console today; you just need to explicitly choose to instantiate either an IOClient or a BrowserClient and pass it in to the code in question.

DartBot:
*This comment was originally written by mi...&#064;apaq.dk*
_____
Huh. Why didn't I notice that until now? Please excuse a newbie. :)

Reading "http package" I was just thinking of some future http functionality - I didn't know it was a package that already existed. Thank you for guiding me.

hermanbergwerf:
Is anything planned to resolve this? This blocks me from using `http.get` etc. functions. I would actually be ok with manually setting what Client is used by `_withClient` once in the top-level package (the application). Currently the only 'clean' alternative is passing clients around through all of my code so the top-level package can specify the right client for the environment.

nex3:
This is blocked on dart-lang/sdk#24581. Once full cross-platform imports are fully supported and stable, I'll fix this ASAP.

[...]

PR #198, opened 2018-09-12
Make `package:http/http.dart` support all platforms

jakemac53:
Fixes https://github.com/dart-lang/http/issues/22

Adds config specific imports to make `package:http/http.dart` support all platforms, and allow the `Client` factory constructor to return a valid `Client` for the web platform.

This should eliminate almost all need for the platform specific imports for consumers, although it does also add the `io_client.dart` public import.

Passes presubmit internally, with edits in only 3 files (they use the IoClient constructor directly, and pass in an HttpClient). Externally build_runner now supports config specific imports as well, I am currently working on validating everything works as expected there with this change.

### New Features

* The regular `Client` factory constructor is now usable anywhere that `dart:io`
  or `dart:html` are available, and will give you an `IoClient` or
  `BrowserClient` respectively.
* The `package:http/http.dart` import is now safe to use on the web (or
  anywhere that either `dart:io` or `dart:html` are available).

### Breaking Changes

* In order to use or reference the `IoClient` directly, you will need to import
  the new `package:http/io_client.dart` import. This is typically only necessary
  if you are passing a custom `HttpClient` instance to the constructor, in which
  case you are already giving up support for web.

### DIFF
--- a/lib/src/browser_client.dart
+++ b/lib/src/browser_client.dart
@@ -1,0 +1,110 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:html';
+import 'dart:typed_data';
+
+import 'base_client.dart';
+import 'base_request.dart';
+import 'byte_stream.dart';
+import 'exception.dart';
+import 'streamed_response.dart';
+
+BaseClient createClient() => BrowserClient();
+
+// TODO(nweiz): Move this under src/, re-export from lib/http.dart, and use this
+// automatically from [new Client] once sdk#24581 is fixed.
+
+/// A `dart:html`-based HTTP client that runs in the browser and is backed by
+/// XMLHttpRequests.
+///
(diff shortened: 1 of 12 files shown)


zoechi:
The `environment` constraint lower limit is correct to support conditional imports?

lib/src/browser_client.dart
```diff
@@ -0,0 +1,110 @@
...
+
+import 'base_client.dart';
+import 'base_request.dart';
+import 'byte_stream.dart';
+import 'exception.dart';
+import 'streamed_response.dart';
+
+BaseClient createClient() => BrowserClient();
+
+// TODO(nweiz): Move this under src/, re-export from lib/http.dart, and use this
```
natebosch: can delete this comment now?

jakemac53: done

jakemac53:
> The `environment` constraint lower limit is correct to support conditional imports?

Technically, yes. We only recently actually had support in build systems and things, but we can't express that dependency unfortunately (ideally, we would be able to say _if_ you have a dependency on build_runner, it must be >x.x.x (which supports these imports), but we can't express that).

nex3:
I'm swamped with Sass work and it looks like Nate has this review in hand, so I'm going to remove myself.

PR #219, opened 2018-12-14
Add a small example

mit-mit:


example/main.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:http/http.dart' as http;
+import 'dart:convert' as convert;
```
natebosch: [nit] put this import first with a blank line in between

https://www.dartlang.org/guides/language/effective-dart/style#do-place-dart-imports-before-other-imports

mit-mit: Done

example/main.dart
```diff
@@ -0,0 +1,19 @@
...
+
+main(List<String> arguments) {
+  // This example uses the Google Books API to search for books about http.
+  // https://developers.google.com/books/docs/overview
+  var url = "https://www.googleapis.com/books/v1/volumes?q={http}";
+
+  // Await the http get response, then decode the json-formatted responce.
```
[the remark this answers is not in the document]
mit-mit: Done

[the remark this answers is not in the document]
mit-mit: I tried to update `readme.md` also, but doesn't look like the code `client.get(response.bodyFields['uri'])` works anymore -- I don't see a `bodyFields` on `response`?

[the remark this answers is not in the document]
kevmoo: 🤷‍♂️ – maybe we should sync the example in the readme w/ the one you're adding?

example/pubspec.yaml
```diff
@@ -0,0 +1,8 @@
+name: httpexample
```
kevmoo: This isn't necessary, right? Just drop the pubspec. The package site picks this up fine.

mit-mit: How would you run the example then?

kevmoo: `dart example/example.dart` 🤷‍♂️

mit-mit:
```
$ ls -a
.		..		main.dart
$
```

mit-mit: Never mind, my mistake, I had moved this out into another directory structure. Working now...

PR #249, opened 2019-03-06
Update README to match current behavior

### LINKED ISSUE #240: README out of date?
From the README:

> The easiest way to use this library is via the top-level functions, although they currently only work on platforms where dart:io is available

Following works just fine on the web now:
```dart
import 'package:http/http.dart' show get;

Future<void> main() async
{
  final response = await get('https://google.com');
  print(response.body);
}

```

natebosch:
Closes #240

- Remove the browser specific sections and caveats - these are no longer
  necessary to call out since this package uses config specific imports.
- Use single quoted strings consistently for style.
- Update samples to use async/await.
- Update doc links to point to the pub site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,26 +5,22 @@ A composable, Future-based library for making HTTP requests.
 
 This package contains a set of high-level functions and classes that make it
 easy to consume HTTP resources. It's platform-independent, and can be used on
-both the command-line and the browser. Currently the global utility functions
-are unsupported on the browser; see "Using on the Browser" below.
+both the command-line and the browser.
 
 ## Using
 
-The easiest way to use this library is via the top-level functions, although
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,26 +5,22 @@ A composable, Future-based library for making HTTP requests.
...
-  print("Response status: ${response.statusCode}");
-  print("Response body: ${response.body}");
-});
+var url = 'http://example.com/whatsit/create';
+var response = await http.post(url, body: {'name': 'doodle', 'color': 'blue'});
+print('Response status: ${response.statusCode}');
+print('Response body: ${response.body}');
 
-http.read("http://example.com/foobar.txt").then(print);
+http.read('http://example.com/foobar.txt').then(print);
```
jakemac53: should we use await here too?

PR #261, opened 2019-03-19
added a note on using this library in Flutter

jeffmikels:
Documentation should help the end user use the library. This library will not work in Flutter unless the app permissions are set properly. Adding this note helps new users make the proper changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,29 @@ class UserAgentClient extends http.BaseClient {
   }
 }
 ```
+## Using on the Browser
+The HTTP library can be used on the browser via the BrowserClient class in package:http/browser_client.dart. This client translates requests into XMLHttpRequests. For example:
+
+```dart
+import 'dart:async';
+import 'package:http/browser_client.dart';
+
+main() async {
+  var client = new BrowserClient();
+  var url = '/whatsit/create';
+  var response =
+      await client.post(url, body: {'name': 'doodle', 'color': 'blue'});
+  print('Response status: ${response.statusCode}');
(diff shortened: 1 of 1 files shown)


jeffmikels:
I signed it.

README.md
```diff
@@ -67,3 +67,29 @@ class UserAgentClient extends http.BaseClient {
...
 ```
+## Using on the Browser
+The HTTP library can be used on the browser via the BrowserClient class in package:http/browser_client.dart. This client translates requests into XMLHttpRequests. For example:
+
+```dart
+import 'dart:async';
+import 'package:http/browser_client.dart';
+
+main() async {
+  var client = new BrowserClient();
```
jakemac53: Note that the regular `Client` class from the regular `package:http/http.dart` import should also give you a `BrowserClient`. This is a more portable mechanism so it is probably what we should advertise.

I believe that the `browser_client.dart` is pretty much only exists for historical reasons at this point, although its possible that it has some additional apis on top of the regular Client interface (not sure though).

README.md
```diff
@@ -67,3 +67,29 @@ class UserAgentClient extends http.BaseClient {
...
+  var url = '/whatsit/create';
+  var response =
+      await client.post(url, body: {'name': 'doodle', 'color': 'blue'});
+  print('Response status: ${response.statusCode}');
+  print('Response body: ${response.body}');
+}
+```
+
+## Using in Flutter
```
mit-mit: This is now covered in https://github.com/flutter/website/pull/3548

Issue #225, opened 2019-01-11
Throws exception 'Failed to parse header value'


sleith:
Hi,

> Http lib version 0.12.0+1
> Flutter 1.0.0 • channel beta • https://github.com/flutter/flutter.git
> Framework • revision 5391447fae (6 weeks ago) • 2018-11-29 19:41:26 -0800
> Engine • revision 7375a0f414
> Tools • Dart 2.1.0 (build 2.1.0-dev.9.4 f9ebf21297)

I tried http.get(). It works fine on response code 200.
But when my server returns 401 Unauthorized, it throws exception "Failed to parse header value".
Please find the header below i got from Postman:

> Date →Fri, 11 Jan 2019 10:04:09 GMT
> Server →kong/0.11.2
> Content-Type →application/json; charset=utf-8
> WWW-Authenticate →Bearer realm="service" error="invalid_token" error_description="The access token is invalid or has expired"
> Vary →Accept-Encoding
> Content-Encoding →gzip
> Content-Length →97
> Keep-Alive →timeout=15, max=100
> Connection →Keep-Alive

Any idea why it failed parse header value?
Thank you

zoechi:
Can you please post the full exception output?

sleith:
Hi,
i print(exception) only show that text "Failed to parse header value".
Here's the code and url you can try to see this issue:

```
Future<http.Response> testHit() async {
    Map<String, String> headers = new Map<String, String>();
    headers['Content-Type'] = "application/json";
    headers['Accept'] = "application/json";
    headers['nik'] = "1805304321";
    headers['deviceNumber'] = "869058030276537";
    headers['role'] = "SALES_AGENT";
    headers['Authorization'] = "Bearer dXKCloWJ0KQp05nQSbfmj6EvO6sILGB5}";

    http.Response response;
    try {
      response = await http.get(
          "https://api-uat.homecredit.co.id/smart-headline/v1/headline/",
          headers: headers);
      longPrint("--> Requeststatus: " + response.statusCode.toString());

      if (response.statusCode == 200) {
        longPrint("->" + response.body);
        return response;
      } else {
        print("->Request failed with status:" + response.statusCode.toString());
      }
    } catch (exception) {
      print(exception);
      if (response != null) {
        print("Request failed with status: " + response.statusCode.toString());
      } else {
        print("response null");
      }
    }
    return null;
  }
```

If you run on postman it will give output.
Thank you :)

sleith:
Hi,
[...]

Issue #15, opened 2015-06-05
http headers Content-Type problems


DartBot:
_Originally opened as dart-lang/sdk#17405_

*This issue was originally filed by r.spil...&#064;gmail.com*
_____
A Content-Type header is ignored when sending http.post.

If I call http.post with a header &quot;Content-Type&quot;, the server doesn't get the value specified if there is also a body. Instead, it always gets &quot;text/plain; charset=utf-8&quot;.

If I change the header key to &quot;content-type&quot;, everything works as expected.

There are several issues here.

1) I would expect the header key to be used case-insensitive.
2) Requests set body has as side-effect that the \_contentType is set.

This side-effect is documented, however it explicitly mentions &quot;Content-Type&quot;. And so does Requests `ContentType get \_contentType`.

I've attached two source files to demonstrate the problem.

______
**Attachments:**
[httpheadersserver.dart](https://storage.googleapis.com/google-code-attachments/dart/issue-17405/comment-0/httpheadersserver.dart) (288 Bytes)
[httpheadersclient.dart](https://storage.googleapis.com/google-code-attachments/dart/issue-17405/comment-0/httpheadersclient.dart) (243 Bytes)
[pubspec.yaml](https://storage.googleapis.com/google-code-attachments/dart/issue-17405/comment-0/pubspec.yaml) (91 Bytes)

DartBot:
<a href="https://github.com/kasperl"><img src="https://avatars.githubusercontent.com/u/2156198?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kasperl](https://github.com/kasperl)**

----

*Added [Area-IO](../labels/Area-IO), [Triaged](../labels/Triaged) labels.*

DartBot:
*This comment was originally written by r.sp...&#064;gmail.com*
_____
I think the problem is in base_request.dart:

&nbsp;BaseRequest(this.method, this.url)
&nbsp;&nbsp;&nbsp;&nbsp;: headers = &lt;String, String>{};

This map is not case-insensitive, it should be. A few lines higher, I've found the following comment:

&nbsp;&nbsp;// TODO(nweiz): make this a HttpHeaders object
&nbsp;&nbsp;/// The headers for this request.
&nbsp;&nbsp;final Map&lt;String, String> headers;

I tried to figure out if I could make a patch, but I encountered two problems:
- I don't understand how I would be able to use the LinkedHashMap external factory that accepts alternative equality and hash functions for keys
[...]

Issue #184, opened 2018-07-19
Remove charset from "Content-Type" header


Pandazaur:
Hello,

I need to post some data to a distant API. But it looks like the API wants `Content-Type: application/json` but I'm sending a `Content-Type: application/json; charset=utf-8`.

I forced the headers:
```dart
Map<String, String> headers = {
            HttpHeaders.AUTHORIZATION: 'Bearer ${this._token}',
            HttpHeaders.CONTENT_TYPE: 'application/json'
};
```

but keep sending with the charset. Is there a possibility to remove it from the `Content Type`header ?

Thanks

cskau-g:
Looks like charset ("encoding") is always added on Content-Type in case it's not set:
https://github.com/dart-lang/http/blob/master/lib/src/request.dart#L84-L92

For good measure I checked with the RFCs and it seems the "parameter" part of the Content-Type value is either mandatory or optional depending on the particular MIME type/subtype:
https://tools.ietf.org/html/rfc2045#section-5

As far as I can tell the 'application/json' MIME type does explicitly _not_ define a "charset" parameter: 
https://www.iana.org/assignments/media-types/application/json

Ultimately this points to the http package being buggy. I don't think it's valid to blindly add parameters to just any MIME type. In fact, while probably done with good intentions, silently adding anything should probably be avoided.

jheyne:
**Work-around:** Set the body first _(this sets default headers)_, then set the headers.

jefioliveira:
I have the same problem, however the above solution did not work in my case.

jheyne:
For clarity, this is what worked for me:
    Request request = new Request('POST', Uri.parse(url));
    request.body = body;
    request.headers['content-type'] = 'text/xml';
    request.headers['accept'] = 'text/xml';

aloisdeniel:
I encountered the same issue, but @jheyne workaround didn't work for me.

We should be able to write `request.encoding = null`.

cmregazzo:
Good afternoon,

I have the same problem.

The lib is automatically placing charset = utf-8 in the request header, so I'm having trouble consuming some REST WebServices, making it impossible to use it.

The issue is already open for one year.

Is there no solution to such a problem?

Thank you.

Cristian Regazzo

Dudecor3:
Has anyone solved this yet? 

It seems to be more of a problem when there encoding more complex types such as 

`class MyClass {
    ClassOne class1;
    ClassTwo class2;
}`

[...]

PR #269, opened 2019-04-21
Ignore charset encoding in header

jefioliveira:
I need to ignore the charset in the header.

Issue #15
issue #225

### DIFF
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -152,24 +152,27 @@ abstract class BaseClient implements Client {
     if (url is String) url = Uri.parse(url);
     var request = new Request(method, url);
 
-    if (headers != null) request.headers.addAll(headers);
-   if(encoding != null) request.encoding = encoding; else {
-     request.isIgnoreEncoding = true;
-   }
-      if (body != null) {
-        if (body is String) {
-          request.body = body;
-        } else if (body is List) {
-          request.bodyBytes = body.cast<int>();
-        } else if (body is Map) {
-          request.bodyFields = body.cast<String, String>();
-        } else {
-          throw new ArgumentError('Invalid request body "$body".');
-        }
+    if (headers != null) {
+      request.headers.addAll(headers);
+    }
+    if (encoding != null) {
+      request.encoding = encoding;
+    } else {
+      request.isIgnoreEncoding = true;
+    }
+    if (body != null) {
+      if (body is String) {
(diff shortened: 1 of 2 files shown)


jefioliveira:
I signed it!

lib/src/base_client.dart
```diff
@@ -153,20 +153,23 @@ abstract class BaseClient implements Client {
...
-    if (body != null) {
-      if (body is String) {
-        request.body = body;
-      } else if (body is List) {
-        request.bodyBytes = body.cast<int>();
-      } else if (body is Map) {
-        request.bodyFields = body.cast<String, String>();
-      } else {
-        throw new ArgumentError('Invalid request body "$body".');
+   if(encoding != null) request.encoding = encoding; else {
```
natebosch: We will need to run `dartfmt` and the body of the `if` needs to be wrapped in `{}`.

lib/src/request.dart
```diff
@@ -56,7 +58,10 @@ class Request extends BaseRequest {
     _defaultEncoding = value;
     var contentType = _contentType;
     if (contentType == null) return;
-    _contentType = contentType.change(parameters: {'charset': value.name});
+    if (isIgnoreEncoding)
+      _contentType = contentType.change(parameters: {});
```
natebosch: Any `if` with an `else` needs both blocks in braces.

https://www.dartlang.org/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures

cskau-g:
The PR description isn't very descriptive. Can you please edit it to say what this is and why we (all Dart HTTP users) need it?
The two referenced issues seem unrelated?

Have you considered how this will work with Request.encoding?

cskau-g:
Ah, reviewing #184 I think I see that this PR is trying to address that issue specifically?

Parts of this change is a bit misleading since this doesn't really have anything to do with encoding, but is about whether we latch the charset on in Content-Type.
The way I see the problem is that there isn't really a mechanism for controlling the charset directly through the API, and it's automatically latch on where it really should be fully user controlled.
From skimming this, it doesn't really seem like an elegant solution to do that, so I think we might need to put some more thought into how we go about this.

Another aside is that something like this definitely needs tests.
We want to make sure we don't silently break clients who rely on the quirk of charset being set.

mit-mit:
Closing as CLA wasn't signed

PR #275, opened 2019-05-13
remove ignores

devoncarew:
- remove no longer necessary `ignore: uri_does_not_exist`

I don't think these are necessary anymore - I don't see analysis issues when I remove them.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -8,11 +8,8 @@ import 'dart:typed_data';
 
 import 'base_client.dart';
 import 'base_request.dart';
-// ignore: uri_does_not_exist
 import 'client_stub.dart'
-    // ignore: uri_does_not_exist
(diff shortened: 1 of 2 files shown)


devoncarew:
@jakemac53 @kevmoo

kevmoo:
My understanding: this breaks DDC if these are removed!

jakemac53:
These are necessary for build_web_compilers <2.0.0, the analyzer never properly implemented these imports so they give warnings when they shouldn't. This only happens at compile time when only one of those sources will be available to the analyzer (the one it should select).

devoncarew:
@jakemac53, is https://github.com/dart-lang/sdk/issues/34179 the same issue?

jakemac53:
https://github.com/dart-lang/sdk/issues/34177 is the one for imports

kevmoo:
We can close this out, right?

PR #278, opened 2019-05-15
Fixed #277, allow multiple fields of the same name in multipart request

### LINKED ISSUE #277: Multipart form request with fields of the same name only retains one field value
With Multipart form data, it is allowed to have multiple fields of the same name with different values and they all are submitted to the server.

A curl example of that is below (from Mailgun API documentation):

```bash
curl -s --user 'api:YOUR_API_KEY' \
   https://api.mailgun.net/v3/routes \
   -F priority=0 \
   -F description='Sample route' \
   -F expression='match_recipient(".*@YOUR_DOMAIN_NAME")' \
   -F action='forward("http://myhost.com/messages/")' \
   -F action='stop()'
```

Here is the test case:
```dart
  test('with multiple fields of the same name', () {
    var request = http.MultipartRequest('POST', dummyUrl);
    request.fields['field1'] = 'value1';
    request.fields['field1'] = 'value2';

    expect(request, bodyMatches('''
        --{{boundary}}
        content-disposition: form-data; name="field1"

        value1
        --{{boundary}}
        content-disposition: form-data; name="field1"

        value2
        --{{boundary}}--
        '''));
  });
```

The above test case fails because we store fields in a Map and so `value2` replaces the previously set `value1` for `field1`.

AfzalivE:
Converted the fields Map to List of MapEntry objects

### DIFF
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -39,14 +39,16 @@ class MultipartRequest extends BaseRequest {
   static final Random _random = Random();
 
   /// The form fields to send for this request.
-  final List<MapEntry<String, String>> fields;
+  // final List<MapEntry<String, String>> fields;
+
+  final Map<String, List<String>> fields;
 
   /// The private version of [files].
   final List<MultipartFile> _files;
 
   /// Creates a new [MultipartRequest].
   MultipartRequest(String method, Uri url)
-      : fields = [],
+      : fields = {},
         _files = <MultipartFile>[],
         super(method, url);
 
@@ -58,13 +60,15 @@ class MultipartRequest extends BaseRequest {
   int get contentLength {
     var length = 0;
 
-    fields.forEach((field) {
-      length += "--".length +
-          _BOUNDARY_LENGTH +
-          "\r\n".length +
-          utf8.encode(_headerForField(field.key, field.value)).length +
-          utf8.encode(field.value).length +
-          "\r\n".length;
+    fields.forEach((name, values) {
+      values.forEach((value) {
+        length += "--".length +
+            _BOUNDARY_LENGTH +
+            "\r\n".length +
+            utf8.encode(_headerForField(name, value)).length +
+            utf8.encode(value).length +
+            "\r\n".length;
+      });
     });
 
     for (var file in _files) {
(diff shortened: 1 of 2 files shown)


AfzalivE:
@kevmoo @jakemac53

Sadly this looks like it will break existing API. Is there a better way of approaching this that won't?

Thanks

natebosch:
If we do anything here a `List<MapEntry<String, String>>` would not be my choice for API. Either `Map<String, List<String>>` or some kind of multimap thing would be better.

I had a hard time figuring out if the spec has anything to say about duplicated fields - from what I saw I think the Mailgun API in question should work if you use `fields['action[0]']=...`. Does that work for you?

AfzalivE:
> If we do anything here a `List<MapEntry<String, String>>` would not be my choice for API. Either `Map<String, List<String>>` or some kind of multimap thing would be better.
> 
> I had a hard time figuring out if the spec has anything to say about duplicated fields - from what I saw I think the Mailgun API in question should work if you use `fields['action[0]']=...`. Does that work for you?

It doesn't say anything about duplicated fields (presumably because that's supposed to work everywhere).

`fields['action[0]']` just adds a field called `action[0]` in the payload. Not sure if that is what you're suggesting. This is what I tried:

```
request.fields['action[0]'] = routeData.action[0];
request.fields['action[1]'] = routeData.action[1];
```

natebosch:
In the issue you mentioned mailgun. I found a thread that seemed to indicate they should accept requests in the `action[n]` format:

https://groups.google.com/d/msg/spray-user/5kSZ87OnfkE/rOzoNhzPRD0J

AfzalivE:
> In the issue you mentioned mailgun. I found a thread that seemed to indicate they should accept requests in the `action[n]` format:
> 
> https://groups.google.com/d/msg/spray-user/5kSZ87OnfkE/rOzoNhzPRD0J

I just tried postman with the action[n] format and that didn't work at all. I wouldn't be surprised if they changed things since 2013 though.

AfzalivE:
Okay changed it to a list so the interface is now:

```
    request.fields['field1'] = ['value1'];
    request.fields['field2'] = ['value2'];
```

natebosch:
@mit-mit @jonasfj - can you triage and make a decision on whether we want a breaking change for this use case?

cskau-g:
RE: @natebosch https://github.com/dart-lang/http/pull/278#issuecomment-496700971
> I had a hard time figuring out if the spec has anything to say about duplicated fields

[RFC2388](https://tools.ietf.org/html/rfc2388#section-5.5) says:
> The relationship of the ordering of fields within a form and the ordering of returned values within "multipart/form-data" is **not defined by this specification, nor is the handling of the case where a form has multiple fields with the same name**.

[RFC7578](https://tools.ietf.org/html/rfc7578#section-5.2) follows up by saying:
> Form parts with identical field names MUST NOT be coalesced.

In general form and query data encoding is more de-facto standards than hard specs.

FWIW, Chromium encodes duplicate fields in `multipart/form-data`:
```
Content-Disposition: form-data; name="myfield"
..
Content-Disposition: form-data; name="myfield"
```
the same way it does for `application/x-www-form-urlencoded`:
```
?myfield=val1&myfield=val2
```
by writing it multiple times.

AfzalivE:
Since this hasn't been touched in 4 years, I'm going to close this.

Issue #295, opened 2019-09-03
Consider making multipart_request _boundaryString prefix configurable


DartMen:
One is unable to set a custom boundary value as it is hardcoded with "dart-http-boundary-".

PR #300, opened 2019-09-07
Adjustable boundary (#295)

DartMen:
One would be able to adjust the boundary prefix. When not set, defaults to "dart-http-boundary-".

### DIFF
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -44,10 +44,14 @@ class MultipartRequest extends BaseRequest {
   /// The private version of [files].
   final List<MultipartFile> _files;
 
+  /// The boundary prefix. When not set, defaults to "dart-http-boundary-".
+  String boundaryPrefix;
+
   /// Creates a new [MultipartRequest].
   MultipartRequest(String method, Uri url)
       : fields = {},
         _files = <MultipartFile>[],
+        boundaryPrefix = "dart-http-boundary-",
         super(method, url);
 
   /// The list of files to upload for this request.
@@ -160,7 +164,7 @@ class MultipartRequest extends BaseRequest {
 
   /// Returns a randomly-generated multipart boundary string
   String _boundaryString() {
-    var prefix = "dart-http-boundary-";
+    var prefix = boundaryPrefix;
     var list = List<int>.generate(
         _BOUNDARY_LENGTH - prefix.length,
         (index) =>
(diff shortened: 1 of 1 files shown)


lib/src/multipart_request.dart
```diff
@@ -44,10 +44,14 @@ class MultipartRequest extends BaseRequest {
   /// The private version of [files].
   final List<MultipartFile> _files;
 
+  /// The boundary prefix. When not set, defaults to "dart-http-boundary-".
+  String boundaryPrefix;
```
kevmoo: Make this final

DartMen: Consider the following:
~~~dart
      var request = MultipartRequest(method, Uri.parse(url));
      request.boundaryPrefix = "custom-boundary-";
~~~
When making it final, the compiler says: 'boundaryPrefix' can't be used as a setter because it is final.


How about a setter and private backing field?
~~~dart
  /// The boundary prefix. When not set, defaults to "dart-http-boundary-".
  set boundaryPrefix(String value) => _boundaryPrefix = value;
  String _boundaryPrefix = "dart-http-boundary-";

  MultipartRequest(String method, Uri url)
      : fields = {},
        _files = <MultipartFile>[],
        super(method, url);
~~~

natebosch: How about aiming for a usage like:

```dart
var request = MultipartRequest(method, Uri.parse(url), boundaryPrefix: "custom-boundary-");
```

```dart
final String boundaryPrefix;

MultipartRequest(String method, Uri url, {String boundaryPrefix})
    : fields = {},
      _files = <MultipartFile>[],
      boundaryPrefix = boundaryPrefix ?? "dart-http-boundary-",
      super(method, url);
```

DartMen: > How about aiming for a usage like:
> 
> ```dart
> var request = MultipartRequest(method, Uri.parse(url), boundaryPrefix: "custom-boundary-");
> ```
> 
> ```dart
> final String boundaryPrefix;
> 
> MultipartRequest(String method, Uri url, {String boundaryPrefix})
>     : fields = {},
>       _files = <MultipartFile>[],
>       boundaryPrefix = boundaryPrefix ?? "dart-http-boundary-",
>       super(method, url);
> ```

I can imagine not many users would use this adjustable property, that's why I was avoiding the population of the ctor.

Let me implement your example.

natebosch:
Can you elaborate on why this is useful?

DartMen:
> Can you elaborate on why this is useful?

To be able to exactly mimic web browsers. But I realize more work has to be done to accomplish this, as the "randomly-generated multipart boundary string" must also become adjustable.

PR #317, opened 2019-09-13
Fix a broken test for redirects

natebosch:
This test added a `.catchError` callback with some expects, but the
`Future` was completing normally so the callback and expects never ran.
The reason the response came back successfully was that the IO client
does not consider a `302` to be a redirect when the method is `'POST'`.

https://github.com/dart-lang/sdk/blob/ed9e89ea388e4bc0142bf6e967d4ca11999bdfdc/sdk/lib/_http/http_impl.dart#L355-L365

- Refactor to async/await to clean up the noise of `.then`,
  `.whenComplete`, and `.catchError` and make the logic easier to
  follow.
- Switch the request type to `'GET'` so that the exception occurs.
- Move server starting and stopping to a `setUp` and `tearDown`.
- Fix the expectation to look for the wrapped `ClientException` instead
  of the exception thrown by the IO client. We also lose the underlying
  exception and have only the message so we can't check the length of
  the `redirects` field.
- Remove the now unused test utility to get a matcher on the old
  exception type.

### DIFF
--- a/test/io/request_test.dart
+++ b/test/io/request_test.dart
@@ -10,61 +10,48 @@ import 'package:test/test.dart';
 import 'utils.dart';
 
 void main() {
-  test('.send', () {
-    expect(
-        startServer().then((_) {
-          var request = http.Request('POST', serverUrl)
-            ..body = 'hello'
-            ..headers['User-Agent'] = 'Dart';
+  setUp(startServer);
 
-          expect(
-              request.send().then((response) {
-                expect(response.statusCode, equals(200));
-                return response.stream.bytesToString();
-              }).whenComplete(stopServer),
-              completion(parse(equals({
-                'method': 'POST',
-                'path': '/',
-                'headers': {
-                  'content-type': ['text/plain; charset=utf-8'],
-                  'accept-encoding': ['gzip'],
-                  'user-agent': ['Dart'],
-                  'content-length': ['5']
-                },
-                'body': 'hello'
-              }))));
-        }),
-        completes);
+  tearDown(stopServer);
+
+  test('.send', () async {
+    final request = http.Request('GET', serverUrl)
+      ..body = 'hello'
+      ..headers['User-Agent'] = 'Dart';
+
+    final response = await request.send();
+
+    expect(response.statusCode, equals(200));
+    final bytesString = await response.stream.bytesToString();
+    expect(
+        bytesString,
+        parse(equals({
+          'method': 'GET',
+          'path': '/',
+          'headers': {
+            'content-type': ['text/plain; charset=utf-8'],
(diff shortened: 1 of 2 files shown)


test/io/request_test.dart
```diff
@@ -10,61 +10,48 @@ import 'package:test/test.dart';
...
+            'content-type': ['text/plain; charset=utf-8'],
+            'accept-encoding': ['gzip'],
+            'user-agent': ['Dart'],
+            'content-length': ['5']
+          },
+          'body': 'hello',
+        })));
+  });
+
+  test('#followRedirects', () async {
```
jakemac53: nit: change the name while you are at it? `#followRedirects = false` or something? It is testing the opposite of what it sounds like right now.

jakemac53: Maybe there should be a real follow redirects test also?

test/io/request_test.dart
```diff
@@ -10,61 +10,48 @@ import 'package:test/test.dart';
...
-            expect(error.redirects.length, equals(2));
-          });
-          expect(
-              future.catchError((_) {}).then((_) => stopServer()), completes);
-          expect(future, completes);
-        }),
-        completes);
+        request.send(),
+        throwsA(isA<http.ClientException>()
+            .having((e) => e.message, 'message', 'Redirect limit exceeded')));
```
jakemac53: This used to also test that the number of redirects was correct, which is worth restoring

natebosch: `package:http` doesn't make that possible, we hide the original exception and keep only it's message.

We could add a `Object originalException` to `ClientException`, but it would be a new feature and we'd want to think it through.

natebosch: Note that I did call this out in the commit message 😉 

> We also lose the underlying
> exception and have only the message so we can't check the length of
> the `redirects` field.

jakemac53: So was this a problem before as well? Like the test before never would have gotten the expected error type?

jakemac53: I suppose you could check on the server side how many requests came through but likely not worth doing.

natebosch: > So was this a problem before as well? Like the test before never would have gotten the expected error type?

Yes, previously adding a `expectAsync1` around the error callback would make the test time out. Fixing the request type to `'GET'` would then make it fail the first `expect` because it gets a `HttpClientException` and not a `RedirectException`.

test/io/request_test.dart
natebosch: sure, I can add one in a followup

test/io/request_test.dart
jakemac53: :man_facepalming:

PR #318, opened 2019-09-13
Add a test for a successful redirect

natebosch:


natebosch:
Ugh, I didn't notice that the Dart version is included in the request and our test server just parrots back the request, so we can't check the entire thing. I'll just check the `path` part since that's what we care about.

PR #319, opened 2019-09-13
Drop an obsolete hack for async as argument name

natebosch:
This was working around a bug where in an `async` method you couldn't
use `async` as a named argument in a call.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: dart
 
 dart:
   - dev
-  - 2.2.0
+  - 2.4.0
 
 dart_task:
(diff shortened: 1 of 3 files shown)


natebosch:
Gross, SDKs as late as `2.3.0` still complain about this :(

https://github.com/dart-lang/sdk/issues/24637#issuecomment-531399985

I can either bump the min SDK or leave this in for a while... My preference is to bump the SDK - any thoughts @jakemac53 ?

jakemac53:
bump it :fist_right: :fist_left:

PR #321, opened 2019-09-15
Fixes #293, added property `url` to BaseResponse.

### LINKED ISSUE #293: Get final redirected URL from response

YizhePKU:
The final url is extract from XMLHttpRequest.responseURL(on browsers)
or HttpClientResponse.redirects.last.location(on vm).
Also added tests for vm.
No tests were written for browser, since server tests for brower is
broken and needs to be fixed.


piotrsed (approved):

natebosch:

PR #328, opened 2019-10-02
Refactor to async/await

natebosch:
Use an `async*` for `MultipartRequest.finalize()`. 

Refactor `.then` chains to async methods.

Add a test demonstrated that errors are forwarded though
a multipart request finalize stream.


natebosch:
I added a test for what I thought was the new behavior. As it turns out the TODO was outdated and errors already were forwarded correctly. I haven't been able to set up a scenario where an error isn't forwarded.

natebosch:
Whoa, I didn't know that `Stream.error` and `Stream.value` were introduced in `2.5.0`...

PR #354, opened 2019-12-28
DELETE method now accepts an optional body parameter

kwent:
Cf. https://github.com/dart-lang/http/issues/206

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.0+4
+
+* DELETE method now accepts an optional body parameter.
+
 ## 0.12.0+3
 
 * Documentation fixes.
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -129,6 +129,8 @@ Future<Response> patch(url,
 /// content-type of the request will be set to
 /// `"application/x-www-form-urlencoded"`; this cannot be overridden.
 ///
+/// [encoding] defaults to [utf8].
+///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
 /// the same server, you should use a single [Client] for all of those requests.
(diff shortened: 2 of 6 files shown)


kevmoo:
Will need an update to the changelog and the pubspec!

kwent:
@kevmoo done !

kwent:
Hi, how could i help to ship this ?

kevmoo:
@natebosch thoughts here?

@kwent – please rebase on the latest bits...

kwent:
Sure thing ! Done

natebosch:
I don't have strong opinions here. I'm somewhat inclined to say we should not do this since it's a feature that should generally not be used and there is a workaround by manually constructing a request and using the `send` API. Generally I prefer that odd use cases like this aren't encoded in the happy path, and instead are pushed to an odd solution.

In other words I agree with https://github.com/dart-lang/http/issues/206#issuecomment-434265328 and https://github.com/dart-lang/http/issues/206#issuecomment-453254835

Perhaps @mit-mit can help us find a reviewer who can put more time towards thinking about this and come to a decision.

kwent:
Per https://tools.ietf.org/html/rfc7231#section-4.3.5

```
   A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.
```

Does not say it's not allowed tho...

kwent:
I also don’t see that issue in a production setup. My guess is that the dev hot reloading in rails is not reloading the adapter postgis adapter and so the model is reading raw values of the columns.

mit-mit:
@natebosch @zichangg thoughts on how to get a conclusion here?

zichangg:
To be honest, I'm against the PR. Two reasons:

1. This PR doesn't solve any problem, which can't be solved with the existing APIs.

2. It doesn't provide a concise/easy way of some common use cases. If "DELETE with payload" cases are widely used, it probably makes sense to update our APIs to save users from writing more codes. But I don't think this is the case.

3. It tries to support something "undefined". It is not as bad as violating the standard. But this is not a good idea, which potentially indicates our APIs relying undefined behaviors.

mit-mit:
Closing. @kwent sorry that we were not able to take this.

PR #359, opened 2020-01-03
Synchronously finalize headers

### LINKED ISSUE #351: MultipartRequest does not contain Content-Type header
Content-Type header is not included in request when sending MultipartRequest.

```dart
import 'package:http_parser/http_parser.dart' as mime;

void main() {
  _run();
}

Future<void> _run() async {
  final uri = Uri.parse("http://localhost:8080/aaa");
  final req = http.MultipartRequest("POST", uri);

  final files = <http.MultipartFile>[];
  files.add(await http.MultipartFile.fromPath(
    "textfile",
    "./main.dart",
    contentType: mime.MediaType("text", "plain"),
  ));

  final sres = await req.send();
  final res = await http.Response.fromStream(sres);
  print(res.body);
}
```

Request dump by server:

```
2019/12/27 14:58:54 POST /aaa HTTP/1.1
Host: localhost:8080
Accept-Encoding: gzip
Content-Length: 76
User-Agent: Dart/2.7 (dart:io)

--dart-http-boundary-GcF.kzHYrmQE0Gpl3ByUUa.WC6SNn7po0k+F3r3iFULEmDsBZ3W--
```

### env

* Dart VM version: 2.7.0 (Mon Dec 2 20:10:59 2019 +0100) on "macos_x64"
* http: 0.12.0+3

### additional

With http: 0.12.0+2, Content-Type header exists:

```
2019/12/27 14:55:36 POST /aaa HTTP/1.1
Host: localhost:8080
Accept-Encoding: gzip
Content-Length: 76
Content-Type: multipart/form-data; boundary=dart-http-boundary-Wq2xuOJiK-ZZ3QFowkbQWC0-3Dv+167P3sKF2eqOcxb3d4oXm7d
User-Agent: Dart/2.7 (dart:io)

--dart-http-boundary-Wq2xuOJiK-ZZ3QFowkbQWC0-3Dv+167P3sKF2eqOcxb3d4oXm7d--
```

natebosch:
Fixes #351

An `async*` method is not started until after the stream has a listener,
if a caller reads the headers before listening to the stream, they would
see the value before they have been set. Move the header assignment and
other finalization work to the wrapping method and keep the `async*`
method handling only the stream content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.12.0+4-dev
 
-* Internal changes.
+* Fix a bug setting the `'content-type'` header in `MultipartRequest`.
 
 ## 0.12.0+3
 
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -85,13 +85,15 @@ class MultipartRequest extends BaseRequest {
   /// Freezes all mutable fields and returns a single-subscription [ByteStream]
   /// that will emit the request body.
   @override
-  ByteStream finalize() => ByteStream(_finalize());
-
-  Stream<List<int>> _finalize() async* {
-    // TODO(nweiz): freeze fields and files
-    var boundary = _boundaryString();
+  ByteStream finalize() {
+    // TODO: freeze fields and files
+    final boundary = _boundaryString();
     headers['content-type'] = 'multipart/form-data; boundary=$boundary';
     super.finalize();
+    return ByteStream(_finalize(boundary));
+  }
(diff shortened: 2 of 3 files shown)


natebosch:
> It is likely worth writing a regression test for this if possible

Yup, I'm trying to figure out the best way to do that.

natebosch:
I added a test that fails before the change and passes after.

I put the test in with the IO client tests since it's really an issue of the interaction between the IO client and `MultipartRequest`. 

PTAL

PR #365, opened 2020-01-14
Remove usages of Delegating*.typed

natebosch:
These usages all pass static type checking without being wrapped in
extra layers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 0.12.1-dev
+
 ## 0.12.0+4
 
 * Fix a bug setting the `'content-type'` header in `MultipartRequest`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.12.1-dev
 
+* Remove dependency on `package:async`.
+
 ## 0.12.0+4
 
 * Fix a bug setting the `'content-type'` header in `MultipartRequest`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 0.12.1-dev
+
```
grouma: Need an entry?

pubspec.yaml
```diff
@@ -1,13 +1,12 @@
...
-version: 0.12.0+4
+version: 0.12.1-dev
 homepage: https://github.com/dart-lang/http
 description: A composable, multi-platform, Future-based API for HTTP requests.
 
 environment:
   sdk: ">=2.4.0 <3.0.0"
 
 dependencies:
-  async: ">=1.10.0 <3.0.0"
```
kevmoo: One less dependency! 💯

CHANGELOG.md
kevmoo: Non-user visible change. I think not...

natebosch: I guess the removal of the `async` dep is technically user facing... In general for changes which are not user facing our new approach is to require the section header, but not necessarily an entry.

https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance

pubspec.yaml
natebosch: In practice pretty much everyone is going to pick up `async` transitively, but yeah it's always nice to make the graph a little less connected 😄

PR #385, opened 2020-02-20
Implements #295 multipart_request custom boundary

DartMen:
Useful to exactly mimic web browser file uploads #295.

### DIFF
--- a/lib/src/multipart_request.dart
+++ b/lib/src/multipart_request.dart
@@ -46,18 +46,28 @@ class MultipartRequest extends BaseRequest {
   /// The list of files to upload for this request.
   final files = <MultipartFile>[];
 
-  MultipartRequest(String method, Uri url) : super(method, url);
+  /// Set a custom boundary.
+  final String _customBoundary;
+
+  MultipartRequest(String method, Uri url, {String boundary})
+      : _customBoundary = boundary,
+        super(method, url);
 
   /// The total length of the request body, in bytes.
   ///
   /// This is calculated from [fields] and [files] and cannot be set manually.
   @override
   int get contentLength {
+    var boundaryLength = _boundaryLength;
+    if (_customBoundary != null) {
+      boundaryLength = _customBoundary.length;
+    }
+
     var length = 0;
 
     fields.forEach((name, value) {
       length += '--'.length +
-          _boundaryLength +
+          boundaryLength +
           '\r\n'.length +
           utf8.encode(_headerForField(name, value)).length +
           utf8.encode(value).length +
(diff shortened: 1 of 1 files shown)


mit-mit:
@zichangg WDYT?

lib/src/multipart_request.dart
```diff
@@ -46,18 +46,28 @@ class MultipartRequest extends BaseRequest {
...
 
   /// The total length of the request body, in bytes.
   ///
   /// This is calculated from [fields] and [files] and cannot be set manually.
   @override
   int get contentLength {
+    var boundaryLength = _boundaryLength;
+    if (_customBoundary != null) {
+      boundaryLength = _customBoundary.length;
+    }
```
natebosch, line 64:
```suggestion
    final boundaryLength = _customBoundary?.length ?? _boundaryLength;
```

lib/src/multipart_request.dart
```diff
@@ -87,7 +97,7 @@ class MultipartRequest extends BaseRequest {
   @override
   ByteStream finalize() {
     // TODO: freeze fields and files
-    final boundary = _boundaryString();
+    var boundary = _customBoundary ?? _boundaryString();
```
natebosch, line 100:
```suggestion
    final boundary = _customBoundary ?? _boundaryString();
```

lib/src/multipart_request.dart
```diff
@@ -46,18 +46,28 @@ class MultipartRequest extends BaseRequest {
   /// The list of files to upload for this request.
   final files = <MultipartFile>[];
 
-  MultipartRequest(String method, Uri url) : super(method, url);
+  /// Set a custom boundary.
+  final String _customBoundary;
```
zichangg, line 50: Some validation checks are skipped. You should run checks for the length(1 to 70 are valid) and invalid characters. Throw an exception if `_customBoundary` is not valid.

two possible ways:
1. Create getter and setter, validate inside setter.
2. Inside constructor.
I don't have strong opinion here. If you think `final` is necessary, put checks inside constructor. Otherwise, create a pair of setter and getter.

            The only mandatory parameter for the multipart  Content-Type
            is  the  boundary  parameter,  which  consists  of  1  to 70
            characters from a set of characters known to be very  robust
            through  email  gateways,  and  NOT ending with white space.
            (If a boundary appears to end with white  space,  the  white
            space  must be presumed to have been added by a gateway, and
            should  be  deleted.)   It  is  formally  specified  by  the
            following BNF:
          
            boundary := 0*69<bchars> bcharsnospace
          
            bchars := bcharsnospace / " "
          
            bcharsnospace :=    DIGIT / ALPHA / "'" / "(" / ")" / "+"  /
            "_"
                           / "," / "-" / "." / "/" / ":" / "=" / "?"

mit-mit:
Closing as assumed stale

Issue #387, opened 2020-02-25
Support detachSocket() in IOClient responses.


jamiewest:
Currently there is no way to detach a socket from IOClient responses. This is important when needing to upgrade a socket (i.e. `WebSocket.fromUpgradedSocket()`). 

To provide this ability we could extend the `StreamedResponse` class to allow for this method through the `IOClient`. 

IOClient Implementation:
```dart
class IOClient extends http.BaseClient {
  /// The underlying `dart:io` HTTP client.
  HttpClient _inner;

  IOClient([HttpClient inner]) : _inner = inner ?? HttpClient();

  @override
  Future<StreamedResponse> send(BaseRequest request) async {

   ...

       return IOStreamedResponse(
          response.handleError(
              (HttpException error) =>
                  throw ClientException(error.message, error.uri),
              test: (error) => error is HttpException),
          response.statusCode,
          contentLength:
              response.contentLength == -1 ? null : response.contentLength,
          request: request,
          response: response,
          headers: headers,
          isRedirect: response.isRedirect,
          persistentConnection: response.persistentConnection,
          reasonPhrase: response.reasonPhrase);
    } on HttpException catch (error) {
      throw ClientException(error.message, error.uri);
    }
  }
}
```

IOStreamResponse Implementation:
```dart
class IOStreamedResponse extends StreamedResponse {
  HttpClientResponse _inner;
  
  /// The stream from which the response body data can be read.
  ///
  /// This should always be a single-subscription stream.
  final ByteStream stream;

  /// Creates a new streaming response.
  ///
  /// [stream] should be a single-subscription stream. 
  IOStreamedResponse(Stream<List<int>> stream, int statusCode,
      {int contentLength,
      BaseRequest request,
      Map<String, String> headers = const {},
      bool isRedirect = false,
      bool persistentConnection = true,
      String reasonPhrase,
      HttpClientResponse response
      })
      : stream =  toByteStream(stream),
        super(stream, statusCode,
            contentLength: contentLength,
            request: request,
            headers: headers,
            isRedirect: isRedirect,
            persistentConnection: persistentConnection,
            reasonPhrase: reasonPhrase) {
              _inner = response;
            }

  Future<Socket> detachSocket() async {
    return await _inner.detachSocket();
  }

  static ByteStream toByteStream(Stream<List<int>> stream) {
    if (stream is ByteStream) return stream;
    return ByteStream(stream);
  }
}
````

Usage:
```dart
var response = await client.send(request);
[...]

PR #388, opened 2020-02-26
add io_streamed_response

jamiewest:
Addresses #387, adds support for detaching a socket when using an IOClient.


jamiewest:
@googlebot I signed it!

jamiewest:
cc @kevmoo, @natebosch

lib/src/io_streamed_response.dart
```diff
@@ -0,0 +1,40 @@
...
+      HttpClientResponse inner})
+      : super(stream, statusCode,
+            contentLength: contentLength,
+            request: request,
+            headers: headers,
+            isRedirect: isRedirect,
+            persistentConnection: persistentConnection,
+            reasonPhrase: reasonPhrase) {
+    _inner = inner;
+  }
```
natebosch:
```suggestion
      : _inner = inner,
        super(stream, statusCode,
            contentLength: contentLength,
            request: request,
            headers: headers,
            isRedirect: isRedirect,
            persistentConnection: persistentConnection,
            reasonPhrase: reasonPhrase);
```

lib/src/io_streamed_response.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import 'dart:io';
+
+import 'base_request.dart';
+import 'streamed_response.dart';
+
+/// An HTTP response where the response body is received asynchronously after
+/// the headers have been received.
+class IOStreamedResponse extends StreamedResponse {
+  HttpClientResponse _inner;
```
natebosch:
```suggestion
  final HttpClientResponse _inner;
```

lib/src/io_streamed_response.dart
```diff
@@ -0,0 +1,40 @@
...
+            isRedirect: isRedirect,
+            persistentConnection: persistentConnection,
+            reasonPhrase: reasonPhrase) {
+    _inner = inner;
+  }
+
+  /// Detaches the underlying socket from the HTTP server.
+  Future<Socket> detachSocket() async {
+    return await _inner.detachSocket();
+  }
```
natebosch:
```suggestion
  Future<Socket> detachSocket()=> _inner.detachSocket();
```

lib/src/io_streamed_response.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
```
natebosch:
```suggestion
// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```

jamiewest: Updated in 95e2ed5.

lib/src/io_streamed_response.dart
jamiewest: Updated in 95e2ed5

lib/src/io_streamed_response.dart
jamiewest: Updated in 95e2ed5

lib/src/io_streamed_response.dart
jamiewest: Updated in 95e2ed5

jamiewest:
@kevmoo and @natebosch, [package_resolver](https://github.com/dart-lang/package_resolver) does not like the version, [pubspec.yaml](https://github.com/dart-lang/package_resolver/blob/962014d7c80b3fadb03975d7d804d3fd856cbde9/pubspec.yaml#L13).

natebosch:
We don't need to move to `0.13.0`, we can use `0.12.1`. Pub's use of semver is a little unusual with the leading number is `0`. The move from `0.12.0` to `0.13.0` is considered a breaking major version bump.

jamiewest:
That makes sense, thanks @natebosch.

jamiewest:
@natebosch sorry one other question, do I need to keep the ‘-dev‘ on the version as well?

lib/src/io_client.dart
```diff
@@ -44,7 +45,7 @@ class IOClient extends BaseClient {
         headers[key] = values.join(',');
       });
 
-      return StreamedResponse(
+      return IOStreamedResponse(
```
natebosch: Let's change the return type on line 28 to `Future<IOStreamedResponse>`.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+
+* Adds a new return type for `IOClient` that includes the ability to detach a 
+  socket. To use or reference `IOStreamedResponse` directly, you will need to 
+  import the new `package:http/io_streamed_response.dart` import. Then cast 
+  a `StreamedResponse` as an `IOStreamedReesponse` to call the `detachSocket()`
+  method. 
+
 ## 0.12.1-dev
 
 * Remove dependency on `package:async`.
```
natebosch: If we change the return type we won't need to ask people to cast. If the client they are using is statically an `IOClient` then the response will also be statically an `IOStreamedResponse`. No cast should be necessary.

Also, the imports mentioned here don't line up with the implementation. Finally, I don't think we need the changelog section.

```suggestion
## 0.12.1-dev

* Add `IOStreamedResponse` which includes the ability to detach the socket.
  When sending a request with an `IOClient` the response will be an
  `IOStreamedResponse`.
* Remove dependency on `package:async`.
```

test/io/client_test.dart
```diff
@@ -9,6 +9,7 @@ import 'dart:io';
 
 import 'package:http/http.dart' as http;
 import 'package:http/src/io_client.dart' as http_io;
+import 'package:http/src/io_streamed_response.dart' as http_io_response;
```
natebosch: I think if we import `package:http/io_client` instead then we can pick up both types we need in one place, and it would better represent real uses of this package.
```suggestion
import 'package:http/io_client.dart' as http_io;
```

jamiewest: 👍 Updated in b699b87.

CHANGELOG.md
jamiewest: 👍 This is much cleaner, thanks! Updated in b699b87.

lib/src/io_client.dart
jamiewest: 👍 Updated in b699b87.

natebosch:
Thanks!

I'll publish after some more internal validation.

jamiewest:
Thanks for all your help!

PR #411, opened 2020-04-28
Adding possibility of send body into a request using get method

lucas-tomasi:
GET method now accepts an optional body parameter


mit-mit:
Closing as the CLA is not signed

PR #435, opened 2020-06-27
Fix README.md

tbm98:


tbm98:
> @googlebot I signed it!

natebosch:
The example code is correct and the modification would make it incorrect. Did you hit this because you specifically needed `implements http.IOClient` (that isn't what the example is for)?

I'm not sure that specific use case is worth documenting. It's not a pattern we generally recommend and the analyzer should already give make it clear what needs to happen.

Why did you need something that implemented the IO specific interface?

tbm98:
Hi.
My code in old version work normal but after update package http, it
require add prefix `IO`

Vào 3:23 Th 3, 30 thg 6 2020 Nate Bosch <notifications@github.com> đã viết:

tbm98:
> The example code is correct and the modification would make it incorrect. Did you hit this because you specifically needed `implements http.IOClient` (that isn't what the example is for)?
> 
> I'm not sure that specific use case is worth documenting. It's not a pattern we generally recommend and the analyzer should already give make it clear what needs to happen.
> 
> Why did you need something that implemented the IO specific interface?

Sorry about that, I used IOClient.
but for example already use `http.BaseClient` and `http.Client`
It should also be `http.StreamedResponse` and `http.BaseRequest`

natebosch:
Thanks!

Issue #383, opened 2020-02-14
Cannot send body for Delete request !


meysam810:

bambinoua:
https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

PR #439, opened 2020-07-03
added optional body to delete

juumixx:
As requested in issues
https://github.com/dart-lang/http/issues/383
and
https://github.com/dart-lang/http/issues/206

### DIFF
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -125,8 +125,10 @@ Future<Response> patch(url,
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> delete(url, {Map<String, String> headers}) =>
-    _withClient((client) => client.delete(url, headers: headers));
+Future<Response> delete(url,
+        {Map<String, String> headers, body, Encoding encoding}) =>
+    _withClient((client) =>
+        client.delete(url, headers: headers, body: body, encoding: encoding));
 
 /// Sends an HTTP GET request with the given headers to the given URL, which can
 /// be a [Uri] or a [String], and returns a Future that completes to the body of
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -42,8 +42,9 @@ abstract class BaseClient implements Client {
       _sendUnstreamed('PATCH', url, headers, body, encoding);
 
   @override
-  Future<Response> delete(url, {Map<String, String> headers}) =>
+  Future<Response> delete(url, {Map<String, String> headers, body}) =>
       _sendUnstreamed('DELETE', url, headers);
+
   @override
   Future<String> read(url, {Map<String, String> headers}) async {
     final response = await get(url, headers: headers);
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -42,8 +42,9 @@ abstract class BaseClient implements Client {
       _sendUnstreamed('PATCH', url, headers, body, encoding);
 
   @override
-  Future<Response> delete(url, {Map<String, String> headers, body}) =>
-      _sendUnstreamed('DELETE', url, headers);
+  Future<Response> delete(url,
+          {Map<String, String> headers, body, Encoding encoding}) =>
+      _sendUnstreamed('DELETE', url, headers, body, encoding);
 
   @override
   Future<String> read(url, {Map<String, String> headers}) async {
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -44,7 +44,7 @@ abstract class BaseClient implements Client {
   @override
   Future<Response> delete(Object url,
           {Map<String, String>? headers, Object? body, Encoding? encoding}) =>
-      _sendUnstreamed('DELETE', url, headers);
+      _sendUnstreamed('DELETE', url, headers, body, encoding);
 
   @override
   Future<String> read(Object url, {Map<String, String>? headers}) async {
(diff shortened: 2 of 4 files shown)


benedictjohannes (changes requested):
Well, on a second look, it looks like the delete method still *not* send the body to the `_sendUnstreamed` function. The body should be forwarded to `_sendUnstreamed`. With it, apart from the `BaseClient`, the `Client` class should also be updated. 

You should look into it.

lib/src/base_client.dart
```diff
@@ -42,8 +42,10 @@ abstract class BaseClient implements Client {
       _sendUnstreamed('PATCH', url, headers, body, encoding);
 
   @override
-  Future<Response> delete(url, {Map<String, String> headers}) =>
-      _sendUnstreamed('DELETE', url, headers);
+  Future<Response> delete(url,
+          {Map<String, String> headers, body, Encoding encoding}) =>
+      _sendUnstreamed('DELETE', url, headers, encoding);
```
benedictjohannes, line 47: the `body` should also be forwarded to `_sendUnstreamed`. The `Client` abstract class should also be updated to include the updated method, and other classes implementing `Client` would probably need to be checked to be  updated to ensure `body` (and `encoding`) parameters passed are actually forwarded to `Request` method.

benedictjohannes (changes requested):

lib/src/base_client.dart
juumixx: Sorry, somehow missed that in the rebase. It should be fixed now.

juumixx:
As @benedictjohannes suggested, I also added the additional parameters to the global function and abstract Client class, as opposed to only the BaseClient in the last commit.

benedictjohannes (approved):
So far it looks good, I'd vouch for this PR being merged

lib/http.dart
```diff
@@ -125,8 +125,10 @@ Future<Response> patch(url,
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> delete(url, {Map<String, String> headers}) =>
-    _withClient((client) => client.delete(url, headers: headers));
+Future<Response> delete(url,
+        {Map<String, String> headers, body, Encoding encoding}) =>
```
valehasadli, line 129: Could you tell me please, what is the exact type definition for the body?

benedictjohannes: I can directly send a Map to body, which tries to convert it to (JSON) String, or String. The body field work the same across HTTP verb methods, where it's not included in the GET or OPTION as these verbs doesn't support request body and inside it. Meanwhile, DELETE does support body, hence this pull request. 

You can observe that type definition for body isn't also referenced in other methods in this file. Hence my support for this PR

AhmedKhattab95:
any updates on this

leecommamichael:
Please let this be merged. It's a huge context switch to go all the way from my editor to this PR.

benedictjohannes:
@natebosch can you please review and/or merge this? Its been a while.

natebosch:
We've had this request a few times so far, and each time we push back against the idea of adding an argument that should _almost never_ be used. Especially since there is an existing workaround of using `client.send` and constructing the request manually. As a general principle we prefer to support the common case directly, and uncommon cases with workarounds, so that we don't clutter the common API with uncommon complexity.

The big thing was it is also a breaking change.

Our current situation might push us to adopt this API:
- The complexity cost is fairly low, especially since, as was mentioned in this thread, adding the argument makes this match other similar methods.
- This seems to be coming up frequently.
- The big one - we are about to make a breaking version change anyway for the migration to null safety.

cc @lrhn in case you have any thoughts or feel strongly that we shouldn't do this.

I think I'm leaning towards getting this updated and merged. It will take a bit of time to verify that I can get it landed without breaking internal users or the SDK.

benedictjohannes:
@natebosch This does not look like it will break API compatibility since the `body` parameter is optional. IMHO `body` in `DELETE` the fact that HTTP spec does support it should be enough to make the go-to HTTP handler in any language (this package) to support it.  PPS: I've seen several API expects `body` in `DELETE`, and I develop (backend) applications with this pattern.

lib/http.dart
lrhn: We lack documentation of the `body` parameter on the other similar methods too, and the documentation in general isn't amazing, so we should give it an overhaul soon.
No need to do anything *special* here, it's the same `body` as all the other `body` parameters.

lrhn:
The top-level on is definitely non-breaking. The instance method is not something I would consider a breaking change. Nobody else is intended or expected to implement the interface. This is a tool package, intended to be used as-is.

Adding it during a major version increment makes it doubly not breaking, so if it's a change we do want, now is definitely the time to do it.

I think it's a fine addition, and I'm fine adding it now.

natebosch:
> The instance method is not something I would consider a breaking change. Nobody else is intended or expected to implement the interface. This is a tool package, intended to be used as-is.

The interface is specifically designed for extending or implementing.

https://github.com/dart-lang/http/blob/d473635f0dbb72936664696162127c2209986356/README.md#L49-L52

> I think it's a fine addition, and I'm fine adding it now.

 There are at least a few implementations of this class, though I need to check whether any override this method. Please wait to add this until we get things prepped.

lib/http.dart
benedictjohannes: Do you have plans regarding the documentation overhaul? If you share your plans the community might be able to help.

natebosch (approved):

jakemac53 (approved):

natebosch:
Thanks everyone for the contributions and discussion!

This will roll out once we are ready to publish the null safety migration for this package. https://github.com/dart-lang/http/issues/501

PR #470, opened 2020-09-22
Remove unused dart:async imports

MichaelRFairhurst:
Since Dart 2.1, Future and Stream have been exported from dart:core


jakemac53 (approved):

mit-mit:
Looks like we can merge this @MichaelRFairhurst ?

PR #472, opened 2020-09-24
Enable null safety

kevmoo:


kevmoo:
Sadly, `test/streamed_request_test.dart` is failing on safari – but all of the other tests are passing

lib/src/io_streamed_response.dart
```diff
@@ -10,19 +10,19 @@ import 'streamed_response.dart';
 /// An HTTP response where the response body is received asynchronously after
 /// the headers have been received.
 class IOStreamedResponse extends StreamedResponse {
-  final HttpClientResponse _inner;
+  final HttpClientResponse? _inner;
```
kevmoo, line 13: This is funny. When this class is instantiated in this package, `_inner` is always non-null.

... but it's also exposed as a Type publicly. So switching `inner` to be required in the ctor is a breaking change. 🤷

lib/src/utils.dart
```diff
@@ -12,11 +12,11 @@ import 'byte_stream.dart';
 ///
 ///     mapToQuery({"foo": "bar", "baz": "bang"});
 ///     //=> "foo=bar&baz=bang"
-String mapToQuery(Map<String, String> map, {Encoding encoding}) {
+String mapToQuery(Map<String, String> map, {Encoding? encoding}) {
   var pairs = <List<String>>[];
   map.forEach((key, value) => pairs.add([
-        Uri.encodeQueryComponent(key, encoding: encoding),
-        Uri.encodeQueryComponent(value, encoding: encoding)
+        Uri.encodeQueryComponent(key, encoding: encoding ?? utf8),
```
kevmoo, line 18: `Uri.encodeQueryComponent` `encoding` param is now non-null with a default.

So we have to hoist the default value here for a `null` `encoding`

lib/src/browser_client.dart
```diff
@@ -52,16 +52,17 @@ class BrowserClient extends BaseClient {
     request.headers.forEach(xhr.setRequestHeader);
 
     var completer = Completer<StreamedResponse>();
+
+    // TODO(kevmoo): not sure what to do here...
+    // ignore: void_checks
     unawaited(xhr.onLoad.first.then((_) {
-      // TODO(nweiz): Set the response type to "arraybuffer" when issue 18542
```
kevmoo, line 56: This issue is now closed

lib/src/browser_client.dart
```diff
@@ -52,16 +52,17 @@ class BrowserClient extends BaseClient {
...
-      // is fixed.
-      var blob = xhr.response as Blob ?? Blob([]);
+      var blob = xhr.response as Blob;
```
kevmoo, line 59: According to the annotations on the `dart:html` types, `response` is never null.

lib/src/browser_client.dart
```diff
@@ -52,16 +52,17 @@ class BrowserClient extends BaseClient {
     request.headers.forEach(xhr.setRequestHeader);
 
     var completer = Completer<StreamedResponse>();
+
+    // TODO(kevmoo): not sure what to do here...
+    // ignore: void_checks
```
kevmoo, line 57: Here and below – not sure why we're getting this `void_checks` hint...

lib/src/base_request.dart
```diff
@@ -99,7 +99,7 @@ abstract class BaseRequest {
   /// single-subscription to ensure that no data is dropped. They should also
   /// freeze any additional mutable fields they add that don't make sense to
   /// change after the request headers are sent.
-  ByteStream finalize() {
+  ByteStream? finalize() {
```
kevmoo, line 102: Seems an interesting implementation detail that leaks around. Subclasses should *not* return `null` – but this does.

🤷

lib/src/base_request.dart
```diff
@@ -30,10 +30,10 @@ abstract class BaseRequest {
   ///
   /// This defaults to `null`, which indicates that the size of the request is
   /// not known in advance. May not be assigned a negative value.
-  int get contentLength => _contentLength;
-  int _contentLength;
+  int? get contentLength => _contentLength;
+  int? _contentLength;
 
-  set contentLength(int value) {
+  set contentLength(int? value) {
```
jakemac53, line 36: This throws on null today, we should make it non-nullable instead.

We can use `ArgumentError.checkNotNull` to keep the throwing for legacy mode if we like.

lib/src/base_request.dart
jakemac53: Should this return an empty ByteStream or something?

lib/src/io_client.dart
```diff
@@ -18,26 +18,26 @@ BaseClient createClient() => IOClient();
 /// A `dart:io`-based HTTP client.
 class IOClient extends BaseClient {
   /// The underlying `dart:io` HTTP client.
-  HttpClient _inner;
+  HttpClient? _inner;
```
jakemac53, line 21: This looks like it is not nullable to me.

lib/src/base_request.dart
kevmoo: It doesn't throw on `null` – just if `<0`

lib/src/io_streamed_response.dart
jakemac53: It is not a breaking change to unmigrated code and if it is assumed to be non-null then it is also arguably not breaking (just breaks in a different way).

pubspec.yaml
```diff
@@ -1,15 +1,19 @@
 name: http
-version: 0.12.2
+version: 0.12.3-nullsafety
 homepage: https://github.com/dart-lang/http
 description: A composable, multi-platform, Future-based API for HTTP requests.
 
 environment:
-  sdk: ">=2.4.0 <3.0.0"
+  sdk: '>=2.10.0-0 <2.11.0'
```
jakemac53, line 8:
```suggestion

# Cannot be published until null safety ships
publish_to: none
```

pubspec.yaml
```diff
@@ -1,15 +1,19 @@
 name: http
-version: 0.12.2
+version: 0.12.3-nullsafety
```
jakemac53, line 2: I think we are generally going with breaking changes for non-allow-listed packages, although that isn't totally decided

lib/src/base_request.dart
kevmoo: good idea!

pubspec.yaml
kevmoo: We're going to HAVE to publish this, I think, to enable migration of flutter gallery – or are we going to just use git deps?

lib/src/io_streamed_response.dart
kevmoo: ack

kevmoo: @jakemac53 – well. Hrm. Or we just say that `detachSocket` doesn't work if `inner` was `null` when provided.

pubspec.yaml
natebosch: Please add the `publish_to: none` until we have a min SDK version that adds it to the allow list if we need to publish.

Publishing with a lower bound of `2.10.0-0` is not safe, and we need to prevent it from happening.

kevmoo: done!

Should I change the lower bound?

lib/src/base_request.dart
natebosch: +1, this looks like it needs to accept `null`.

pubspec.yaml
natebosch: No, there is no lower bound at which it is safe to publish until we have a known SDK version that supports it so keeping it at `2.10.0-0` is fine.

pubspec.yaml
jakemac53: Oh it sounds like from the changelog entry this is in the allow list, even so we really only don't want to do breaking changes for flutter depended on packages.

kevmoo: Flutter doesn't depend on this directly – at least I don't think so.

A lot of stuff in the Flutter ecosystem does, though...

jakemac53: cc @natebosch I would still do a major version bump.

We need to stop setting a bad example here.

natebosch: > Oh it sounds like from the changelog entry this is in the allow list

It isn't.

> I would still do a major version bump.

+1. We'll want a breaking change for https://github.com/dart-lang/http/issues/424 in any case. I can try to get to that next week so we can make a decision before we need to publish this.

kevmoo:
where are we on this review, kind sirs?

lib/src/browser_client.dart
natebosch: This will be worked around by https://github.com/dart-lang/pedantic/pull/69 if we get that in.

This issue looks like https://github.com/dart-lang/linter/issues/2185

I think we can keep the lint and point the todo at that issue.

lib/src/browser_client.dart
natebosch: What do you mean by this? I think it probably is true that if `onLoad.first` doesn't have an error this would be non-null, but I don't know what you mean by "the annotations on the `dart:html` types".

pubspec.yaml
```diff
@@ -1,15 +1,22 @@
...
```
natebosch, line 2:
```suggestion
version: 0.13.0-nullsafety-dev
```

kevmoo:
PTAL – cleaned things up

jakemac53:
Please don't force push because now we can no longer see what you changed compared to our last review :(

kevmoo:
Gah. Right. I wanted to revert a bunch of stuff and that seemed easiest.

lib/src/base_request.dart
jakemac53: Ah yes I misinterpreted it :D

jakemac53 (approved):

jhancock4d:
Is there a beta package with this enabled available? I can't find it and this is blocking in almost any project to being able to safely turn on null safety.

natebosch:
Publishing is tracked in https://github.com/dart-lang/http/issues/501#issuecomment-756449873

PR #474, opened 2020-09-26
Prepare publish 0.12.3

MichaelRFairhurst:
This has to be published & pulled into the dart SDK in order to be rolled into google3 right, now that http has been unforked?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.3
+
+* Remove unnecessary dart:async imports.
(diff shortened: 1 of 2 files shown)


natebosch:
We first pull to the SDK, then google3 with the sync, and then we can publish.

https://dart-review.googlesource.com/c/sdk/+/164902

natebosch:
Dropping this for now. If we need to do a backpublish we'll need to branch before null safety migration was merged.

PR #477, opened 2020-09-29
Add dependency override for pkg:test

kevmoo:
Because pkg:test does not allow pkg:http ^0.13.0

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -18,5 +18,7 @@ dev_dependencies:
   test: ^1.16.0-nullsafety.4
 
 dependency_overrides:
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -18,5 +18,7 @@ dev_dependencies:
   test: ^1.16.0-nullsafety.4
 
 dependency_overrides:
+  # Because pkg:test does not support pkg:http ^0.13.0
+  test: ^1.16.0-nullsafety.5
```
jakemac53, line 22: Note that just `^1.16.0-nullsafety` should be sufficient, although this is also likely fine.

jakemac53 (approved):

PR #478, opened 2020-09-29
Enabled and fix a number of lints

kevmoo:
Including require API types so we are clear
about which params are not allowed to be null – like `url`

### DIFF
--- a/example/main.dart
+++ b/example/main.dart
@@ -1,7 +1,7 @@
 import 'dart:convert' as convert;
 import 'package:http/http.dart' as http;
 
-void main(List<String> arguments) async {
+Future<void> main(List<String> arguments) async {
   // This example uses the Google Books API to search for books about http.
   // https://developers.google.com/books/docs/overview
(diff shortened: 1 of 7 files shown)


example/main.dart
```diff
@@ -1,7 +1,7 @@
 import 'dart:convert' as convert;
 import 'package:http/http.dart' as http;
 
-void main(List<String> arguments) async {
+Future<void> main(List<String> arguments) async {
```
natebosch, line 4: Let's drop this lint instead. There are too many valid reasons to violate it.

natebosch (approved):

example/main.dart
kevmoo: Really? I think these are fine.

natebosch: I forwarded you an email with some discussion about this lint and why it will never be included in `package:pedantic`.

This is not one I want to enforce in any Dart team owned packages.

PR #485, opened 2020-10-16
BrowserClient - withCredentials 'true' by default.

nwtnwrkshp:
'withCredentials = true' is needed for browser cookies in Flutter Web XmlHttpRequests.
Because current pattern requires a parameterless constructor, only way to change the setting is to set it by default. 
There is no harm in having it always set to true - credentials are ignored if not present.

### DIFF
--- a/lib/src/browser_client.dart
+++ b/lib/src/browser_client.dart
@@ -36,8 +36,8 @@ class BrowserClient extends BaseClient {
   /// Whether to send credentials such as cookies or authorization headers for
   /// cross-site requests.
   ///
-  /// Defaults to `false`.
-  bool withCredentials = false;
+  /// Defaults to `true`.
+  bool withCredentials = true;
 
   /// Sends an HTTP request and asynchronously returns the response.
   @override


jamotaylor:
@googlebot I signed it!

kartercs (approved):

LucaIaconelli (approved):

kevmoo:
At a minimum we should have a changelog entry – and ideally a test.

I'm not super comfortable about changing the default here without some more info.

Thoughts @natebosch ?

@mit-mit would someone on your side of the pond want to look?

LucaIaconelli:
Any news about this @kevmoo? for Flutter web it would be very useful

Thanks

paxcodes:
I want to help move this along by creating a test. But I can't run tests in `html`

Dec-05-2020 09-14-11

I can, however run tests in `io`

Dec-05-2020 09-15-50

Is this expected behaviour?

kevmoo:
> I want to help move this along by creating a test. But I can't run tests in `html`
> 
> Dec-05-2020 09-14-11
> 
> I can, however run tests in `io`
> 
> Dec-05-2020 09-15-50
> 
> Is this expected behaviour?

You may need to run from the console: `pub run test -p vm,chrome`

kevmoo:
It looks like CI passed the tests fine

kevmoo:
@natebosch @jakemac53  – thoughts on this?

natebosch:
I don't have enough background to evaluate this. On the surface I'm uncomfortable with the idea of having a different default than the browser for anything related to security, but I can see the appeal of picking a more popular default rather than have a bunch of cross-platform code need to deal with a single-platform concern.

jlubeck:
If we can't get it to true by default, can we at least get a way to set it?? It's impossible to use this package with a cookies enabled backend on web until this is solved, which is a shame... I would hate to have to use another one after so many other succedsful implementations that I made with this awesome package.

Thank you!

LucaIaconelli:
Completely agreed with @jlubeck 🙏🙏

mikeesouth:
I'm also very interested in getting this merged and a new http release asap. cookie enabled backends are more or less standard for new pages so it will be very useful for Flutter for Web.

@jlubeck have you found any workaround? Alternative package or so.

jlubeck:
@mikeesouth not yet. I'm currently convincing my users (it's an internal system) to use iPads until this is solved, since on devices you can force the cookies in the headers. Browsers on the other hand disallow this practice, so it's not possible to work around it AFAIK

jakemac53:
Fwiw this is what it would look like to configure this on your own in a way that still works across platforms https://gist.github.com/jakemac53/204d6adfe04eafa7e4d432fa77b3b4b1 (I only implemented `get` but you would just do the same for the others).

This uses the exact same strategy as the http package itself uses, but with a custom browser client that sets `withCredentials` to true.

jakemac53:
We could consider exposing this `withCredentials` configuration on the top level methods, and document well that it only has any effect on the browser. Or we could allow configuring the default with a zone variable, although that gets a bit gross.

mikeesouth:
> Fwiw this is what it would look like to configure this on your own in a way that still works across platforms https://gist.github.com/jakemac53/204d6adfe04eafa7e4d432fa77b3b4b1 (I only implemented `get` but you would just do the same for the others).
> 
> This uses the exact same strategy as the http package itself uses, but with a custom browser client that sets `withCredentials` to true.

Nice, I ended up with the same solution after finding a similar BrowserClient cast in a StackOverflow thread (see below for link and credit). I've wrapped all my http calls in a custom client that I create like this:
```
http.Client _getHttpClient() {
  final http.Client cookieEnabledClient = http.Client();
  if (cookieEnabledClient is BrowserClient) {
    cookieEnabledClient.withCredentials = true;
  }

  return cookieEnabledClient;
}
```

This correctly sets response cookies in the browser, it also sends the cookies as request cookies to the service.

I'm not sure if these steps were necessary but I also took these steps in order for everything to work smoothly:
* I edited my hosts file and configured "127.0.0.1 app.myproject-dev.com"
* Start my flutter for web project with --web-hostname=app.myproject-dev.com and --web-port=8888
* On the server side I set `Access-Control-Allow-Origin=app.myproject-dev.com`

I found the solution in this SO thread, kudos to Piotr Temp https://stackoverflow.com/a/65039765/72950

I would of course really appreciate a more official solution to this so I'm still hoping this PR gets merged or that a similar fix is merged.

jakemac53:
@mikeesouth your solution works but I believe it requires an import of `package:http/browser_client.dart` which means it isn't cross platform - the solution I linked does work cross platform by using conditional imports. So that would be the main difference, although your solution is a bit easier if you only need it to work on the web.

mikeesouth:
> @mikeesouth your solution works but I believe it requires an import of `package:http/browser_client.dart` which means it isn't cross platform - the solution I linked does work cross platform by using conditional imports. So that would be the main difference, although your solution is a bit easier if you only need it to work on the web.

Nice, thanks for that clarification. I'm currently web only but I will move this to my shared library soon and then I need it to be cross platform. I'll use your suggested solution 👍

nwtnwrkshp:
> I'm also very interested in getting this merged and a new http release asap. cookie enabled backends are more or less standard for new pages so it will be very useful for Flutter for Web.
> 
> @jlubeck have you found any workaround? Alternative package or so.

I know this is a big no-no in almost all cases but I'm actually editing the package source code on my local machine to get it to work.

Also, as part of my CI build, I replace that line of code. It's a horrible work-around I know, but I had to get it working. Project is currently web only and not in production.

nwtnwrkshp:
> I'm not sure if these steps were necessary but I also took these steps in order for everything to work smoothly:
> 
> * I edited my hosts file and configured "127.0.0.1 app.myproject-dev.com"
> * Start my flutter for web project with --web-hostname=app.myproject-dev.com and --web-port=8888
> * On the server side I set `Access-Control-Allow-Origin=app.myproject-dev.com`

@mikeesouth In your above steps
* You don't need to edit hosts file for cookies to work in the browser (this is tested with Chrome and Edge), it will work on localhost
* Having a fixed debugging port is only necessary for cross-origin requests
* Allowing CORS is not related to cookies

shumiyao (approved):

shumiyao:
Can someone eligible approve the review?

nwtnwrkshp:
> ~Can someone eligible approve the review?~
> 
> Sorry, my mistake. I tried http.post method by modifying the local file (withCredentials = true), and it caused an error without showing any clues about it (in the Chrome inspector, it says "CORS error"). Is changing the withCredentials default value to true a truly safe change?

@shumiyao Yes, it is a safe change. The CORS error is unrelated - that is a server-side setting. To enable CORS on your local dev environment, set up the web server to allow XHR requests from whichever port Flutter Web is running on e.g. http://localhost:12345. As posted previously, you will have to use the `--web-port = 12345` command to set the port that Flutter Web runs on.

shumiyao:
@nwtnwrkshp 

Thank you for a follow up. I think that adding more details on the caught error might be out of scope of this issue.

nwtnwrkshp:
@kevmoo Has there been any further discussion on this? With Flutter Web having moved to stable, surely this is a key step in allowing API integration?

motu2119:
Can't merge this pull request yet? I'm waiting for it to be merged.

kevmoo:
I'm really hesitant to change the default of a value with security implications.

See https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

Thoughts @natebosch @jakemac53 @mit-mit ?

nwtnwrkshp:
> I'm really hesitant to change the default of a value with security implications.
> 
> See https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials
> 
> Thoughts @natebosch @jakemac53 @mit-mit ?

Ok fair enough. I can see the security loophole in the below paragraph (from linked docs), where a responding endpoint could set unwanted cookies:

> In addition, this flag is also used to indicate when cookies are to be ignored in the response. 

So if this is the case, what alternative approach would allow the inclusion of credentials to solve #485?

nwtnwrkshp:
Will close this because it's not a complete solution.

BunnyTheLifeguard:
@jakemac53 Thanks for the gist. Your solution looked nice on paper, but did not work for unfortunately. Did something change that maked it need some readjustments?

This issue needs more attention btw. is 2022 and still such a basic functionality missing 😕

jakemac53:
Updated the gist for null safety + changes to the `url` parameter in the http package.

maheini:
I made a simple demo Flutter in flutter web, based on the solution from @jakemac53 (thanks) -> have a look inside main.dart: [Here is the Repo](https://github.com/maheini/flutter_web_cookie_support)

jakemac53:
> I made a simple demo Flutter in flutter web, based on the solution from @jakemac53 (thanks) -> have a look inside main.dart: [Here is the Repo](https://github.com/maheini/flutter_web_cookie_support)

Note that your solution there doesn't use conditional imports - it will _only_ work on flutter web and not elsewhere.

maheini:
Yes, that's true, thanks for this note. I will add a comment inside the main.dart.

BunnyTheLifeguard:
But what's the point of this then? Weren't httponly cookies specifically made to not be accessible via code and therefore should be automatically be attached to subsequent requests? 🤔

maheini:
@BunnyTheLifeguard Hmm... I'm not totally familiar with cookies in general, but as far as I know, you have (or can) fetch Cookies inside the "Set-cookie" header via http request on all platforms except web. My demo is just for making cookie storage inside the browser possible.

BunnyTheLifeguard:
@maheini I don't know how you guys do it, but no cookies for me:

```dart
Future<Map<String, dynamic>> login(String email, String password) async {
    dynamic result;

    _loggedInStatus = Status.authenticating;
    notifyListeners();

    Client client = Client();
    if (client is BrowserClient) {
      client.withCredentials = true;
    }

    Response response = await client.post(loginURI,
        headers: baseHeaders,
        body: jsonEncode(<String, String>{
          'email': email,
          'password': password,
        }));
...
```

Started with `flutter run -d chrome --web-port 3000 --web-hostname 127.0.0.1`, also using chromium based Brave, but this should not be an issue, no?

maheini:
Seems legit for me...
-Have you tested your API with a browser extension like RESTer? Does this request return a "Set-cookie" header?
-Which response do you get from your post call? a successful code?
-Side note: The provided solution will only set a cookie on your browser, but you will never get access to it in your code/ response. So you might not notice it, even if it's set.

c-seeger:
facing the same problem all CORS header set + withCredentials, getting a valid http 200 back from my login endpoint, Chrome shows both cookies without any errors in the response cookie tab, but they are not stored.

c-seeger:
I made it! I forgot to set `withCredentials` for the follow up requests.

PR #489, opened 2020-10-29
Migrate BrowserClient from blob to arraybuffer

mmotorny:
This allows BrowserClient to be used on Cobalt (which
doesn't support blob in XHR).

"blob" was used as a workaround for the Dart issue #18542.
Since the issue is closed, using "arraybuffer" is more
straightforward and efficient.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,7 @@ homepage: https://github.com/dart-lang/http
 description: A composable, multi-platform, Future-based API for HTTP requests.
 
 environment:
-  sdk: '>=2.10.0-2.0.dev <2.11.0'
+  sdk: '>=2.12.0-0 <2.13.0'
 
 # Cannot be published until null safety ships
 publish_to: none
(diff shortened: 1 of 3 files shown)


jakemac53:
Please also update the CHANGELOG.md file, otherwise this LGTM

mmotorny:
Jake, please take another look.

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ homepage: https://github.com/dart-lang/http
 description: A composable, multi-platform, Future-based API for HTTP requests.
 
 environment:
-  sdk: '>=2.10.0-2.0.dev <2.11.0'
+  sdk: '>=2.10.0-2.0.dev <2.13.0'
```
jakemac53, line 7:
```suggestion
  sdk: '>=2.12.0-0 <2.13.0'
```

jakemac53 (approved):

mmotorny:
All checks passed. Please merge.

pubspec.yaml
mmotorny: Done.

Issue #492, opened 2020-11-04
Unhandled exception: NoSuchMethodError: The getter 'length' was called on null. Receiver: null Tried calling: length


ArpanBhavsar:
On passing null value in map it stops the request and show an unhandled exception.

```
Unhandled exception:
NoSuchMethodError: The getter 'length' was called on null.
Receiver: null
Tried calling: length
#0      Object.noSuchMethod  (dart:core-patch/object_patch.dart:51:5)
#1      _Uri._uriEncode  (dart:core-patch/uri_patch.dart:44:23)
#2      Uri.encodeQueryComponent  (dart:core/uri.dart:1103:17)
#3      mapToQuery.<anonymous closure> 
package:http/src/utils.dart:19
#4      CastMap.forEach.<anonymous closure>  (dart:_internal/cast.dart:288:8)
#5      _LinkedHashMapMixin.forEach  (dart:collection-patch/compact_hash.dart:377:8)
#6      CastMap.forEach  (dart:_internal/cast.dart:287:13)
#7      mapToQuery 
package:http/src/utils.dart:17
#8      Request.bodyFields= 
package:http/src/request.dart:137
#9      BaseClient._sendUnstreamed 
package:http/src/base_client.dart:85
#10     BaseClient.post 
package:http/src/base_client.dart:32
#11     post.<anonymous closure> 
package:http/http.dart:70
#12     _withClient 
package:http/http.dart:166
```

 It should show user that they have to handle that exception or else the null value should be set to any default value like ''.

natebosch:
I think nulls are best handled by the calling code, especially following the null safety migration. Here is a snippet that can handle the fallback if you need it:

```dart
Map<String, String> defaultNulls(Map<String, String> original) =>
    {for (var key in original.keys) key: original[key] ?? ''};
```

PR #493, opened 2020-11-04
Solved: Unhandled exception: NoSuchMethodError: The getter 'length' was called on null. Receiver: null Tried calling: length #492

ArpanBhavsar:
Solved issue: #492 by handling exception and added parameter named **passNull** to pass null value through map inside body. With **passNull=false** and **passNull=null** it will throw exception for notNull value inside body. And with **passNull=true** it will let the user make request with null inside body.


ArpanBhavsar:
@nex3 @sethladd @kevmoo @mdempsky please test and review my code.

natebosch:
With the null safety migration this is a `Map<String, String>`, not a `Map<String, String?>`. A `null` value should never end up in that map in a migrated app. I don't think it makes sense to allow `null` for opted out callers, especially since we didn't allow that before.

I'd recommend handling the defaulting manually before calling into the http package. Adding an extra argument is more API noise than it's worth, and it's also breaking.

Thanks for the discussion!

PR #496, opened 2020-11-09
Prepare to publish null safe version

kevmoo:


pubspec.yaml
```diff
@@ -1,15 +1,13 @@
 name: http
-version: 0.13.0-nullsafety-dev
+version: 0.13.0-nullsafety
```
mit-mit, line 2: I thought we recommended starting with `nullsafety.0` ?

jakemac53: Either one is fine, good to have some consistency though and the `.0` version makes it a bit more obvious how the versioning will be handled so that seems a bit better.

That is not the pattern we did originally, but I am in favor of making it the new pattern.

kevmoo:
LGTE? (look good to everyone?)

jakemac53 (approved):

pubspec.yaml
kevmoo: fixed

PR #497, opened 2020-11-09
fix doc comments

kevmoo:


lib/testing.dart
```diff
@@ -22,4 +22,8 @@
 ///           200,
 ///           headers: {'content-type': 'application/json'});
 ///     });
+library http.testing;
```
jakemac53, line 25: Why do we need this?

lib/src/base_request.dart
```diff
@@ -6,6 +6,8 @@ import 'dart:collection';
 
 import 'package:meta/meta.dart';
 
+import '../http.dart';
```
natebosch, line 9: can we do something more narrow?

lib/testing.dart
kevmoo: Something to hang the comment off. Minimizes the chance dartdoc gets confused and thinks one is annotating the import directive

lib/src/base_request.dart
kevmoo: A bit. Sadly, `get` is in the root lib. Added a `show` so it's clear.

jakemac53 (approved):

PR #507, opened 2020-12-08
Tighten argument types to Uri

### LINKED ISSUE #375: Consider tightening arguments that are `dynamic`
The `dynamic url` which may be either a `String` or `Uri` is not an idiomatic pattern in Dart 2. You lose static checking and are forced to read the doc comment to get help with usage that is normally provided in your IDE. The migration would be easy, shift the `Uri.parse` call to the call site instead of inside the library code.

The downside here is that there is potentially a _lot_ of code to migrate since this is such a widely used package. That is mitigated somewhat by the fact that you'll get static checking for the migration in the vast majority of cases, only dynamic invocations would be runtime failures.

natebosch:
Fixes #375

Make all arguments that were `Object` in order to allow either `String`
or `Uri` accept only `Uri`. This gives better static checking for
calling code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ null safety allow list.
 
 * Add `const` constructor to `ByteStream`.
 * Migrate `BrowserClient` from `blob` to `arraybuffer`.
+* **BREAKING** All APIs which previously allowed a `String` or `Uri` to be
+  passed now require a `Uri`.
 
 ## 0.12.2
 
--- a/example/main.dart
+++ b/example/main.dart
@@ -4,7 +4,8 @@ import 'package:http/http.dart' as http;
 void main(List<String> arguments) async {
   // This example uses the Google Books API to search for books about http.
   // https://developers.google.com/books/docs/overview
-  var url = 'https://www.googleapis.com/books/v1/volumes?q={http}';
+  var url =
+      Uri.https('www.googleapis.com', '/books/v1/volumes', {'q': '{http}'});
 
   // Await the http get response, then decode the json-formatted response.
   var response = await http.get(url);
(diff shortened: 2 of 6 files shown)


jakemac53 (approved):

natebosch:
@bwilkerson - this seems like a good candidate for `dart fix`.

Can you point me to any docs on how to write the migration for this? What we'd like for a series of APIs, anytime a `String` is passed we want to wrap it with `Uri.parse()`. Would this be supported?

bwilkerson:
> Can you point me to any docs on how to write the migration for this?

The docs haven't been published yet, but I'll send you a pointer to where they live. Feedback on what's missing from or confusing in the docs would be appreciated.

> What we'd like for a series of APIs, anytime a String is passed we want to wrap it with `Uri.parse()`. Would this be supported?

I'd need to try it to be sure, but I think the answer is "yes", though maybe not as cleanly as might be hoped. It supports both adding and removing parameters, so I think you could tell it that the old parameter was removed and that a new parameter with the same name was added and that it's argument value depends on the value of the old argument. It might get confused by having them be the same name, though, hence the need to test it.

natebosch:
I took a brief look at the `dart fix` stuff and from what I can tell it does not let us express this change. I'm going to hold off on trying to write the `fix_data.yaml` for this. I think we could probably publish it as a fix later when that becomes possible.

@bwilkerson 

We don't have a `changeParameterType` transform. My first hunch was to model is as ` removeParameter` to drop the one that is `dynamic`, and then `addParameter` to add the one that is `Uri`. I can't figure out with that approach how to grab the previous argument expression as a template variable to use in a code template. Is that possible?

Would it be reasonable to add a `changeParamterType` transform explicitly? Should I file this as an issue on the SDK?

jakemac53:
This would be a high impact fix given that this is one of the most popular packages on pub, so if we can push on adding the ability to provide a fix we should.

PR #508, opened 2020-12-15
Make SocketException catchable

DrBu7cher:
You can check if that fixes everything, but for me it worked as expected. Before I had the annoying issue that my app would crash completely without any try-catch etc. helping in any way if my phone was simply offline. Now I can catch the SocketException with the try-catch around my http.get.

```dart
Future<User> getMe() async {
    try {
      final http.Response response = await http.get(
        baseUrl + "user/",
        headers: await defaultHeaders,
      );
      print("getMe: ${json.decode(utf8.decode(response.bodyBytes))}");
      if (response.statusCode == 200) {
        return User.fromJson(json.decode(utf8.decode(response.bodyBytes)));
      } else {
        String errorMsg =
            Error.fromJson(json.decode(utf8.decode(response.bodyBytes)))
                .message;
        return Future.error(errorMsg);
      }
    } catch (e) {
      print("getMe: $e");
      // THIS IS WHERE THE SOCKETEXCEPTION IS CATCHED
      return Future.error(_prepareError(e));
    }
  }
```

### DIFF
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -166,6 +166,8 @@ Future<T> _withClient<T>(Future<T> Function(Client) fn) async {
   var client = Client();
   try {
     return await fn(client);
+  } catch (e) {
+    rethrow;
   } finally {
     client.close();
   }
--- a/lib/src/base_client.dart
+++ b/lib/src/base_client.dart
@@ -89,8 +89,11 @@ abstract class BaseClient implements Client {
         throw ArgumentError('Invalid request body "$body".');
       }
     }
-
-    return Response.fromStream(await send(request));
+    try {
+      return Response.fromStream(await send(request));
+    } catch (e) {
+      rethrow;
+    }
   }
 
   /// Throws an error if [response] is not successful.
--- a/lib/src/io_client.dart
+++ b/lib/src/io_client.dart
@@ -57,6 +57,8 @@ class IOClient extends BaseClient {
           persistentConnection: response.persistentConnection,
           reasonPhrase: response.reasonPhrase,
           inner: response);
+    } on SocketException catch (error) {
+      throw ClientException(error.message, error);
     } on HttpException catch (error) {
       throw ClientException(error.message, error.uri);
     }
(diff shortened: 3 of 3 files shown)


GefaketHD:
@googlebot I singed it!

natebosch:
I would not expect any of these changes to make a meaningful difference to the exceptions you are able to catch.

Can you give a specific narrow reproduction case where this changes behavior?

GefaketHD:
Sure, I think I already did. But it should be an issue with every http-request sent. If I put my phone offline via firewall (not by disabling WiFi) and use the non-forked version of http, the whole app crashes with a SocketException that was uncatched. I tried a lot in my own code until I looked into http and changed the things I changed for that PR. :D The SocketException was thrown deeper in http and not correctly pushed up to where I could catch it with the provided try-catch I put in the PR as an example.

natebosch:
> The SocketException was thrown deeper in http and not correctly pushed up to where I could catch it

I'm not able to see that happen. Here is what I tried:

```dart

void main() async {
  try {
    await http.get(Uri.http('www.google.com', ''));
  } catch (e) {
    print('Caught: $e');
  }
  print('Done');
}
```

I disabled wifi and ran this application. The output was:

```
Caught: SocketException: Failed host lookup: 'www.google.com' (OS Error: nodename nor servname provided, or not known, errno = 8)
Done
```

I am able to catch the exception without applying the diff in this PR.

I don't know how to reproduce the behavior you are seeing. And I would not expect that the changes in this diff would have any impact on the behavior of this package.

gustav3d:
Hows it going ?

kenthinson:
@natebosch I am having the same issue. http on MacOS desktop. You can reproduce by trying to do a http.post while wifi is off on the laptop.

0xNF:
I am experiencing the issue this pull request aims to solve.

This is related to issue #551. In my case I am connecting directly to an ip address over https on the local network. My wifi is on, but the ip in question doesn't exist and thus isn't routable. The socket exception gets reported but never managed to hit the `catch` block.

natebosch:
@0xNF - do you have a minimal reproduction you can share? I have not been able to see locally a `SocketException` which escapes a `catch`

piotrsed (approved):

natebosch:
See [484274785](https://critique.corp.google.com/#review/484274785) which includes a code sample putting a `.catchError` _before_ the `.timeout` to catch the SocketException.

I'll close this PR for now since it does not make a meaningful difference about what exceptions can be caught.

Issue #511, opened 2020-12-23
Header injection and path forgery [security issue]


n0npax:
I believe there is a security issue with the current implementation of `Request`.

The full example can be found [here](https://github.com/n0npax/dart-test-http)

## What's wrong:
`Request` is passing method verb as is to the stream without any kind of validation.
basically:
```dart
  var r = Request(
      "GET http://example.com/ HTTP/1.1\r\nHost: example.com\r\nLLAMA:",
      Uri(scheme: "http", path: "/llama", host: "localhost"));
  var rs = await r.send();
```

generates request like:
```bash
sudo nc -l 127.0.0.1 80 
GET HTTP://EXAMPLE.COM/ HTTP/1.1
HOST: EXAMPLE.COM
LLAMA: /llama HTTP/1.1
user-agent: Dart/2.10 (dart:io)
accept-encoding: gzip
content-length: 0
host: localhost
```

## What I'd expect
If HTTP verb(method) is not a part of the known set of verbs (GET/PUT/HEAD/...) I'd expect an exception to be thrown

## Why this is a security risk
If the developer is using `Request` to abstract generating HTTP calls and he's accepting a `method` param from the user, the user can do some magic like header injection or path forgery.
This can be exploited in many ways and seems to be quite important *especially in case there is a reverse proxy is in place*. A proxy may just pass someone's request to any host.
By running  [snippet](https://github.com/n0npax/dart-test-http)  behind a proxy, I was nicely redirected([like this](https://github.com/n0npax/dart-test-http#critical-path)) to `example.com` which was injected as per the example above.
Let's assume I'm replacing `example.com` with  `my-hackery-uservice.org` and the victim is working in a company behind the proxy. This means I can redirect calls with headers/cookies(tokens) and blah blah blah.

n0npax:
`StreamedRequest` is also affected.

PR #512, opened 2020-12-23
Gh 511 http header injection fix

n0npax:
PoC of solution for issue https://github.com/dart-lang/http/issues/511


natebosch:
cc @lrhn 

Should we consider doing this check in an `assert`?

I don't think I'd frame this as a security issue - it's more of a potential footgun for authors. I don't know how often I'd expect an author to make a mistake with this API, but I'm on board with adding some rails if we think that it's likely. An `assert` adds some help for authors without any overhead in production.

n0npax:
Please have a look at https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-26137 & https://bugs.python.org/issue39603
which is a very similar issue. Base on this I believe it's not just a foot gun, as the same risk exists here.

I'm happy to change `contains` to `assert`. I'll update the PR soon

OrenSavichWS:
https://github.com/dart-lang/http/pull/512/commits/17bc4bf944104b66c1691db5c171664d679ab7a5
Hi, is this commit the fix for CVE-2020-35669? 

if not, what is?

n0npax:
https://github.com/dart-lang/http/pull/512/commits/818cff614641715015603639f0125b91881456ad is a fix. The remaining commits are just unit tests and some lint fixes.

To fix stuff properly there may be a need to bump the version in https://github.com/dart-lang/http/blob/master/pubspec.yaml#L2 & release, but it's up to maintainers to decide if they prefer `0.13.1` or backport to `0.12.3` 🤷🏻‍♂️

n0npax:
@lrhn can I ask for a review please?

n0npax:
@kevmoo can I ask for a review please ?

kevmoo:
@lrhn ?

lib/src/base_request.dart
```diff
@@ -88,8 +88,18 @@ abstract class BaseRequest {
...
-      : headers = LinkedHashMap(
+  BaseRequest(String method, this.url)
+      : method = [
+          'HEAD',
+          'GET',
+          'POST',
+          'PUT',
+          'PATCH',
+          'DELETE',
+        ].contains(method)
```
lrhn, line 99: Where does this list come from?

The [HTTP RFC](https://tools.ietf.org/html/rfc7231#section-4.3) specifies:
 * GET 
 * HEAD
 * POST
 * PUT
 * DELETE
 * CONNECT
 * OPTIONS
 * TRACE

as defined standard methods, and we are missing a couple of these. The IANA [method registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) is much larger.

The only real requirement is that it's a ["token"](https://tools.ietf.org/html/rfc7230#section-3.2.6), to allowing servers to support custom methods. so maybe just check that.
Something like:
```dart
static final _tokenRE = RegExp(r"^[\w!#%&'*+\-.^`|~]+$");
static String _validateMethod(String method) {
  if (!_tokenRE.hasMatch(method)) {
    throw ArgumentError.value(method, "method", "Not a valid method");
  }
  return method;
}
...
BaseRequest(String method, this.url) : method = _validateMethod(method);
```

This should still ensure that the method is parsed as a single token by the receiver, but not put any unnecessary restrictions on it.
(I'm using a `RegExp` here, if performance is an issue, it can probably be optimized).

lib/src/base_request.dart
```diff
@@ -88,8 +88,18 @@ abstract class BaseRequest {
...
+            ? method
+            : throw StateError('invalid HTTP method.'),
```
lrhn, line 101: Should be an argument error.

test/request_test.dart
```diff
@@ -334,4 +334,10 @@ void main() {
       expect(request.toString(), 'POST $dummyUrl');
     });
   });
+
+  group('#invalid method', () {
+    test('Invalid method', () {
+      expect(() => http.Request('LLAMA', dummyUrl), throwsStateError);
```
lrhn, line 340: An actually invalid method would then contain, e.g., a space or a disallowed character like `[` or `]`. Try `"LLAMA[0]"` or `"A B"`.

test/request_test.dart
```diff
@@ -334,4 +334,10 @@ void main() {
       expect(request.toString(), 'POST $dummyUrl');
     });
   });
+
+  group('#invalid method', () {
```
lrhn, line 338: The name of the group is the getter/method being tested, so just use `"#method"`.
Then name the test to describe the error, maybe `"must be a token"`.

test/streamed_request_test.dart
```diff
@@ -24,4 +24,9 @@ void main() {
       expect(() => request.contentLength = 10, throwsStateError);
     });
   });
+  group('#invalid method', () {
```
lrhn, line 27: Same naming as in other test: `"#method"` and `"must be token"`.

lrhn:
I'd test this properly, not just in an assert. Any overhead is negligible compared to doing a network request.

I agree that it's probably not a *big* security issue since you can equally easily do a direct TCP connection and send *any* request block. If malicious code is running in the same isolate as something else, you should assume that something else to be thoroughly compromised anyway, so unless code is taking the `method` value as input from a remote source, and not validating it at all (which is already dangerous when one of the values are `DELETE`), then it's very hard to do something bad with this mis-feature which isn't also possible without it.
It's still an unnecessary risk, someone could be living dangerously and accepting `method` values from strangers, so we might as well plug it.

test/streamed_request_test.dart
n0npax: fixed. Thanks

test/request_test.dart
n0npax: fixed. Thanks

test/request_test.dart
n0npax: fixed. Thanks

lib/src/base_request.dart
n0npax: fixed. Thanks

lib/src/base_request.dart
n0npax: If I understand correctly, all methods listed in the IANA registry can be matched just by `^[\w\-]+$`.

I've committed: `"^[\w!#%&'*+\-.^|~]+$"`, but can I ask why characters like `#%&...` are expected to be considered as valid?

lrhn:
I don't know if this package uses the `dart:io` HTTP operations. If it does, we might be able to remove this again when they're fixed. I believe the browser `XMLHttpRequest` is already validating the method.

lrhn (approved):

kevmoo:
Should also bump the version and and an entry to the changelog!

n0npax:
@kevmoo done, hope it's ok 🤞🏻

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 0.13.2-dev
+## 0.13.3-dev
+
+## 0.13.2
+
+* Fix `CRLF` in `BaseRequest`
```
lrhn, line 5: This could use some context or explanation (`CRLF` is not defined).
Maybe:

 * Validate the `method` parameter of `BaseRequest`. Must now be a valid "token".

This is correct. It doesn't say why, but validation is generally a good thing.

lrhn (approved):

CHANGELOG.md
n0npax: fixed as per suggestion, Thank you for all hints!

natebosch:
@googlebot I consent.

natebosch (approved):

n0npax:
@googlebot I consent.

PR #521, opened 2021-01-21
Add support for request timeout

natebosch:
Wire up a timer to perform the appropriate type of aborting and resource
cleanup for each client. Handles canceling the timer when the request
completes.

### DIFF
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// A composable, [Future]-based library for making HTTP requests.
+import 'dart:async';
 import 'dart:convert';
 import 'dart:typed_data';
 
@@ -27,6 +28,10 @@ export 'src/streamed_response.dart';
 
 /// Sends an HTTP HEAD request with the given headers to the given URL.
 ///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
+///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
 /// the same server, you should use a single [Client] for all of those requests.
@@ -38,6 +43,10 @@ Future<Response> head(Uri url,
         (client) => client.head(url, headers: headers, timeout: timeout));
 
 /// Sends an HTTP GET request with the given headers to the given URL.
+///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
 ///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
@@ -65,6 +74,10 @@ Future<Response> get(Uri url,
 ///
 /// [encoding] defaults to [utf8].
 ///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
+///
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
 Future<Response> post(Uri url,
@@ -90,6 +103,10 @@ Future<Response> post(Uri url,
 /// `"application/x-www-form-urlencoded"`; this cannot be overridden.
 ///
 /// [encoding] defaults to [utf8].
+///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
 ///
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
@@ -118,6 +135,10 @@ Future<Response> put(Uri url,
 ///
 /// [encoding] defaults to [utf8].
 ///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
+///
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
 Future<Response> patch(Uri url,
@@ -129,6 +150,10 @@ Future<Response> patch(Uri url,
         headers: headers, body: body, encoding: encoding, timeout: timeout));
 
 /// Sends an HTTP DELETE request with the given headers to the given URL.
+///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
 ///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
@@ -149,6 +174,10 @@ Future<Response> delete(Uri url,
 /// The Future will emit a [ClientException] if the response doesn't have a
 /// success status code.
 ///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
+///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
 /// the same server, you should use a single [Client] for all of those requests.
@@ -166,6 +195,10 @@ Future<String> read(Uri url,
 ///
 /// The Future will emit a [ClientException] if the response doesn't have a
 /// success status code.
+///
+/// If [timeout] is not null the request will be aborted if it takes longer than
+/// the given duration to complete, and the returned future will complete as an
+/// error with a [TimeoutException].
 ///
 /// This automatically initializes a new [Client] and closes that client once
 /// the request is complete. If you're planning on making multiple requests to
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -32,8 +32,10 @@ export 'src/streamed_response.dart';
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> head(Uri url, {Map<String, String>? headers}) =>
-    _withClient((client) => client.head(url, headers: headers));
+Future<Response> head(Uri url,
+        {Map<String, String>? headers, Duration? timeout}) =>
+    _withClient(
+        (client) => client.head(url, headers: headers, timeout: timeout));
 
 /// Sends an HTTP GET request with the given headers to the given URL.
 ///
@@ -42,8 +44,10 @@ Future<Response> head(Uri url, {Map<String, String>? headers}) =>
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> get(Uri url, {Map<String, String>? headers}) =>
-    _withClient((client) => client.get(url, headers: headers));
+Future<Response> get(Uri url,
+        {Map<String, String>? headers, Duration? timeout}) =>
+    _withClient(
+        (client) => client.get(url, headers: headers, timeout: timeout));
 
 /// Sends an HTTP POST request with the given headers and body to the given URL.
 ///
@@ -64,9 +68,12 @@ Future<Response> get(Uri url, {Map<String, String>? headers}) =>
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
 Future<Response> post(Uri url,
-        {Map<String, String>? headers, Object? body, Encoding? encoding}) =>
-    _withClient((client) =>
-        client.post(url, headers: headers, body: body, encoding: encoding));
+        {Map<String, String>? headers,
+        Object? body,
+        Encoding? encoding,
+        Duration? timeout}) =>
+    _withClient((client) => client.post(url,
+        headers: headers, body: body, encoding: encoding, timeout: timeout));
 
 /// Sends an HTTP PUT request with the given headers and body to the given URL.
 ///
@@ -87,9 +94,12 @@ Future<Response> post(Uri url,
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
 Future<Response> put(Uri url,
-        {Map<String, String>? headers, Object? body, Encoding? encoding}) =>
-    _withClient((client) =>
-        client.put(url, headers: headers, body: body, encoding: encoding));
+        {Map<String, String>? headers,
+        Object? body,
+        Encoding? encoding,
+        Duration? timeout}) =>
+    _withClient((client) => client.put(url,
+        headers: headers, body: body, encoding: encoding, timeout: timeout));
 
 /// Sends an HTTP PATCH request with the given headers and body to the given
 /// URL.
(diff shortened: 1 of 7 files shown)


natebosch:
cc @lrhn @jakemac53 

I'd like to get some initial feedback on the approach and whether this is worth doing.

I think this is worth doing because timeouts are a common case for wanting to abort requests, and they are easier to get wrong than to get right. The naive approach of `.timeout` doesn't clean up resources and the entire result is still going to get read. Even with a general mechanism to abort requests a timeout argument is handy because we can handle the timer cancellation which would otherwise be boilerplate. We also likely can't arrive at a solid design for request aborting in general, but the API design with `timeout:` argument is fairly straightforward.

Please let me know what you think of this initial version, if we think it looks like a good approach I can start work on tests. I'm not sure how easy it is going to be to cover all the behavior here well.

lib/src/browser_client.dart
```diff
@@ -50,34 +67,36 @@ class BrowserClient extends BaseClient {
       ..responseType = 'arraybuffer'
       ..withCredentials = withCredentials;
     request.headers.forEach(xhr.setRequestHeader);
-
-    var completer = Completer<StreamedResponse>();
+    if (timeout != null) {
+      onTimeout = () {
+        xhr.abort();
```
jakemac53, line 72: Should we be removing the xhr from _xhrs here? It seems to me that currently that would only happen once the request completes with either an error or success. Presumably calling abort does make it complete with an error, but then i would only expect to see the call to `abort` here, otherwise we are likely trying to complete the completer twice?

jakemac53: Maybe we should remove the xhr whenever the completer completes actually instread of doing it manually in all these places?

jakemac53 (commented):
The api lgtm

lib/http.dart
```diff
@@ -32,8 +32,10 @@ export 'src/streamed_response.dart';
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> head(Uri url, {Map<String, String>? headers}) =>
-    _withClient((client) => client.head(url, headers: headers));
+Future<Response> head(Uri url,
+        {Map<String, String>? headers, Duration? timeout}) =>
```
lrhn, line 36: Document the parameter.

What happens when the timeout duration has passed? I'm *expecting* the returned future to complete with a `TimeoutException`. If so, that's a reasonable default, and there isn't really a good alternative.

lib/src/base_client.dart
```diff
@@ -68,12 +88,12 @@ abstract class BaseClient implements Client {
...
   /// internal HTTP errors should be wrapped as [ClientException]s.
   @override
-  Future<StreamedResponse> send(BaseRequest request);
+  Future<StreamedResponse> send(BaseRequest request, {Duration? timeout});
 
   /// Sends a non-streaming [Request] and returns a non-streaming [Response].
   Future<Response> _sendUnstreamed(
       String method, Uri url, Map<String, String>? headers,
-      [body, Encoding? encoding]) async {
+      {dynamic body, Encoding? encoding, Duration? timeout}) async {
```
lrhn, line 96: Personally, I usually make all parameters required for internal functions.
It's not great API, but it's incredibly simple and efficient.

lib/src/browser_client.dart
lrhn: Seems like we remove it on `completer.future.whenComplete` above, so it should be fine.

lib/src/browser_client.dart
```diff
@@ -41,20 +41,45 @@ class BrowserClient extends BaseClient {
...
+        completer.completeError(TimeoutException('Request aborted', timeout));
+      };
+    }
 
     unawaited(xhr.onLoad.first.then((_) {
       var body = (xhr.response as ByteBuffer).asUint8List();
+      timer?.cancel();
       completer.complete(StreamedResponse(
```
lrhn, line 83: Should probably guard by an
```dart
  if (!completer.isCompleted) { 
    completer.complate(...)
  }
```
It's probably technically possible for the timer to have completed the completer, but the callback not having been called yet (it's not a sync completer!), and then an unguarded `completer.complete` here will throw a `StateError`. Basically a race condition.
Ditto below.


Or maybe it can't happen because the timeout is a timer event, so a top-level event loop event of its own, and it completing a future will happen in a microtask of the *same* event, the only microtask scheduled by that event, so nothing can get between them. Then that might be worth documenting.

lib/src/io_client.dart
```diff
@@ -23,42 +24,82 @@ class IOClient extends BaseClient {
...
+      }, test: (error) => error is HttpException).transform<List<int>>(
+          StreamTransformer.fromHandlers(handleDone: (sink) {
+        timer?.cancel();
+        sink.close();
+      }));
+
+      if (timeout != null) {
+        onTimeout = () {
+          // TODO, is this necessary? How will it surface?
+          response.detachSocket().then((socket) => socket.destroy());
```
lrhn, line 84: Should you throw a `TimeoutException` here?

lib/src/io_client.dart
```diff
@@ -23,42 +24,82 @@ class IOClient extends BaseClient {
...
+    _send(request, timeout, completer);
+    return completer.future;
+  }
+
+  Future<void> _send(BaseRequest request, Duration? timeout,
+      Completer<IOStreamedResponse> completer) async {
     var stream = request.finalize();
 
+    Timer? timer;
+    late void Function() onTimeout;
```
lrhn, line 38: Can make this nullable instead of late, and do `onTimeout?.call();` or `onTimeout!()` in the timer callback.
That could potentially be cheaper than a late field. (Unless the compiler is being competent about late non-nullable fields and using `null` as placeholder anyway)

lib/src/browser_client.dart
```diff
@@ -41,20 +41,45 @@ class BrowserClient extends BaseClient {
...
+  Future<StreamedResponse> send(BaseRequest request, {Duration? timeout}) {
+    final completer = Completer<StreamedResponse>();
+    _send(request, timeout, completer);
+    return completer.future;
+  }
+
+  Future<void> _send(BaseRequest request, Duration? timeout,
+      Completer<StreamedResponse> completer) async {
+    Timer? timer;
+    late void Function() onTimeout;
```
lrhn, line 53: Maybe avoid the `late` and do:
```dart
  HttpRequest? xhrOptional;
  if (timeout != null) {
    timer = Timer(timeout, () {
      xhrOptional?.abort();
      completer.completeError(TimeoutException("Request aborted", timeout));
    });
  }
  var bytes = await request.finalize().toBytes();
  var xhr = xhrOptional = HttpRequest();
```

lrhn (approved):

lib/http.dart
natebosch: Yes I think completing with a `TimeoutException` is the most useful API.

If it turns out that it is more common to want to ignore the timeout and have a `Future` that doesn't complete instead we could drop a utility for it in `package:async`

```dart
extension IgnoreTimeout<T> on Future<T> {
  Future<T> ignoreTimeout() => catchError((_) {}, test:(e) => e is TimeoutException);
}
```

lrhn: I don't think anyone ever wants a future which doesn't complete.
(They might think they do, but then they end up doing `try { await something; } finally { never runs!!!?! }` and change their mind).

lib/src/io_client.dart
natebosch: Yes, looks like I need to handle this on the response stream.

lib/src/io_client.dart
natebosch: I think I have a slight preference for `late` as it matches the intended semantics more closely IMO. Outside of performance concerns do you have a preference one way or the other?

lib/src/browser_client.dart
natebosch: Done.

lib/src/base_client.dart
natebosch: Sure, this library is small enough that it isn't much of a readability hit.

natebosch:
One more question about the API design.

Some backround: There are differences between behaviors with the `send` API. In the browser the `Future` won't complete until the entire response body is read, while on the vm the `Future` will complete as soon as headers are received, and the response body `stream` is kept open until the entire body is read. I do think it's plausible we could change this in the browser - perhaps not going all the way to a real streaming implementation, but potentially at least making the `Future` complete with the headers. I don't know if we need to block on that.

The immediate question is, should the Stream emit the Timeout or not. If it doesn't the timeout would apply only to receiving the headers and you'd need to cancel a stream subscription to get a full timeout.
My current implementation emits the timeout, which is convenient because it makes the `Future` only APIs work as expected without any extra effort.

lib/src/io_client.dart
lrhn: I consider `late` an unsafe and dynamic feature at the same level of as, well, `dynamic`. I try to avoid it as much as possible, just as I do `dynamic`. Late variables have their uses, where it makes things easier, but it never makes them *safer*. 

Using `late` (and `dynamic`) allows statically undetectable run-time errors with visible marker at the use position. I prefer to use something nullable and `!` when possible *because* it is more verbose and therefore highlights the assumption at the read (by adding `!`).

This is a fairly fundamental library, and I think we should always choose safety (and performance) over brevity.

lrhn:
It sounds like we have two APIs, one future based and one stream based.
They should probably act differently.

It's hard to say what a "timeout" means if the author cannot specify it. Do they need the entire result before that time (when do I need to move on, with or without a response?), or do they just need some evidence that the remote end exists and is willing to answer before that time (how long do I wait if the remote server is down?). I can see both use-cases as valid.

Is there any way to *cancel* the timeout? So, set a timeout of 5 seconds, which will then kill the connection after that time (and put an exception on the stream), but if I get a header and plausible start of the response before that time, then I can cancel the timeout (or start a new timer). That would make it easier for the client to control the timeout behavior without having to know up-front whether they're waiting for a non-remote-end timeout or a large response.

lib/src/io_client.dart
jakemac53: Using `late` here I believe is more about choosing something that matches the underlying semantics/implementation than brevity. I don't feel super strongly either way, but this is always assigned, and I don't agree it is any safer to use `!` everywhere - we would not be adding any null checks either way.

jakemac53: Hmm actually on re-reading it I think this is only assigned if `timeout` is non-null, in which case making it nullable does match the semantics better, so I retract my previous comment and agree it should be nullable.

natebosch: I wasn't taking enough advantage of flow analysis. Turns out we don't need `late` or `?`. If I switch a few lines around then definite assignment analysis takes over.

natebosch:
I can see the use cases for both variants of timeout. One thing I'm not sure of is whether most authors would want to timeout the connection (which they can do [with a `dart:io` client](https://api.dart.dev/stable/2.10.5/dart-io/HttpClient/connectionTimeout.html)) or the headers.

natebosch:
> I can see both use-cases as valid.

I agree that both are probably valid, it's hard to say whether both are worth the API space. After some discussion I think we are leaning towards the timeout applying to the Stream as well. It's slightly less general, but also a friendlier API for the case it solves, and my hunch is that is the more common one.

What we can do, though, is hold on to a little room for the other one. If we name the argument `contentTimeout` instead of `timeout` then it's easier to add a `headerTimeout` argument later. It would still be a breaking change and a major version bump, but the vast majority of uses would not be impacted.

Currently I'm leaning towards using the name `timeout` for the `Future` APIs, and `contentTimeout` for the `Future<Stream>` API. We _could_ add a `headerTimeout` to the `Future` APIs but it starts to get weird to have a `Future` which might be timed out within some period, or might wait indefinitely (or to some longer separate timeout) without any way to observe what is happening. I don't know if anyone is asking for a timeout specifically for headers either - I think most users expect a `Future` to have a single timeout, and the name `timeout` is nicest for that. I'd be happy to discuss using the name `contentTimeout` for all of these arguments consistently if anyone wants to bikeshed a bit more or thinks it is important to leave room to add a `headerTimeout` for these other APIs as well.

> Is there any way to cancel the timeout?

Not that I can think of without a difficult to understand API.

lib/http.dart
natebosch: Added docs.

lib/src/browser_client.dart
natebosch: I went ahead and guarded them all with a `completer.isCompleted` checks in both client implementations. I think it's probably more checks than are needed, but adding the checks won't introduce bugs and we can always remove them later with an explanation of why it isn't required.

natebosch:
After some further discussion we think there is more risk in getting this API wrong than there is in delaying it until after the dust settles on NNBD and we can take more time to design it alongside general request aborting and decide about the rollout plan from there. We may end up with a slower alternate-import type migration for request aborting, and so we could use that same approach to deliver a timeout feature alongside abort with a gradual migration.

We should continue to use this discussion to inform our future design, but I don't think I'll hold the null safe publish on it.

natebosch:
One thing I'm noticing about this as I look at overrides of `send` in other client implementations - it isn't very friendly to deal with the `Timer`. From an overrider standpoint the `Future<void> abortOn` seems easier. It's unfortunate that the design for customizing a client deals with overriding a public facing API - if those were separate signatures we could give a friendly-to-use API for users and a friendly-to-implement API for overriders.

lib/src/base_request.dart
```diff
@@ -117,11 +118,18 @@ abstract class BaseRequest {
   /// the request is complete. If you're planning on making multiple requests to
   /// the same server, you should use a single [Client] for all of those
   /// requests.
-  Future<StreamedResponse> send() async {
+  ///
+  /// If [contentTimeout] is not null the request will be aborted if it takes
+  /// longer than the given duration to receive the entire response. If the
+  /// timeout occurs before any reply is received from the server the returned
+  /// future will as an error with a [TimeoutException]. If the timout occurs
```
lrhn, line 125: Missing a "will *complete* as" here?

I'd usually write "the returned future will complete with a [TimeoutException] error."

lib/src/base_request.dart
```diff
@@ -117,11 +118,18 @@ abstract class BaseRequest {
...
+  /// after the reply has been started but before the entire body has been read
```
lrhn, line 126: Comma before "but"? (And then probably also after "read").

Maybe "has been started" -> "has started".

lib/src/base_request.dart
```diff
@@ -117,11 +118,18 @@ abstract class BaseRequest {
   /// the request is complete. If you're planning on making multiple requests to
   /// the same server, you should use a single [Client] for all of those
   /// requests.
-  Future<StreamedResponse> send() async {
+  ///
+  /// If [contentTimeout] is not null the request will be aborted if it takes
```
lrhn, line 122: I'd usually say "If [contentTimeout] is provided, the ...."
I know you can technically provide a `null` argument, but users will understand it anyway. After all, you talk about the duration afterwards, and `null` is not a duration.

lib/src/base_request.dart
```diff
@@ -117,11 +118,18 @@ abstract class BaseRequest {
...
+  /// longer than the given duration to receive the entire response. If the
+  /// timeout occurs before any reply is received from the server the returned
```
lrhn, line 124: Comma after "server"?

lrhn (approved):
API LGTM.

lib/src/base_request.dart
DeD1rk: Don't think so, the '... but ...' is a single description of a condition, whereas a comma would suggest the two parts being separate sentences.

lrhn: I read "but before the entire body has been read" as parenthetical.
If it's not a parenthetical clause, it deserves to be equally prominent to the main clause, and I'd change `but` to `and`.
I'd still have a comma after `read` in either case.

amondnet (approved):

piotrsed (approved):

lukacat10:
This PR contains important features, has gone under extensive reviews, and should be merged by now.  
There are only a few improvements left on the documentation, and some merge conflicts.  
@natebosch Can you please address this? So an approval can be made?

RyanKhanh (approved):

PR #552, opened 2021-03-27
Added ```isSuccessful``` method to ```BaseResponse```

agnamc9:
Added ```isSuccessful``` method when ```statusCode``` is in the range [200, 300).


agnamc9:
@googlebot I signed it!

Issue #550, opened 2021-03-22
[Error] the Content-Length header occurred more than once, at most one is allowed.


EgHubs:
when i try to connect over the following link:
` http://192.168.4.1/wifisave?s=asdasd&p=10001010101010&ip=192.168.1.202&gw=192.168.1.1&sn=255.255.255.0&email=new@test.com&pass=123456789&alarm=1`
I get the following error

> The Content-Length header occurred more than once, at most one is allowed.

even thought it was working before, I don't know if the upgrade of the latest version of the library did affect that..
the line of code:
```
try {
String overAllLink='http://192.168.4.1/wifisave?s=asdasd&p=10001010101010&ip=192.168.1.202&gw=192.168.1.1&sn=255.255.255.0&email=new@test.com&pass=123456789&alarm=1';
    response = await http.get(Uri.parse(overAllLink));
  } catch (e) {
    print(e);
  }
```

LucasAschenbach:
I'm getting the same error message.
@EgHubs Did you find a solution for this problem?

EgHubs:
Unfortunately no, i'am stuck in here

LucasAschenbach:
@EgHubs In your http request, is the content-length field in the response spelled out `'Content-Lenght'` or `'content-length'`?

(I have been testing a few different APIs, all of which worked fine. However, they all spelled content-length with lower-case letters whereas the API which causes the exception spells it with capital `C` and capital `L`.)

EgHubs:
If you mean how I get the error message I get it in the same way that I have referenced, with capital letters.

LucasAschenbach:
For instance, here is a response where the content-length field-name is written with lower case letters:

```
StatusCode: 200, ReasonPhrase: '', Version: 3, Content: {
  "args": {}, 
  "headers": {
    "Accept-Encoding": "gzip, deflate", 
    "Cache-Control": "no-cache", 
    "Content-Length": "0", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "Host": "httpbin.org", 
    "X-Amzn-Trace-Id": "Root=1-607574e7-7b8854500cab78fb48eedeae"
  }, 
  "origin": "88.69.152.24", 
  "url": "https://httpbin.org/get"
}
, Headers:
{
  server: gunicorn/19.9.0
  date: Tue, 13 Apr 2021 10:39:35 GMT
  access-control-allow-credentials: true
  access-control-allow-origin: *
}{
  content-length: 359  // <-- lower case
[...]

PR #555, opened 2021-04-13
Merge http_retry into this package

natebosch:
Update all references to `http_retry` to `http/retry`. Shorten the
README content a bit and put it a separate section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,33 @@ class UserAgentClient extends http.BaseClient {
   }
 }
 ```
+
+## Retrying requests
+
+`package:http/retry.dart` provides a class [`RetryClient`][RetryClient] to wrap
+an underlying [`http.Client`][Client] which transparently retries failing
+requests.
+
+[RetryClient]: https://pub.dev/documentation/http/latest/retry/RetryClient-class.html
+[Client]: https://pub.dev/documentation/http/latest/http/Client-class.html
+
+```dart
+import 'package:http/http.dart' as http;
+import 'package:http/retry.dart';
+
+Future<void> main() async {
+  final client = RetryClient(http.Client());
+  try {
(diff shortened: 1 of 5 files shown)


natebosch:
These commits have already been accepted into a dart-lang repo. The CLA shouldn't be an issue for this merge.

kevmoo (approved):

natebosch:
I'll need to temporarily enable merge commits to preserve history here.

Unfortunately I think we'll have a double merge, since the HEAD commit of this branch is already a merge. I think that's probably fine.

@kevmoo  do you have any concerns about that?

jakemac53 (approved):
Generally this LGTM, although it is a bit unfortunate to have the separate library, I assume that is to avoid import conflicts?

natebosch:
> I assume that is to avoid import conflicts?

Yes. If everyone is following the advice of using an import prefix for `http`, and is _not_ doing that for `http_retry` then in theory it should be safe, but I'm not confident that is the case.

I'm also not entirely sure it makes sense to have `RetryClient` in the general library for this package. It's a bit specialized IMO.

jakemac53:
> I'm also not entirely sure it makes sense to have `RetryClient` in the general library for this package. It's a bit specialized IMO.

Ya it just isn't much code, andits likely very tree-shakeable code as well. Plus we don't typically do separate libraries. I don't think it is a big deal either way though so LGTM.

EgHubs:
Sorry for referencing here, but since version 13.0, there been many errors that we all facing, please review them in the issue center
Mine here

https://github.com/dart-lang/http/issues/550

natebosch:
We synced offline and decided to go forward with separate libraries, and also to merge this manually instead of through a github PR.

PR #565, opened 2021-04-29
Remove expectation that GET without a body has content-length 0.

aam:
See https://dart-review.googlesource.com/c/sdk/+/194881 that is going to remove content-length:0 header for GET and HEAD requests that has no body.


natebosch (approved):

natebosch:
@google I consent.

Issue #465, opened 2020-08-30
Add support for getting the progress of a file upload / download


lirantzairi:
Hi,

When a user uploads a file in my app, I want to show him a progress bar. I'm looking for a way to listen to the amount of bytes sent in MultipartRequest, but I think it's not supported. Apparently you can do this with dart:io (https://stackoverflow.com/questions/50455131/flutter-how-to-get-upload-download-progress-for-http-requests?rq=1, https://medium.com/flutter-community/how-to-show-download-progress-in-a-flutter-app-8810e294acbd).

Can you add a support for this? Thanks.

syedtalha:
also, if the same could be supported on web as well.

tlueder:
Hi,

I've opened a PR to add a onSendProgress callback.

Any help in advancing it as much appreciated.

For example I have no idea how to write a meaningful test for this feature.

Thanks

xOldeVx:
What about it?

PR #579, opened 2021-05-27
Add onSendProgress callback

tlueder:
This fixes the upload part of https://github.com/dart-lang/http/issues/465 

I had also tried to include the download progress, but that did not work as expected.
And it is why easier to do this while consuming the response stream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.13.4-dev
 
+* Add `onSendProgress` callback
+
 ## 0.13.3
 
 * Validate that the `method` parameter of BaseRequest is a valid "token".
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -8,6 +8,7 @@ import 'dart:typed_data';
 
 import 'src/client.dart';
 import 'src/exception.dart';
+import 'src/progress.dart';
 import 'src/request.dart';
 import 'src/response.dart';
 import 'src/streamed_request.dart';
@@ -20,6 +21,7 @@ export 'src/client.dart';
 export 'src/exception.dart';
 export 'src/multipart_file.dart';
 export 'src/multipart_request.dart';
+export 'src/progress.dart';
 export 'src/request.dart';
 export 'src/response.dart';
 export 'src/streamed_request.dart';
@@ -61,12 +63,25 @@ Future<Response> get(Uri url, {Map<String, String>? headers}) =>
 ///
 /// [encoding] defaults to [utf8].
 ///
+/// If [onSendProgress] is provided it will be called to indicate
+/// the upload progress
+///
 /// For more fine-grained control over the request, use [Request] or
 /// [StreamedRequest] instead.
-Future<Response> post(Uri url,
-        {Map<String, String>? headers, Object? body, Encoding? encoding}) =>
-    _withClient((client) =>
-        client.post(url, headers: headers, body: body, encoding: encoding));
+Future<Response> post(
+  Uri url, {
+  Map<String, String>? headers,
+  Object? body,
+  Encoding? encoding,
+  Progress? onSendProgress,
+}) =>
+    _withClient((client) => client.post(
+          url,
+          headers: headers,
+          body: body,
+          encoding: encoding,
+          onSendProgress: onSendProgress,
+        ));
 
 /// Sends an HTTP PUT request with the given headers and body to the given URL.
 ///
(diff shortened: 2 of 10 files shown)


tlueder:
@googlebot I signed it!

jasonroland (approved):

NazarenoCavazzon:
Hope this gets merged, it would really help

tomassasovsky:
@tlueder @jasonroland @piotrsed any updates on this?

tlueder:
Hi, I can rebase this. But without a clear path forward I don't see the point.

Currently I just use this plugin from my repro via dependency override, works great.

pkgs/http/lib/src/io_client.dart
```diff
@@ -63,7 +67,24 @@ class IOClient extends BaseClient {
...
+        await ioRequest.addStream(
+          stream.map(
+            (chunk) {
+              loaded += chunk.length;
+              onSendProgress(loaded, request.contentLength);
+              return chunk;
+            },
+          ),
+        );
+        response = await ioRequest.close();
```
yinxulai, line 84: Why to needs close this ioRequest

tlueder: It's been a while since I've last had a look at this.

But I need the response from the ioRequest for the code that follows and I guess that was the easiest way I could think of.

But if that is not a good idea or best practice I'm all in for improvements.

kevmoo:
@brianquinlan @devoncarew @natebosch

kevmoo:
Please fix merge conflicts!

tlueder:
@kevmoo done

natebosch:
This looks like a breaking change, and we have been avoiding breaking API changes. This is only breaking for a subset of users, but I don't know what proportion.

We are currently getting through a major version bump without too much difficulty, but that was a version bump without breaking API changes. cc @brianquinlan for thoughts on the practicality of a breaking API change and another version bump.

tlueder:
This is a API change for sure but I would argue against a breaking one. This adds a new optional feature. And should not break anything for anyone not using it.

Can you elaborate what you mean by "This is only breaking for a subset of users"?
Maybe I can change the code to make sure it won't affect them.

tlueder:
Of course, I haven't even thought about that.
Thanks for the information.

brianquinlan:
> We are currently getting through a major version bump without too much difficulty, but that was a version bump without breaking API changes. cc @brianquinlan for thoughts on the practicality of a breaking API change and another version bump.

1. As you said, this is definitely a breaking change (in particular it will definitely break at least `package:cronet_http` and `package:cupertino_http`).
2. I wonder if there is a way to implement this without breaking the API? For example, we could define a new class like:

```dart
interface class ProgressTracker {
  void onSendProgress(int loaded, int total);
}

class ProgressTrackingRequest extends StreamedRequest implements ProgressTracker {
   ...
}
```

And `send` implementations could check if the `Request` implements that class and, if so, could update the progress e.g.

```dart
if (request is ProgressTracker) {
  request.onSendProcess(...);
}
```

That's a bit ugly.

3. I'm not sure why tracking upload progress is more important than tracking download progress. If we are going to make a breaking change then we should do both so we don't have to update the major version just to add the obvious complementary feature.
4. I'd be fine with another version bump but maybe we should wait a few months and accumulate all the breaking changes into one release? Is there an easy way to do this? Like have a 2.0 branch or something?

tlueder:
1. Understood, and thanks for your input I was not aware of that.
2. I'd rather wait for a version bump instead of a "quick fix". But in the end, I don't really care so much about the how this is implemented, because I don't know enough about dart best practice and what side effects this can have. I just need the progress data and if this is done a different way? I'm happy to update my apps if that means I can get rid of the dependency override and rebasing my fork.
3. It is not that tracking the download is not important, but in contrast to the upload it can be done easily on the consuming side as all the needed data is already available. Just not as callback function. And when I wrote this, 2 years ago, I was not able to do this and my attempts felt just wrong. So I left it at that, but any hints in the right direction are welcome.

brianquinlan:
Hey @kevmoo 

What is your expectation here in terms of review?

kevmoo:
@brianquinlan – and maybe @devoncarew – need to figure out (1) if this is a pattern we want to support and (2) if this approach is good.

tlueder:
Hi @brianquinlan,

sorry for the noice here. I needed my branch updated and forgot that it is linked to this MR.

I don't know what I can or should do to advance this MR.

brianquinlan:
I think that we have to coordinate landing this (or something like this) with an semver major release and I don't have the bandwidth to work on that right now.

retendo:
What's your proposed workaround to observe the upload progress in Flutter applications at this time?

Issue #581, opened 2021-06-14
Null check exception in io_client.dart file 


guyluz11:
While using [firebase_auth_rest](https://pub.dev/packages/firebase_auth_rest) I came across null check exception from `io_client.dart` file.
[Linking](https://github.com/Skycoder42/firebase_auth_rest/issues/3) the issue to here.

![image](https://user-images.githubusercontent.com/9304740/121868353-8d107d80-cd09-11eb-820b-8086455e761e.png)

`Dart SDK version: 2.13.1 (stable) (Fri May 21 12:45:36 2021 +0200) on "linux_x64"`
Running `Native dart` on Linux (no browser).

**To reproduce** clone all the project from [here](https://github.com/guyluz11/firedart?organization=guyluz11&organization=guyluz11) and run the main file in example folder.
Small changes to the example are needed for it to work:

0. Create firebase project, create new account in the project: in the left drawer click on `Authentication --> Users --> Add user`. 
1. Add your cloud firebase information and user in the top.
```
const apiKey = 'Project Settings -> General -> Web API Key';
const projectId = 'Project Settings -> General -> Project ID';
const email = 'you@server.com';
const password = '1234';
```
2. Replace the part that stop the stream (in the bottom)
```
 finally {
    await firestore?.signOut();
  }

  exit(0);
```
with 
`finally {}`

3. Leave the stream to run couple of hours 3-6.
    I guess some token becoming invalid and the exception will pop up.

guyluz11:
The exception is in the file `io_client.dart:30:36` in line [`await _inner!.openUrl`](https://github.com/dart-lang/http/blob/e89b190936d53d0e36148436283e28ba1091b35a/lib/src/io_client.dart#L30)

```
   /// Sends an HTTP request and asynchronously returns the response.
  @override
  Future<IOStreamedResponse> send(BaseRequest request) async {
    var stream = request.finalize();

    try {
      var ioRequest = (await _inner!.openUrl(request.method, request.url))
        ..followRedirects = request.followRedirects
        ..maxRedirects = request.maxRedirects
        ..contentLength = (request.contentLength ?? -1)
        ..persistentConnection = request.persistentConnection;
      request.headers.forEach((name, value) {
        ioRequest.headers.set(name, value);
      });

      var response = await stream.pipe(ioRequest) as HttpClientResponse;

      var headers = <String, String>{};
      response.headers.forEach((key, values) {
        headers[key] = values.join(',');
      });

      return IOStreamedResponse(
[...]

PR #583, opened 2021-06-17
Fix null check exception in IOClient when sending http request (#581)

guyluz11:


guyluz11:
@googlebot I signed it!

kevmoo:
It seems like you have a case where you're trying to use the client after it's closed.

If anything, we should have a check in `send` that throws a clear error if `_inner` is `null`.

CC @natebosch

guyluz11:
Should I open new clean pull request for it without this commit?

kevmoo:
> Should I open new clean pull request for it without this commit?

That'd probably be best!

Issue #567, opened 2021-05-04
Abort or cancel multipart request


intoxicated:
It seems like neither `abort` or `cancel` method are provided from http package, though there is `abort` in dart:io package for httpClientRequest.

Is it safe to call `close` on client for multipart request? 

```dart
class MultipartRequest extends http.MultipartRequest {
  /// Creates a new [MultipartRequest].
  var client = http.Client();

  MultipartRequest(
    String method,
    Uri url, {
    this.onProgress,
  }) : super(method, url);

  final void Function(int bytes, int totalBytes) onProgress;

  void cancel() => client.close();

  @override
  Future<http.StreamedResponse> send() async {
    try {
      var response = await client.send(this);
      var stream = onDone(response.stream, client.close);
      return http.StreamedResponse(
        http.ByteStream(stream),
        response.statusCode,
        contentLength: response.contentLength,
        request: response.request,
        headers: response.headers,
        isRedirect: response.isRedirect,
        persistentConnection: response.persistentConnection,
        reasonPhrase: response.reasonPhrase,
      );
    } catch (_) {
      client.close();
      rethrow;
    }
  }

  Stream<T> onDone<T>(Stream<T> stream, void Function() onDone) =>
      stream.transform(StreamTransformer.fromHandlers(handleDone: (sink) {
        sink.close();
        onDone();
      }));

  /// Freezes all mutable fields and returns a single-subscription [ByteStream]
  /// that will emit the request body.
  @override
  http.ByteStream finalize() {
    final byteStream = super.finalize();
    if (onProgress == null) return byteStream;

    final total = contentLength;
    var bytes = 0;

    final t = StreamTransformer.fromHandlers(
      handleData: (List<int> data, EventSink<List<int>> sink) {
        bytes += data.length;
        if (onProgress != null) onProgress(bytes, total);
        sink.add(data);
      },
    );
    final stream = byteStream.transform(t);
    return http.ByteStream(stream);
  }
}
```

I saw some error were thrown (can't reproduce again but it was complaining about the connection was closed, it had not been reached content length the request claimed to fulfill).

natebosch:
> Is it safe to call `close` on client for multipart request?

I think it should be.

> it was complaining about the connection was closed, it had not been reached content length the request claimed to fulfill

Was that server side or client side?


> It seems like neither `abort` or `cancel` method are provided from http package

Not currently. Work is tracked in https://github.com/dart-lang/http/issues/424

Issue #204, opened 2018-10-04
Cancellable requests


sir-boformer:
I'm using Dart Http in a Flutter app to download large files from a server.

Sometimes the user may want to cancel the request (so that the download is stopped)

The problem is that there is no way to cancel a `Future` in dart. Even when converted to a `Stream` which can be cancelled, the download will still continue running in the background.

iqbalmineraltown:
does `client.close()` stop the download?

vitor-gyant:
So, is this the proposed way to cancel a given request ? this force us to create a new client per request to allow fine-grained control to which request to cancel.

windrunner414:
+1

MatrixDev:
I don't think that `client.close()` should be considered as a proper solution. At best it can be a workaround. Problem is that `Future`s in Dart are not cancellable and wrapping `Future` in `Stream` cancels stream but stream can't cancel `Future`.

As for me this is a **major** language design problem but Dart team is fighting as hard as they can to not implement cancellation... There are multiple bugs with request for this and all of them are closed as `Won't implement`.

intoxicated:
> As for me this is a **major** language design problem but Dart team is fighting as hard as they can to not implement cancellation... There are multiple bugs with request for this and all of them are closed as `Won't implement`.

I'm wondering what are rationales not to implement cancellation? First impression on `Future` for me, I was like "why they wouldn't design `future` as cancelable", because in many cases one may want to cancel whatever future is doing. Probably there are trade-off doing so, but still. 🤔

MatrixDev:
> > As for me this is a **major** language design problem but Dart team is fighting as hard as they can to not implement cancellation... There are multiple bugs with request for this and all of them are closed as `Won't implement`.
> 
> I'm wondering what are rationales not to implement cancellation? First impression on `Future` for me, I was like "why they wouldn't design `future` as cancelable", because in many cases one may want to cancel whatever future is doing. Probably there are trade-off doing so, but still. 🤔

IMHO they don't do it because of the javascript. These futures are 100% copy of promises from JS and they also can't be canceled. It might be that they are using promises under the hood when compiling for web.

As for me, I was told all my life to use resources sparingly and always close/release what I don't need. This situation with non-cancelable futures just drives me crazy...

natebosch:
Closing as a duplicate of https://github.com/dart-lang/http/issues/424 which has more direct discussion of the API.

> I'm wondering what are rationales not to implement cancellation?

[...]

PR #602, opened 2021-08-02
Request cancellation/abort through cancellation-token

FaFre:
This is an effort trying to tackle the problems discussed in #567, #424, and #204.

This solution introduces a `CancellationToken`, which is passed into the http-methods of `Client` e.g.:
```dart
http.get(url,
      cancellationToken: http.CancellationToken(autoDispose: false));
```
This design is avoiding breaking changes.

Despite most  cancellation-token-based solutions (working with a simple `Completer`), this approach is also trying to really `abort()`a request if necessary.

It is possible to add multiple requests to a `CancellationToken`, to cancel them at once when triggered (disposal of the token has to get done manually in that case).

A request can (if requested) get aborted at multiple points:
1. Before the request is getting prepared
2. When the request is getting registered at the `CancellationToken`
3. When the request is running, through a `Stream`-based invoke of `abort()` 

Status:
Cancellation is supported and implemented for both `IOClient` and `BrowserClient`.
Existing tests have been fixed, but no new tests written yet.
It will need some further reviews.

Problems:
The `abort()` on `HttpRequest` (for web) doesn't seem to have any effect during my tests (request time < 800ms) - although it's called during the according `readyState` (https://api.flutter.dev/flutter/dart-html/HttpRequest/abort.html)

### DIFF
--- a/lib/http.dart
+++ b/lib/http.dart
@@ -6,6 +6,7 @@
 import 'dart:convert';
 import 'dart:typed_data';
 
+import 'src/cancelation_token.dart';
 import 'src/client.dart';
 import 'src/exception.dart';
 import 'src/request.dart';
@@ -16,6 +17,7 @@ export 'src/base_client.dart';
 export 'src/base_request.dart';
 export 'src/base_response.dart';
 export 'src/byte_stream.dart';
+export 'src/cancelation_token.dart';
 export 'src/client.dart';
 export 'src/exception.dart';
 export 'src/multipart_file.dart';
@@ -32,8 +34,10 @@ export 'src/streamed_response.dart';
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> head(Uri url, {Map<String, String>? headers}) =>
-    _withClient((client) => client.head(url, headers: headers));
+Future<Response> head(Uri url,
+        {Map<String, String>? headers, CancellationToken? cancellationToken}) =>
+    _withClient((client) => client.head(url,
+        headers: headers, cancellationToken: cancellationToken));
 
 /// Sends an HTTP GET request with the given headers to the given URL.
 ///
@@ -42,8 +46,10 @@ Future<Response> head(Uri url, {Map<String, String>? headers}) =>
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> get(Uri url, {Map<String, String>? headers}) =>
-    _withClient((client) => client.get(url, headers: headers));
+Future<Response> get(Uri url,
+        {Map<String, String>? headers, CancellationToken? cancellationToken}) =>
+    _withClient((client) => client.get(url,
+        headers: headers, cancellationToken: cancellationToken));
 
 /// Sends an HTTP POST request with the given headers and body to the given URL.
 ///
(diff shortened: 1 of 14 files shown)


natebosch:
@FaFre - did you have a chance to check out the CLA agreement linked above?

FaFre:
@natebosch - haven't had the time to read the agreement yet. Will do it today or tomorrow.

FaFre:
@googlebot I signed it!

aytunch:
@FaFre Thank you very much for this Draft and PR
This is a very important feature.
Just to make sure I have a question:
if user is downloading a big file and they are at 50%
when the download is cancelled, the rest of the file will not be downloaded right?
Also does Firebase Storage give the necesaary server support out of the box?
Thanks Fabian.

FaFre:
@aytunch I'm happy you find use in it :+1: 
Well, it depends. The provided API is pretty high level. The only option we can do is to call [abort](https://api.dart.dev/stable/2.10.2/dart-io/HttpClientRequest/abort.html), but the definition, what actually happens with the underlying TCP connection is not given and will probably also vary depending on the underlying platform libraries/native client.

You need to check that on your own with an package sniffer on each platform you want to provide your app for (unless somebody did already research on that...). I'm not familiar with Firebase storage but I assume you just have an basic HTTP-Download. In that case it should be enough to close the TCP connection to stop the download. And this is what you need to confirm with an package sniffer.

aytunch:
I understand @FaFre thanks
Do you have an ETA for this PR?

FaFre:
@aytunch 
I use the feature already in my App's since a couple of months and didn't encounter any problems with it (not targeting web). 

However, the implementation for web may need some further investigation. Although I implemented it like in the documentation described, my requests didn't get cancelled accordingly. I couldn't fix that on my own so far, but it's also possible that this effect is caused by the dart libraries and I do everything correctly here.

FaFre:
@aytunch  While doing some research on the web request I actually discovered that also the dart:io implementation isn't able to abort requests once the server started sending data.

The implementation seems to see a request as done once the server starts sending data. Therefore they cannot get aborted anymore.

```dart
final bigFileUrl = Uri.https('ftp5.gwdg.de',
      '/pub/misc/openstreetmap/planet.openstreetmap.org/pbf/planet-latest.osm.pbf'); //~60GB file
  final req = await HttpClient().getUrl(bigFileUrl);
  unawaited(Future.delayed(const Duration(seconds: 1)).whenComplete(() {
    req.abort();
    print('sent abort');
  }));

  final resp = await req.close();
  print(await resp.toList());
```

The code above confirms with DevTools that the HTTP request is still running and memory is filling up.

@natebosch I saw you also did some research on abort() and opened some issues regarding it. Is this behavior already discussed somewhere?

aytunch:
@FaFre Have you seen this issue?
https://github.com/dart-lang/http/issues/424

natebosch:
> Is this behavior already discussed somewhere?

I think once we have a response, instead of `request.abort()` we need to use response.detachSocket()`.

https://github.com/dart-lang/http/pull/521/files#diff-7a772406b9a2c46d9b0b95da73d10dd5ad9807fa78c9519bac0d247690482957R83


I do have plans to revisit timeouts and request aborting soon. I suspect we won't be able to use this PR because it would be a breaking change that might be too hard to roll out, but I have not yet finished exploring the implications of various designs.

piotrsed (approved):

JonathanPeterCole:
I've been working on a fork with cancellation using a CancellationToken, but I've taken a different approach with the tokens so they're not specific to HTTP requests: https://github.com/JonathanPeterCole/dart-cancellation-token-http
I haven't had a chance to properly test this implementation in a Flutter app yet though.

PR #623, opened 2021-09-30
Added the ability to retrieve `url` from `BaseResponse`

elmoiv:


elmoiv:
@googlebot I signed it!

kevmoo:
Add some documentation to the parameter and the field?

Why are you adding this? How does it help you?

elmoiv:
It comes in handy when working with redirected urls. Like if I want to get the final retrieved URL after all redirects are done.

This adds the ability to get that from a `http.get` instead of the current workaround of setting `followRedirects` to false in `http.Request` or trying to check if response headers has the final url or not.

This was inspired from Python `requests` library

elmoiv:
It appears that Dart CI Chrome tests are broken.

piotrsed (approved):

natebosch:

PR #635, opened 2021-11-01
Added "Follow Redirects" as a top level parameter to all request functions

owenkealey:
On multiple occassions, I have needed to specify "follow redirects" on/off in specific GET/POST requests that I don't feel warranted the boilerplate of implementing BaseRequest. Thus, I added followRedirects as a top level parameter to all core http functions(.get, .post, .head etc).


lib/src/response.dart
```diff
@@ -25,7 +26,18 @@ class Response extends BaseResponse {
   /// [RFC 2616][].
   ///
   /// [RFC 2616]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html
-  String get body => _encodingForHeaders(headers).decode(bodyBytes);
+  String parseBody() {
```
kevmoo, line 29: This seems unrelated to the title of the PR

kevmoo (commented):
Looks like some `build` files were added that should be removed!

lib/src/response.dart
owenkealey: Your'e correct. I use my fork for personal projects and forgot this PR was still open. I am going to close and open a new one with just the follow redirects only.

kevmoo: I'd do seperate PRs for protli and proxy support

PR #657, opened 2022-01-12
Fix response request in MockClient

teodor-appd:
This fix helps solve the following cases:

```dart
MockClient((request) async {
  request.headers.addAll({"Content-Type": "application/json"});
  return Response("{}", 200, request: request);
});
```

By having `baseRequest` instead of `response.request` in `mock_client.dart` we would never be able to capture the new request headers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## 0.13.5-dev
 
 * Allow async callbacks in RetryClient.
+* Use callback returned Response.request instead of the passed request to give
+  more control. This may be breaking for callbacks which return incomplete
+  Responses and relied on the default.
 
 ## 0.13.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## 0.13.5-dev
 
 * Allow async callbacks in RetryClient.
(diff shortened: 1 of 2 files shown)


teodor-appd:
@natebosch Please rescan.

natebosch:
This seems pretty reasonable. I'll need to do some validation.

natebosch:
This is technically breaking, but it doesn't impact any of our internal tests, and I think the chances of it causing much trouble are low. If it does cause a breakage it will be only in a test, and the fix should be straightforward.

I'm comfortable with pushing this out without a major version bump. I think usage of the API at all is low, and it's better to give the more useful thing to the folks actually using it.  @jakemac53 - just to have a second set of eyes on it, do you have any concerns?

jakemac53 (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## 0.13.5-dev
 
 * Allow async callbacks in RetryClient.
+* Use callback returned Response.request instead of the passed request to give
```
jakemac53, line 4: On its own this entry is not easy to understand, maybe just add a mention that this is for the MockHttpClient

natebosch: Done

Issue #645, opened 2021-12-02
Update headers on retry


SeriousMonk:
Hi,
I'm using the RetryClient() to refresh the access token if expired and then try the request again.
However when retrying the request, the old access token is sent and not the new one.
This is how I configured my RetryClient:
```dart
RetryClient client = RetryClient(
  http.Client(),
  retries: 1,
  when: (response) => response.statusCode == 401,
  onRetry: (req, res, retryCount) async{
    await authService.refreshToken();
    req.headers['Authorization'] = sharedPrefs.accessToken;
    }
);
```
I'm guessing that the req element passed to onRetry can't be used to update the req that is actually sent when retrying.
How can I solve this?

Thanks :)

PR #679, opened 2022-03-16
Async support for when, whenError and onRetry

beroso:
Addresses issue #645 .

In a general way, it add support for assigning async callbacks to when, whenError and onRetry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.13.5-dev
+
+* Allow async callbacks in RetryClient.
+
 ## 0.13.4
 
 * Throw a more useful error when a client is used after it has been closed.
--- a/lib/retry.dart
+++ b/lib/retry.dart
@@ -18,16 +18,16 @@ class RetryClient extends BaseClient {
   final int _retries;
 
   /// The callback that determines whether a request should be retried.
-  final bool Function(BaseResponse) _when;
+  final FutureOr<bool> Function(BaseResponse) _when;
 
   /// The callback that determines whether a request when an error is thrown.
-  final bool Function(Object, StackTrace) _whenError;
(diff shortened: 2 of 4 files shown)


kevmoo:
Sadly, we cannot do this without releasing a major version. This is a breaking change.

CC @natebosch

natebosch:
I do think this looks like a nice change, but we can't do a major bump right now.

If you file an issue we can tag it so that it gets attention once we are ready to push through a major version.

beroso:
Thanks for the feedback.

Sorry, but I thought it wasn't a breaking change because I'm using `FutureOr`. The previous tests are passing and also tried to subclass the `RetryClient` with these changes and it worked.

There is a chance to reconsider this PR, because in my opinion it is backwards compatible?

@kevmoo , @natebosch, could you show me an example of api breaking with these changes? Maybe I'm missing the point.

natebosch:
Any code with `@override` of these methods will break. This is non breaking for _uses_ of the client, but it is breaking for _implementations_ of the client. Both are supporting by this package.

beroso:
@natebosch I don't get it, because I just changed the type of private variables, not methods. The private vars are not exposed by the API. And the implementations constructors continues to working properly.

Could you show a concrete example of breaking?

natebosch:
Ah, I think @kevmoo and I were mistaken that there were changed parameters on methods instead of constructors.

I do think this is non-breaking. I'll need to investigate whether it has impact on my current approach for https://github.com/dart-lang/http/issues/424

beroso:
Thank you!

lrhn:
I believe this is practically non-breaking (almost everything is technically breaking, but it requires people to go through some hoops to be broken).
Changes look good to me, so just need @natebosch to be fine too,

natebosch (approved):

PR #681, opened 2022-03-17
Added "Follow Redirects" as a top level parameter to all request functions

owenkealey:
This is a duplicate of an earlier(now closed) PR with only those changes that implement the "follow redirects" change. 

On multiple occassions, I have needed to specify "follow redirects" on/off in specific GET/POST requests that I don't feel warranted the boilerplate of implementing BaseRequest. Thus, I added followRedirects as a top level parameter to all core http functions(.get, .post, .head etc).


owenkealey:
@kevmoo Here is the PR for just follow redirects

PR #682, opened 2022-03-18
Add support for brotli compressed response bodies

owenkealey:
Most websites now prefer to send response bodies compressed with [Brotli](https://en.wikipedia.org/wiki/Brotli) because it usually results in better compression results.

The current dart client doesn't support this and so this PR adds Brotli decompression support.


owenkealey:
@kevmoo And here is the one for brotli. Thanks!

kevmoo:
I doubt we'll want to add a dependency to `pkg:brotli` to a core package.

@natebosch ?

natebosch:
We don't want to take a dependency on this package, but it should be feasible to make it pluggable. I can't recall an existing feature request for this (but I haven't searched). Feel free to file one.

Issue #24, opened 2015-06-05
package:http abstracts request headers as a Map<string, string>; that does not work for repeating headers


DartBot:
_Originally opened as dart-lang/sdk#21802_

_This issue was originally filed by jirkad...&#064;gmail.com_

---

Package http represents request headers as a Map&lt;string, string&gt;, both in requests as well as in responses. That does not work for repeating headers, In that case one needs to fold them into a single string. Which is not always practical or possible.

**What steps will reproduce the problem?**

The most problematic header is Set-Cookie since the concatenation of multiple cookies into one string is tricky to parse.

**Please provide any additional information below.**

Some browsers and libraries e.g. http.cookies in Python can deal with folded Set-Cookie headers.

RFC 6265 says &quot;don't do that&quot; (section 3)

DartBot:
<a href="https://github.com/anders-sandholm"><img src="https://avatars.githubusercontent.com/u/3276024?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [anders-sandholm](https://github.com/anders-sandholm)**

----

*Added [Area-Pkg](../labels/Area-Pkg), [Pkg-Http](../labels/Pkg-Http), [Triaged](../labels/Triaged) labels.*

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

The plan here for both http and shelf is to provide a way to access headers that preserves repetition for cases like this. Something like a multiHeaders hash or a getMultiHeader method would work here. There will probably also be a more specialized way of accessing parsed cookies as well (issue dart-lang/http#20).

Unfortunately I probably won't have time to work on this for a while yet. If you want to take a stab, I can certainly review patches!
______
*Added [PatchesWelcome](../labels/PatchesWelcome) label.*

DartBot:
*This comment was originally written by softwaredev...&#064;gmail.com*
_____
What about copying the API dart:io has?

dart:io.HttpClientResponse:
- List&lt;Cookie> get cookies
- HttpHeaders get headers

dart:io.HttpHeaders seems quite a high-level general class. I would've preferred if it conformed to Map&lt;String, List&lt;String>> interface, though then there would be the problem of naming the set() and add() methods. The Map.addAll would actually mean calling set() for each value, not add())… So it might be better as it is.

[...]

PR #688, opened 2022-04-19
feat: Add feature to parse multiple set-cookies

myConsciousness:
Hi amazing developers,

Perhaps this is already on the agenda before, but the `http` library combines multiple `set-cookie`s in the response header into a comma-separated list that is set in the `headers` field. This is a rather complicated specification and not one that developers can easily handle.

Therefore, I have added the feature to parse `set-cookie`s contained in `headers` to make it easier to handle multiple `set-cookie`s.  This modification is not disruptive and the `headers` field will continue to function as before.

However, in the future you need to reintegrate `cookies` field when you make the `headers` field an object, as was written in the `BaseResponse` TODO.

Also it was necessary to add `universal_io` to the dependencies to use `Cookie` object.

Thank you.

### DIFF
--- a/pkgs/cupertino_http/darwin/cupertino_http/Sources/cupertino_http/utils.h
+++ b/pkgs/cupertino_http/darwin/cupertino_http/Sources/cupertino_http/utils.h
@@ -7,7 +7,7 @@
 #error "This file must be compiled with ARC enabled"
 #endif
 
-typedef void (^_DidFinish)(void *, NSURLSession *session,
+typedef void (^_DidFinish)(void *closure, NSURLSession *session,
                            NSURLSessionDownloadTask *downloadTask,
                            NSURL *location);
 typedef void (^_DidFinishWithLock)(NSCondition *lock, NSURLSession *session,
--- a/pkgs/cupertino_http/darwin/cupertino_http/Sources/cupertino_http/utils.m
+++ b/pkgs/cupertino_http/darwin/cupertino_http/Sources/cupertino_http/utils.m
@@ -1,7 +1,7 @@
 #import "utils.h"
 
 _DidFinish adaptFinishWithLock(_DidFinishWithLock block) {
-  return ^void(void *, NSURLSession *session,
+  return ^void(void *closure, NSURLSession *session,
                NSURLSessionDownloadTask *downloadTask, NSURL *location) {
     NSCondition *lock = [[NSCondition alloc] init];
     [lock lock];


devoncarew:
Thanks for the PR! Some thoughts:

- this is a highly used package, so we generally want to be very conservative about adding a new package dependency. You're adding `package:universal_io` - I assume to get the definition of the `Cookie` class? We'll likely have to remove the dep. It looks like package:universal_io exports the dart:io `Cookie` class in a dart:io context, and defines it's own class in other contexts. I don't know if that makes sense to do in this library, but fyi
- adding @natebosch for review (and @brianquinlan for fyi, in case he has comments on the PR)

myConsciousness:
Thanks for your feedback @devoncarew !

This is exactly what I was thinking too. 

I used to get an error in static analysis when I uploaded a package I was developing that used `dart:io` to `Pub.dev`, and this is reason for adding `universal_io` but now it seems I misunderstood the cause of that error.

And now I have scrutinized the `universal_io` implementation and found no reason to use `universal_io` instead of `dart:io` in this fix. I'm going to refactor my pull request.

Thank you.

lib/src/base_response.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
```
natebosch, line 5: We can't import `dart:io` from this library. `universal_io` is intending to get around that problem, but we don't want to take a dependency on that library either. If we wanted to have this functionality in this library we'd have to reimplement an interface for `Cookie`.

This feels like the type of thing that should be added on in a side package (which could more easily have a dependency on `universal_io`)

```dart
extension Cookies on Response {
  List<Cookie> get cookies {
    // ...
  }
}
```

I suppose the nice thing about implementing it here is being able to cache it, but using an `expando` is probably a fine alternative if caching is important. Probably you'll only read `.cookies` once anyway.

myConsciousness: Thanks for your review @natebosch !

I found several issues have been raised regarding this case of multiple `set-cookie`s, and I feel that a fundamental solution is needed. At the same time, I also understood the difficulties related to `dart:io`. It's quite a deep-rooted problem.

It's possible to add a `Cookie` interface and implementation, but is there a reluctance to add a `Cookie` object as a development policy for the http library?

If we can implement `Cookie` interface in http library, I thought the following example was a very good solution.

```dart
extension Cookies on Response {
  List<Cookie> get cookies {
    // ...
  }
}
```

vaind:
FYI this is related to https://github.com/dart-lang/http/issues/24 which proposes a different approach

test/response_test.dart
```diff
@@ -70,4 +70,50 @@ void main() {
...
+      'set-cookie':
+          // This response header contains 6 set-cookies
+          'AWSALB=AWSALB_TEST; Expires=Tue, 26 Apr 2022 00:26:55 GMT; Path=/,AWSALBCORS=AWSALBCORS_TEST; Expires=Tue, 26 Apr 2022 00:26:55 GMT; Path=/; SameSite=None; Secure,jwt_token=JWT_TEST; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT; SameSite=lax; Secure,csrf_token=CSRF_TOKEN_TEST_1; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT,csrf_token=CSRF_TOKEN_TEST_2; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT,wuuid=WUUID_TEST'
+    });
+
+    expect(response.cookies.length, 6);
+    for (final cookie in response.cookies) {
+      expect(
+          cookie.name,
+          anyOf([
```
vaind, line 88: this makes the test pass if `response.cookies` is always just a single cookie, e.g. the first one, repeated 6 times. This isn't so unimaginable in case of an error in the parsing code.

XNORGATE:
up ,there's still an issue about parsing multiple set-cookies , all set-cookies will be covered by last one

PR #692, opened 2022-05-06
Fixes #293, added property url to BaseResponse. related work(#321)

### LINKED ISSUE #293: Get final redirected URL from response

glassmonkey:
This is almost a copy of this Pull Request #321.

> The final url is extract from XMLHttpRequest.responseURL(on browsers) or HttpClientResponse.redirects.last.location(on vm).
Also added tests for vm.

Additional work includes support for Null Safety and tests created from streams.

I've create new patch, since the original author @YizhePKU mention the following in #293.

> Alas, I've long forgotten about this and have deleted my repo. The content of the patch is still viewable here.

If there are any deficiencies in this pull request, please let me know so I could take additional action.

### DIFF
--- a/lib/src/base_request.dart
+++ b/lib/src/base_request.dart
@@ -136,6 +136,7 @@ abstract class BaseRequest {
           contentLength: response.contentLength,
           request: response.request,
           headers: response.headers,
+          url: response.url,
           isRedirect: response.isRedirect,
           persistentConnection: response.persistentConnection,
           reasonPhrase: response.reasonPhrase);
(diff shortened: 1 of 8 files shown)


glassmonkey:
@googlebot I signed it!

kevmoo:
CC @brianquinlan

kevmoo:
We have to be careful adding APIs here – they are breaking for common base classes.

glassmonkey:
I very much agree with that opinion.
I appreciate your thoughtful review.

As a matter of urgency, I'm going to use my own forked PATCH to get by, so no problem.

natebosch:
I think the main way this is breaking is for stubbed requests, I've found at least one case where `redirects` is `throw UnimplementedError` so this change causes existing tests to fail.

I think if we find it's only breaking for test code it may be OK to go forward with this change anyway, but we'll have to be careful during the rollout.

natebosch:

Issue #696, opened 2022-05-23
Allow the user to select their own `Client` in addition to using the `IOClient` and `BrowserClient` defaults


brianquinlan:
It should be possible for users of `package:http` to select their `Client` implementation in addition to the provided defaults.

Something like:

```dart
void main() {
  if (Platform.isAndroid) {
    http.Client.global = AndroidAPIsClient(<optional configuration>);
  } else if (Platform.isIOS) {
    http.Client.global = iOSAPIsClient(<optional configuration>);
  }
}

void someFunc() {
  final client = http.Client();  // Selects the implementation setup in `main`
  final resp = await client.get(Uri.parse("https://www.example.com/"))
}
```

PR #697, opened 2022-05-23
Implement the ability to run a particular Client implementation in a Zone

brianquinlan:
Fixes https://github.com/dart-lang/http/issues/696


lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
   /// do so can cause the Dart process to hang.
   void close();
 }
+
+const _clientToken = Object();
+
+/// Return the [Client] for the current [Zone], if one has been set.
+///
+/// NOTE: This function is explicitly hidden from the public API.
+Client? getZonedClient() => Zone.current[_clientToken] != null
```
natebosch, line 150:
```suggestion
Client? get zoneClient => Zone.current[_clientToken] != null
```
https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+///
+/// myFunction() {
+///   // Uses the `Client` configured in `main`.
+///   final response = await get(Uri.parse("https://www.example.com/"));
+///   final client = Client();
+/// }
+/// ```
+///
+/// The [client] influences functions (e.g. `get`) and the [Client.new] factory.
+R runClientZoned<R>(R Function() body, Client client,
```
natebosch, line 184: How about `runWithClient`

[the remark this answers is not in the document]
brianquinlan: I don't see a use case but that doesn't mean that there isn't one ;-)

You mean that performance of a single callback? It's hard to imagine that mattering. We would be introducing a tiny inconsistency with `HttpOverrides` but I don't think that matters either.

brianquinlan:
I'm happy to make these changes. Does the overall approach seem reasonable?

natebosch:
> Does the overall approach seem reasonable?

Yeah I think this looks like a good approach. cc @lrhn

lib/src/client.dart
brianquinlan: Done.

natebosch (approved):

aam:
I don't have strong opinion, but it feels strange how global(behind-the-scenes)-nature of Zone variables is intertwined with local-nature of run-with-this-client `runWithClient`. I basically contrast it with https://api.flutter.dev/flutter/dart-io/HttpOverrides-class.html where once you run in zone, zone overrides control everything that happens there.

brianquinlan:
> I don't have strong opinion, but it feels strange how global(behind-the-scenes)-nature of Zone variables is intertwined with local-nature of run-with-this-client `runWithClient`. I basically contrast it with https://api.flutter.dev/flutter/dart-io/HttpOverrides-class.html where once you run in zone, zone overrides control everything that happens there.

I'm not sure what you mean - the implementation approach is very similar to HttpOverrides.

aam:
> > I don't have strong opinion, but it feels strange how global(behind-the-scenes)-nature of Zone variables is intertwined with local-nature of run-with-this-client `runWithClient`. I basically contrast it with https://api.flutter.dev/flutter/dart-io/HttpOverrides-class.html where once you run in zone, zone overrides control everything that happens there.
> 
> I'm not sure what you mean - the implementation approach is very similar to HttpOverrides.

Yeah, true. I got confused by `runWithClient` name.
Would it make sense to make this api identical to existing one - introduce `HttpClientOverrides` class with `runZoned` method and perhaps point out how it is similar to `dart:io` `HttpOverrides`?

brianquinlan:
> > > I don't have strong opinion, but it feels strange how global(behind-the-scenes)-nature of Zone variables is intertwined with local-nature of run-with-this-client `runWithClient`. I basically contrast it with https://api.flutter.dev/flutter/dart-io/HttpOverrides-class.html where once you run in zone, zone overrides control everything that happens there.
> > 
> > 
> > I'm not sure what you mean - the implementation approach is very similar to HttpOverrides.
> 
> Yeah, true. I got confused by `runWithClient` name. Would it make sense to make this api identical to existing one - introduce `HttpClientOverrides` class with `runZoned` method and perhaps point out how it is similar to `dart:io` 
`HttpOverrides`?

I was using the advice here:
https://dart.dev/guides/language/effective-dart/design#classes-and-mixins

Right now my class would only have a single method (i.e. `createClient`) and adding more would break any code that `implements` that class.

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
   /// do so can cause the Dart process to hang.
   void close();
 }
+
+const _clientToken = Object();
+
+/// Return the [Client] for the current [Zone], if one has been set.
```
lrhn, line 147: Document getters using noun phrases. I'd just remove the "Return". (Nit: Which, if it had stayed, should be active form "Returns".)

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+///
+/// NOTE: This function is explicitly hidden from the public API.
```
lrhn, line 149: function -> getter (or property)

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+///  late Client client;
+///  if (Platform.isAndroid) {
+///    client = MyHttpClient();
+///  } else {
+///    client = Client();  // Use the default client.
+///  }
+///  runWithClient(myFunction, () => client);
+/// }
+///
+/// myFunction() {
```
lrhn, line 176: Write return type, probably `void`.

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+///   // Uses the `Client` configured in `main`.
+///   final response = await get(Uri.parse("https://www.example.com/"));
+///   final client = Client();
+/// }
+/// ```
+///
+/// The [client] influences functions (e.g. `get`) and the [Client.new] factory.
```
lrhn, line 183: "Influences" is too vague. Could mean anything.
```dart
/// The [Client] created by [client] is returned by the [Client.new] 
/// factory constructor, and is used by functions like [get].
```
(Don't know where `get` is defined, but make sure to link to it.)

test/io/client_test.dart
```diff
@@ -132,4 +146,22 @@ void main() {
 
     expect(socket, isNotNull);
   });
+
+  test('runWithClient', () {
+    late http.Client client;
+    http.runWithClient(() => client = http.Client(), () => TestClient());
```
lrhn, line 152: Nit: Can use constructor tear-offs, so `TestClient.new` instead of `() => TestClient()`

lrhn, line 152: Can just do
```dart
 http.Client client = http.runWithClient(() => http.Client(), TestClient.new);
```
since you return the value anyway.
(Can also use a constructor tear-off for that, but I don't find that actually more readable:
```dart
 http.Client client = http.runWithClient(http.Client.new, TestClient.new);
```
)

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+R runWithClient<R>(R Function() body, Client Function() client,
```
lrhn, line 184: *Consider* putting the `body` argument last. It reads better as:
```dart
runWIthClient(createClient, () {
   the body 
   of code
});
```
than putting the (usually small) client factory after the bigger body.

(Try it, and compare the readability of the "nested" tests.)

lib/src/client.dart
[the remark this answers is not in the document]
lrhn: It's potentially useful to pass in a factory instead of a single value. You can always pre-compute the value and pass in a function returning it every time, the other direction is not as interesting.

If we do that, we should/must do `Zone.current.bind(factory)` before storing the factory, so we are sure it's called in the parent zone.
That ensures that the user can have a factory like `() => someTest ? MyClient() : Client()` and not get an infinite recursion when that `Client()` is run in the child zone which goes right back and calls the factory again.

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
 }
+
+const _clientToken = Object();
+
+/// Return the [Client] for the current [Zone], if one has been set.
+///
+/// NOTE: This function is explicitly hidden from the public API.
+Client? get zoneClient => Zone.current[_clientToken] != null
+    ? Zone.current[_clientToken] as Client
+    : null;
```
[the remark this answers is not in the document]
brianquinlan: Done

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
   /// do so can cause the Dart process to hang.
   void close();
 }
+
+const _clientToken = Object();
```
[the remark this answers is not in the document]
brianquinlan: The symbol syntax is now for me. Did I capture your intent?

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+/// ```
+/// class MyHttpClient extends http.BaseClient {
+///   @override
+///   Future<http.StreamedResponse> send(http.BaseRequest request) {
+///     // your implementation here
+///   }
+/// }
+///
+/// void main() {
+///  late Client client;
```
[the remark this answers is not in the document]
brianquinlan: Done.

lib/src/client.dart
```diff
@@ -140,3 +141,48 @@ abstract class Client {
...
+///   final response = await get(Uri.parse("https://www.example.com/"));
+///   final client = Client();
+/// }
+/// ```
+///
+/// The [client] influences functions (e.g. `get`) and the [Client.new] factory.
+R runWithClient<R>(R Function() body, Client Function() client,
+        {ZoneSpecification? zoneSpecification}) =>
+    runZoned(body,
+        zoneValues: {_clientToken: client()},
```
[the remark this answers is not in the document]
brianquinlan: Done.

lib/src/client.dart
brianquinlan: Done.

test/io/client_test.dart
brianquinlan: http works back to dart 2.14 so I can't use tearoffs.

brianquinlan: Done (without the tear-offs since they aren't available in Dart 2.14)

lib/src/client.dart
natebosch: I don't think it's necessarily a good idea, but we could consider a named argument for `createClient` and then the caller can put it in either order.

lib/src/client.dart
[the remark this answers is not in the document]
brianquinlan: Done and added a test.

lib/src/client.dart
brianquinlan: Done.

lib/src/client.dart
```diff
@@ -140,3 +141,46 @@ abstract class Client {
   /// do so can cause the Dart process to hang.
   void close();
 }
+
+final _clientToken = Object();
```
lrhn, line 145: Not used any more, can be deleted.

The `#_clientToken` represents an identifier in the current library, in the abstract. There doesn't have to be any declaration of that name.

lib/src/client.dart
lrhn: I considered named parameters, but I'm generally not fond of required named parameters. Because I hate writing the name.

That said, in this case both arguments are likely to be function literals, so it might be confusing to both authors and readers which is which. That is one of the situations where the style guide suggests naming parameters.

I disagree with the ordering. You want the small method body *first*. If it's very short, like a single line, you can see past it and find the other function. A large first function literal makes it very hard to find the second one.

Being consistent has value, so let's just go with body last. If someone disagrees, they can write their own two-parameter wrapper function.

lrhn (approved):
OK with the last three comments addressed.

lib/src/client.dart
natebosch: > Being consistent has value, so let's just go with body last.

Do you mean we should be consistent with [`HttpOverrides.runWithHttpOverrides`](https://api.flutter.dev/flutter/dart-io/HttpOverrides/runWithHttpOverrides.html)? In that case it should be `body` first.

aam (approved):

lib/src/client.dart
brianquinlan: Done.

lib/src/client.dart
brianquinlan: Done.

lib/src/client.dart
brianquinlan: > "Influences" is too vague. Could mean anything.

Done.

brianquinlan:
I think that the only thing to decide is whether the `body` or the `clientFactory` should be first *or* whether they should be keyword arguments.

HttpOverrides has `body` first and I have no idea which function is likely to be larger.

lib/src/browser_client.dart
```diff
@@ -9,13 +9,14 @@ import 'dart:typed_data';
 import 'base_client.dart';
 import 'base_request.dart';
 import 'byte_stream.dart';
+import 'client.dart';
```
natebosch, line 12: [nit] May as well revert `lib/src/browser_client.dart`, `client_stub.dart`, and `io_client.dart`. The type changes aren't needed anymore with the `zoneClient` usage removed also.

lib/http.dart
```diff
@@ -159,7 +159,7 @@ Future<Uint8List> readBytes(Uri url, {Map<String, String>? headers}) =>
     _withClient((client) => client.readBytes(url, headers: headers));
 
 Future<T> _withClient<T>(Future<T> Function(Client) fn) async {
-  var client = Client();
+  var client = zoneClient ?? Client();
```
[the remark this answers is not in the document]
brianquinlan: Doh! Done.

lib/src/browser_client.dart
brianquinlan: Done.

lib/src/client.dart
```diff
@@ -140,3 +141,44 @@ abstract class Client {
   /// do so can cause the Dart process to hang.
   void close();
 }
+
+/// The [Client] for the current [Zone], if one has been set.
+///
+/// NOTE: This property is explicitly hidden from the public API.
```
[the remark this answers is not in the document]
brianquinlan: Done.

brianquinlan:
@lrhn `body` **first** would be consistent with `HttpOverrides`. I just want to be sure that you are happy with that before I submit.

PR #699, opened 2022-05-25
Add Response.url

### LINKED ISSUE #293: Get final redirected URL from response

natebosch:
Closes #321, closes #623, closes #692
Fixes #293

Add a Uri field to BaseResponse with the final, potentially redirected,
url for the content.

The field is nullable for backwards compatibility - TODO: consider if
this should be non-nullable from the start, or if it can be published
nullable first and become non-nullable (and the constructor arg
required) in the next breaking release.

This may break tests which use mocks but do not mock the field used to
read the URL.

### DIFF
--- a/lib/src/base_request.dart
+++ b/lib/src/base_request.dart
@@ -137,6 +137,7 @@ abstract class BaseRequest {
           request: response.request,
           headers: response.headers,
           isRedirect: response.isRedirect,
+          url: response.url,
           persistentConnection: response.persistentConnection,
           reasonPhrase: response.reasonPhrase);
     } catch (_) {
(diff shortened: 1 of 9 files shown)


natebosch:
This is going to need more validation and potentially a rollout plan.

glassmonkey:
Thankyou !!

glhvta:
Hello, could you please update on the status of this PR? We faced the issue in our project that getting the final redirected URL is not possible from http.Response. It would be great to have this functionality. Thanks!

natebosch:
This is blocked on a breaking version change.

tomcruisesfronttooth:
Any update on this?

PR #700, opened 2022-05-25
Use spawnHybrid for the stub server

natebosch:
Move the in-process stub server from `test/io/utils.dart` to
`test/stub_server.dart` as a `hybridMain`. This will allow spawning the
server from the browser tests as well.

The browser tests still do not work against this server. Separate out
the change in how the server is started from the (potential) other
changes with fixes.

Update tests to use the new pattern - they call `startServer` and wait
for the `serverUrl`. No explicit teardown is needed because the
`spawnHybrid` call gets an implicit `addTearDown` which kills the
isolate.

Remove `test/io/utils.dart` entirely. Inline the definition of
`throwsSocketException` at the single use site.
Make the `message` argument to the other copy of `throwsClientException`
optional and use it in place of the version that had been in
`test/io/utils.dart`.


test/utils.dart
```diff
@@ -110,5 +110,19 @@ class _BodyMatches extends Matcher {
...
+  if (message != null) {
+    exception = exception.having((e) => e.message, 'message', message);
+  }
+  return throwsA(exception);
+}
+
+/// Spawn an isolate in the test runner with an http server.
+///
+/// The server isolate will be killed on teardown.
+Future<Uri> startServer() async {
```
kevmoo, line 124: No logic to tear-down the server?

kevmoo (approved):

test/utils.dart
natebosch: No, see the last line of the doc above.

`spawnHybridUri` handles killing the isolate for us. If the isolate is killed there is nothing to tear down.

kevmoo:
Any intention to setup browser testing here, too?

natebosch:
> Any intention to setup browser testing here, too?

Yes, the commit message calls out that this is working towards browser testing but they still do not work.

At the least we'll need to do some different stuff with headers to avoid the browser behavior from clouding the tests. Even after that I was hitting some problems so I'll have to dig more sometime.

PR #706, opened 2022-06-01
Add a conformance test package for http Clients

brianquinlan:


pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -0,0 +1,104 @@
...
+      //        ... |           ...
+      //    ".../1" |           "/"
+      //        "/" |  <no redirect>
+      server = (await HttpServer.bind('localhost', 0))
+        ..listen((request) async {
+          if (request.requestedUri.pathSegments.isEmpty) {
+            unawaited(request.response.close());
+          } else if (request.requestedUri.pathSegments.last == 'loop') {
+            unawaited(request.response
+                .redirect(Uri.parse('http://localhost:${server.port}/loop')));
```
natebosch, line 30: [optional] I prefer using the more specific constructors since we have more semantic info than parse needs.

```suggestion
                .redirect(Uri.http('localhost:${server.port}', 'loop')));
```

Or in this case since we repeat this snippet a bunch, maybe add a little utility like

```dart
extension LocalhostPath on HttpServer {
  Uri localhostUri([String path = '']) => Uri.http('localhost:$port', path);
}
```

Then usage would be

```suggestion
                .redirect(server.localhostUri('loop')));
```



Similar throughout. If you'd find the edits annoying feel free to skip and I can handle in a followup.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
```
natebosch, line 6: Should we try to make these work for testing web implementation of clients as well?

If we want this to be cross platform we'd need to use a `spawnHybrid` call to run the server in the test runner VM.
It would require defining how we'd control the server behavior over a stream channel, I haven't though much yet about a design for such an API.

See https://github.com/dart-lang/http/pull/700
In that case it was easy because the server behavior is static and doesn't interact directly with the test other than through the HTTP calls so we didn't need to do anything but communicate the port.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: Done the first variation. I don't like having too many helpers in my tests because it reduces readability.

pkgs/http_client_conformance_tests/lib/src/request_headers_tests.dart
```diff
@@ -0,0 +1,92 @@
...
+      late HttpHeaders requestHeaders;
+      final server = (await HttpServer.bind('localhost', 0))
+        ..listen((request) async {
+          await request.drain<void>();
+          requestHeaders = request.headers;
+          unawaited(request.response.close());
+        });
+      await client.get(Uri.parse('http://localhost:${server.port}'),
+          headers: {'foo': 'bar', 'Foo': 'Bar'});
+      expect(requestHeaders['foo']!.first, isIn(['bar', 'Bar']));
```
natebosch, line 55: Is there a reason we aren't sure whether one or both values will be included?

```suggestion
      expect(requestHeaders['foo']!.single, isIn(['bar', 'Bar']));
```
```suggestion
      expect(requestHeaders['foo'], unorderedEquals(['bar', 'Bar']));
```

pkgs/http_client_conformance_tests/lib/src/request_headers_tests.dart
```diff
@@ -0,0 +1,92 @@
...
+        });
+      // The `http.Client` API does not offer a way of sending the name field
+      // more than once.
+      await client.get(Uri.parse('http://localhost:${server.port}'),
+          headers: {'list': 'apple, orange'});
+      expect(requestHeaders['list'], ['apple, orange']);
+      await server.close();
+    });
+
+    test('multiple values per header', () async {
```
natebosch, line 75: Is there a difference in the test body from the one above?

Can we drop one of these tests? Maybe with a comment indicating that the signature limitation makes it redundant?

Or should one of the tests have a different body? Or am I overlooking the difference?

natebosch (approved):

pkgs/http_client_conformance_tests/lib/src/request_headers_tests.dart
brianquinlan: No, I screwed up the test. Now there are two:
1. One with multiple headers
2. One with multiple values for one header

PR #707, opened 2022-06-01
Use the Uri.http constructor in docs and tests

natebosch:
Use `Uri.parse` when the string is completely opaque to the calling
code. When the calling code is hardcoding the protocol, use the protocol
specific constructor.

### DIFF
--- a/pkgs/http/README.md
+++ b/pkgs/http/README.md
@@ -15,12 +15,12 @@ you to make individual HTTP requests with minimal hassle:
 ```dart
 import 'package:http/http.dart' as http;
 
-var url = Uri.parse('https://example.com/whatsit/create');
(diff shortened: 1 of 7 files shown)


pkgs/http/README.md
```diff
@@ -87,7 +87,7 @@ import 'package:http/retry.dart';
 Future<void> main() async {
   final client = RetryClient(http.Client());
   try {
-    print(await client.read(Uri.parse('http://example.org')));
+    print(await client.read(Uri.http('example.org', '')));
```
brianquinlan, line 90: I really dislike reading this when the path is empty - as a reader, how do I figure out what that argument is meant to be?

brianquinlan (approved):

PR #708, opened 2022-06-02
Run the redirect tests against the browser client.

brianquinlan:


pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -2,60 +2,40 @@
...
 import 'package:test/test.dart';
 
 /// Tests that the [Client] correctly implements HTTP redirect logic.
 void testRedirect(Client client) async {
   group('redirects', () {
-    late HttpServer server;
+    late String host;
+    late StreamChannel<Object?> httpServerChannel;
+
     setUp(() async {
```
natebosch, line 15: Should we consider using `setUpAll`? The server looks stateless so we shouldn't lose much safety, but we can save some time by using a single isolate and server.

pkgs/http_client_conformance_tests/lib/src/redirect_server.dart
```diff
@@ -0,0 +1,41 @@
+import 'dart:async';
```
natebosch, line 1: [nit] missing copyright block

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -76,24 +56,23 @@ void testRedirect(Client client) async {
...
-      final request = Request('GET', Uri.http('localhost:${server.port}', '/6'))
+      final request = Request('GET', Uri.http(host, '/6'))
         ..followRedirects = true
         ..maxRedirects = 5;
       expect(
           client.send(request),
           throwsA(isA<ClientException>()
               .having((e) => e.message, 'message', 'Redirect limit exceeded')));
-    });
+    }, onPlatform: {'browser': const Skip('not passing')});
```
natebosch, line 66: [nit] can move the const up a level
```suggestion
    }, onPlatform: const {'browser': Skip('not passing')});
```

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -76,24 +56,23 @@ void testRedirect(Client client) async {
...
-            ..maxRedirects = 5;
+      final request = Request('GET', Uri.http(host, '/loop'))
+        ..followRedirects = true
+        ..maxRedirects = 5;
       expect(
           client.send(request),
           throwsA(isA<ClientException>().having((e) => e.message, 'message',
               isIn(['Redirect loop detected', 'Redirect limit exceeded']))));
-    });
+    }, onPlatform: {'browser': const Skip('not passing')});
```
natebosch, line 76:
```suggestion
    }, onPlatform: const {'browser': Skip('not passing')});
```

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -2,60 +2,40 @@
...
 
     test('disallow redirect', () async {
-      final request = Request('GET', Uri.http('localhost:${server.port}', '/1'))
+      final request = Request('GET', Uri.http(host, '/1'))
         ..followRedirects = false;
       final response = await client.send(request);
       expect(response.statusCode, 302);
       expect(response.isRedirect, true);
-    });
+    }, onPlatform: {'browser': const Skip('not passing')});
```
natebosch, line 27:
```suggestion
    }, onPlatform: const {'browser': Skip('not passing')});
```

natebosch, line 27: Should we consider a boolean argument expressing this limitation? The browser client doesn't honor `followRedirect` or `maxRedirects` at all, so we could consider skipping any test which relies on those with `canControlRedirects: false`?

natebosch (approved):

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: I'm not sure. If we are trying to catch issues that **could** be fixed then we want the skip here - because you could implement BrowserClient in a way that would fix this.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: Done

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: Done.

pkgs/http_client_conformance_tests/lib/src/redirect_server.dart
brianquinlan: Done.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: The server is stateless but the guidance for `setUpAll` is to avoid it unless `setUp` is prohibitively slow (which it isn't). 

But I don't have a strong opinion here.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: I'm convinced - fixed!

PR #711, opened 2022-06-02
Run the request body tests against the browser client.

brianquinlan:


pkgs/http_client_conformance_tests/lib/src/request_body_streamed_server.dart
```diff
@@ -0,0 +1,42 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:stream_channel/stream_channel.dart';
+
+/// Starts an HTTP server that captures the content type header and request
```
natebosch, line 11: This header does not match

pkgs/http_client_conformance_tests/lib/src/request_body_tests.dart
```diff
@@ -35,189 +35,121 @@ class _Plus2Encoding extends Encoding {
...
 
+      final serverReceivedContentType = await httpServerQueue.next;
+      final serverReceivedBody = await httpServerQueue.next as String;
+
       expect(serverReceivedContentType, ['image/png; charset=plus2']);
       expect(serverReceivedBody.codeUnits, [1, 2, 3, 4, 5]);
-      await server.close();
     });
-
+/*
```
natebosch, line 152: Delete this instead of commenting it out. We can find it in source history if we need it again.

natebosch (approved):

pkgs/http_client_conformance_tests/lib/src/request_body_streamed_server.dart
brianquinlan: Done.

pkgs/http_client_conformance_tests/lib/src/request_body_tests.dart
brianquinlan: I didn't mean to leave that - it was just sloppiness of my part.

PR #718, opened 2022-06-13
Move conformance test invocation into package:http

brianquinlan:


pkgs/http_client_conformance_tests/lib/http_client_conformance_tests.dart
```diff
@@ -22,6 +22,12 @@ export 'src/response_headers_tests.dart' show testResponseHeaders;
 
 /// Runs the entire test suite against the given [Client].
 ///
+/// If [packageRoot] is set then it will be used as the filesystem root
```
natebosch, line 25: Do we need a publish of `package:test`? Would that simplify this?

natebosch: Could we pass the `package:` URI without resolving it since you and @jakemac53 fixed the handling of `package:`?

brianquinlan: We could but that would only change one line in the `startServer` code.

brianquinlan: I don't know where I went wrong before. I merged in your change. Thanks and sorry for messing this up (somehow).

natebosch (approved):

PR #719, opened 2022-06-14
Fixes #701 IOClient should always throw ClientException

### LINKED ISSUE #701: The dart:io Client should always throw ClientException
In some cases it throws `SocketException` e.g. when the [host name is invalid](https://github.com/dart-lang/http/blob/87d43798ae0fe7264ecda9d0dbc3e491a2abab65/test/io/client_test.dart#L104)

The browser client always throws `ClientException`.

The dart:io Client should always throw ClientException so that exceptions can be handled consistently between clients.

There is significant code that catches `SocketException` so this would be a breaking change. One possible way around this would be to have a custom exception like:

```dart
class SocketClientException extends ClientException implements SocketException {

}
```

brianquinlan:


brianquinlan:
Hey @natebosch ,  it doesn't make sense to merge this until https://github.com/dart-lang/http/pull/718 lands but what do you think about the approach in general?

natebosch:
cc @lrhn for opinions

I think this would probably be an OK compromise. I worry that it's a detail which is a little strange but we'll likely have to live with it forever - a better outcome IMO would be to clean up uses to not need `SocketException`, but that's probably not feasible.

brianquinlan:
> I think this would probably be an OK compromise. I worry that it's a detail which is a little strange but we'll likely have to live with it forever - a better outcome IMO would be to clean up uses to not need `SocketException`, but that's probably not feasible.

I can try to fix all of the uses in Google but outside of Google it will be hard to even know that a migration is necessary.

lrhn:
I think this is a good trade-off, and a reasonable change.
Exceptions should be documented and handled at API boundaries. It's OK to let an exception from a lower level escape, but you must document it in your own API as well, so the users know that it is throw.
Since there is no documentation changes in this CL, I assume the `SocketException` wasn't documented, and the `ClientException` was. In that case, this is a good change!

Being backwards compatible without documenting it is tricky, but acceptable. People may end up depending on undocumented behavior. On the other hand, new code is more likely to just use the documented approach, and old code will just still work.

If we ever get a way to document exceptions in a way that the analyzer can enforce, we can tell people catching `SocketException` that we don't actually throw those, and they should be catching `ClientException`s instead.

I think this makes it more likely that people will do `catch ClientException` instead of just `catch Exception`, which is a significant advantage.

So, OK with me.

brianquinlan:
> Since there is no documentation changes in this CL, I assume the `SocketException` wasn't documented, and the `ClientException` was.

That's correct.

pkgs/http_client_conformance_tests/lib/src/server_errors_test.dart
```diff
@@ -0,0 +1,41 @@
...
+    late final StreamQueue<Object?> httpServerQueue;
+
+    setUpAll(() async {
+      httpServerChannel = await startServer('server_errors_server.dart');
+      httpServerQueue = StreamQueue(httpServerChannel.stream);
+      host = 'localhost:${await httpServerQueue.next}';
+    });
+    tearDownAll(() => httpServerChannel.sink.add(null));
+
+    test('not such host', () async {
```
natebosch, line 27: [nit]
```suggestion
    test('no such host', () async {
```

natebosch (approved):

PR #724, opened 2022-06-23
Cupertino flutter real

brianquinlan:


pkgs/cupertino_http/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   meta: ^1.7.0
   native_library:
     path: ./third_party/native_library
+  plugin_platform_interface: ^2.0.2
```
stuartmorgan, line 20: What is this for? I don't see any Dart code that would use this.

pkgs/cupertino_http/example/README.md
```diff
@@ -0,0 +1,16 @@
+# cupertino_http_example
+
+Demonstrates how to use the cupertino_http plugin.
+
+## Getting Started
```
stuartmorgan, line 5: You probably don't want this boilerplate section.

pkgs/cupertino_http/src/CUPHTTPClientDelegate.m
```diff
@@ -42,7 +42,7 @@ @implementation CUPHTTPClientDelegate {
 - (instancetype)init {
   self = [super init];
   if (self != nil) {
-    taskConfigurations = [NSMapTable strongToStrongObjectsMapTable];
+   taskConfigurations = [[NSMapTable strongToStrongObjectsMapTable] retain];
```
stuartmorgan, line 45: Why are you building this code without ARC?

pkgs/cupertino_http/ios/cupertino_http.podspec
```diff
@@ -0,0 +1,29 @@
...
+Pod::Spec.new do |s|
+  s.name             = 'cupertino_http'
+  s.version          = '0.0.1'
+  s.summary          = 'Flutter Foundation URL Loading System'
+  s.description      = <<-DESC
+  A Flutter plugin for accessing the Foundation URL Loading System.
+                       DESC
+  s.homepage         = 'https://github.com/dart-lang/http/tree/master/pkgs/cupertino_http'
+  s.license          = { :type => 'BSD', :file => '../LICENSE' }
+  s.author           = { 'Flutter Dev Team' => 'flutter-dev@googlegroups.com' }
```
stuartmorgan, line 14: This is probably not the author you want.

stuartmorgan (commented):
Is it intentional that there's no README

pkgs/cupertino_http/ios/cupertino_http.podspec
brianquinlan: I set it to a placeholder for now.

pkgs/cupertino_http/src/CUPHTTPClientDelegate.m
stuartmorgan: Turning off ARC seems like a very large and dangerous hammer to solve that problem. Why not just have a collection where you keep objects whose lifetime you want to guarantee? That's the standard way to manage lifetimes in modern ObjC code.

brianquinlan: That's a great idea! I made a note to do it that way.

brianquinlan:
> Is it intentional that there's no README

The README is already in place:
https://github.com/dart-lang/http/blob/cupertino/pkgs/cupertino_http/README.md

pkgs/cupertino_http/example/.gitignore
```diff
@@ -0,0 +1,47 @@
+# Miscellaneous
+*.class
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
```
natebosch, line 10: Is there some flutter template that sticks this junk in here?

If we need to keep this file, can we delete through line 23?

pkgs/cupertino_http/example/ios/Runner/Assets.xcassets/LaunchImage.imageset/README.md
```diff
@@ -0,0 +1,5 @@
+# Launch Screen Assets
```
natebosch, line 1: Remove this file?

Remove this directory?

pkgs/cupertino_http/example/macos/Runner/Configs/AppInfo.xcconfig
```diff
@@ -0,0 +1,14 @@
...
+// 'flutter create' template.
+
+// The application's name. By default this is also the title of the Flutter window.
+PRODUCT_NAME = cupertino_http_example
+
+// The application's bundle identifier
+PRODUCT_BUNDLE_IDENTIFIER = com.example.cupertinoHttpExample
+
+// The copyright displayed in application information
+PRODUCT_COPYRIGHT = Copyright © 2022 com.example. All rights reserved.
```
natebosch, line 14: does this need a change?

pkgs/cupertino_http/ios/cupertino_http.podspec
```diff
@@ -0,0 +1,29 @@
...
+Pod::Spec.new do |s|
+  s.name             = 'cupertino_http'
+  s.version          = '0.0.1'
+  s.summary          = 'Flutter Foundation URL Loading System'
+  s.description      = <<-DESC
+  A Flutter plugin for accessing the Foundation URL Loading System.
+                       DESC
+  s.homepage         = 'https://github.com/dart-lang/http/tree/master/pkgs/cupertino_http'
+  s.license          = { :type => 'BSD', :file => '../LICENSE' }
+  s.author           = { 'TODO' => 'use-valid-author' }
```
natebosch, line 14: When will we fill in the author?

pkgs/cupertino_http/macos/cupertino_http.podspec
```diff
@@ -0,0 +1,28 @@
...
+Pod::Spec.new do |s|
+  s.name             = 'cupertino_http'
+  s.version          = '0.0.1'
+  s.summary          = 'Flutter Foundation URL Loading System'
+  s.description      = <<-DESC
+  A Flutter plugin for accessing the Foundation URL Loading System.
+                       DESC
+  s.homepage         = 'https://github.com/dart-lang/http/tree/master/pkgs/cupertino_http'
+  s.license          = { :type => 'BSD', :file => '../LICENSE' }
+  s.author           = { 'TODO' => 'use-valid-author' }
```
natebosch, line 14: author?

pkgs/cupertino_http/example/ios/Runner/Assets.xcassets/LaunchImage.imageset/README.md
brianquinlan: This generated by the plugin template. Our other plugins include this as well e.g.
https://github.com/flutter/plugins/tree/main/packages/camera/camera_avfoundation/example/ios/Runner/Assets.xcassets/LaunchImage.imageset

natebosch: I'm not sure why we aren't cleaning up template cruft in other packages. I think we should always.

stuartmorgan: > I'm not sure why we aren't cleaning up template cruft in other packages.

Just because cleaning cruft out of parts of example apps that probably nobody will ever look at is not a priority for us in the  ~90 {example x platform} sparse martrix we maintain in that repository alone. Remembering which parts of the templates—across six different platforms—would need to be adjusted, and how, every time we add or re-generate an example, would take a lot of reviewer energy that we would rather spend elsewhere.

> I think we should always.

PRs for the flutter/plugins+flutter/packages tooling that encode and enforce all the things that need to be adjusted would be welcome, since it would automate the process.

pkgs/cupertino_http/example/macos/Runner/Configs/AppInfo.xcconfig
brianquinlan: Done. And I checked for other copyright notices and couldn't find any.

pkgs/cupertino_http/macos/cupertino_http.podspec
brianquinlan: I'll fix this when Kevin gives me a group name.

natebosch (approved):

PR #728, opened 2022-07-01
added onUploadProgress 

Mehmetyaz:
added onUploadProgress to `MultipartRequest.new` and `Client.send`. (adjusted file movements)

### DIFF
--- a/pkgs/http/README.md
+++ b/pkgs/http/README.md
@@ -255,7 +255,9 @@ In Flutter, you can use a one of many
 If you depend on code that uses top-level functions (e.g. `http.post`) or
 calls the [`Client()`][clientconstructor] constructor, then you can use
 [`runWithClient`](runwithclient) to ensure that the correct
-`Client` is used.
+`Client` is used. When an [Isolate][isolate] is spawned, it does not inherit
+any variables from the calling Zone, so `runWithClient` needs to be used in
+each Isolate that uses `package:http`.
 
 You can ensure that only the `Client` that you have explicitly configured is
 used by defining `no_default_http_client=true` in the environment. This will
@@ -283,6 +285,7 @@ $ dart compile exe --define=no_default_http_client=true ...
 [fetchclient]: https://pub.dev/documentation/fetch_client/latest/fetch_client/FetchClient-class.html
 [flutterhttpexample]: https://github.com/dart-lang/http/tree/master/pkgs/flutter_http_example
 [ioclient]: https://pub.dev/documentation/http/latest/io_client/IOClient-class.html
+[isolate]: https://dart.dev/language/concurrency#how-isolates-work
 [flutterstatemanagement]: https://docs.flutter.dev/data-and-backend/state-mgmt/options
 [provider]: https://pub.dev/packages/provider
 [runwithclient]: https://pub.dev/documentation/http/latest/http/runWithClient.html


Mehmetyaz:
@kevmoo - More than 80 changes appeared in the previous frok. With the new fork, your inspection will be easier.

.gitignore
```diff
@@ -2,3 +2,4 @@
 .dart_tool
 .packages
 pubspec.lock
+.idea
```
kevmoo, line 5: Please remove this line! It should be in your `~/.gitignore_global`

Mehmetyaz: Removed!

pkgs/http/lib/src/client.dart
```diff
@@ -139,7 +140,11 @@ abstract class Client {
   Future<Uint8List> readBytes(Uri url, {Map<String, String>? headers});
 
   /// Sends an HTTP request and asynchronously returns the response.
-  Future<StreamedResponse> send(BaseRequest request);
+  ///
+  /// If [onUploadProgress] defined, the callback will be called when the
+  /// upload progress changes.
+  Future<StreamedResponse> send(BaseRequest request,
```
brianquinlan, line 146: Some drive-by comments:
1. this change is clearly breaking (and would break existing code in practice) and should be paired with a significant version bump
2. IIUC, this is fairly easy functionality for the user to replicate
3. This functionality seems generally useful

Maybe there is another way of providing this functionality? Like set the callback in the constructor? Or as a property?

Mehmetyaz: I just needed to add the callback as argument in MultipartRequest's constructor.  But it will be handled by implementations of `Client.send`.  If you meant the `Client` constructor, I think it's a bad idea.  Because OnUploadProgress's context is send, not `Client`.

Mehmetyaz: Also, adding this feature will only give an error feom old codes if the developer has written developer's own `Client` implementation.

In external cases, there will be no errors or conflicts  because it is a nullable named argument.

Mehmetyaz: I understand. 

I think we can resolve these conflicts with a nullable property to add to `BaseRequest`.  `send` is already take a `BaseRequest`.  I didn't get a chance to try with coding.  I will try as soon as possible.

I think the property you mentioned was for `BaseRequest`.  I thought it wouldn't be a solution because I thought you said it for the `Client`.

Mehmetyaz: I remove the argument from `Client.send`. It's a really useless argument. Because the only place that works `send` with `onUploadProgress` is `BaseRequest` and `send` already gets `BaseRequest`.

brianquinlan:
@natebosch , could you take a look at this?

brianquinlan:
FYI, @natebosch is on holiday until late next week so don't expect feedback until then.

pkgs/http/lib/src/base_request.dart
```diff
@@ -89,14 +89,33 @@ abstract class BaseRequest {
...
     }
     return method;
   }
 
-  BaseRequest(String method, this.url)
+  /// On upload progress callback.
+  ///
+  /// If defined, this callback will be called when the upload progress changes.
+  ///
+  /// In browser, uses XMLHttpRequest's "xhr.upload.onLoad" event.
```
natebosch, line 104: Should we send more than a single event using `progress`?

pkgs/http/lib/src/base_request.dart
```diff
@@ -89,14 +89,33 @@ abstract class BaseRequest {
...
+  ///
+  /// In IO, uses the yield length of the stream. The total length of the bytes
+  /// yielded by the stream at any given moment is "uploaded" and the total
```
natebosch, line 107: @brianquinlan, @lrhn - how meaningful do we expect this to be? Is watching a transform step in a stream being read by `dart:io` HTTP handling an accurate way to understand upload progress?

Mehmetyaz: Not exactly.  While I was writing the altogic_dart package, I duplicated it.  Meanwhile, I split all yields larger than the specified chunk size according to this size.  But adding this to the official package would have caused more confusion. Currently MultipartRequest splits into specific chunks and progress works correctly as the part is actually.  But it needs to be divided into more accurate parts for big files. 

E.g current results like this for big files : 0.1- 1-99.3-100

If large yields are divided, loading can be done with closer ratios.

pkgs/http/lib/src/base_request.dart
Mehmetyaz: I didn't understand exactly what you mean.  I am sorry.

pkgs/http/lib/src/base_request.dart
Mehmetyaz: There is two methods for this. 

1) Like browser client, read stream until done, split by defined chunk size. This may cause more memory consumption. 

2) Split big parts by chunk size in the ByteStream. But this splitting will be done  for browser also and it unnecessary.  Because browser client works different. 

If there is a solution you approve, I can code it.  I also need to find out if the chuck size will be a static value or a parameter.  Parameters can cause confusion.  if it is static how much should it be?

pkgs/http/lib/src/base_request.dart
lrhn, line 104: More words may make this easier to read (to a limit, as always).

> /// When run in a browser, this callback relies on the
> /// browser's `XMLHttpRequest`'s `xhr.upload.onLoad` events.
> ///
> /// When run natively, the number of bytes is 
> /// .... 

I don't understand the description of the native version.
Probably because I don't know what stream we are talking about, and how you can possibly know the total length of a stream before it's done.

pkgs/http/lib/src/base_request.dart
```diff
@@ -89,14 +89,33 @@ abstract class BaseRequest {
...
+  /// length of the stream is "total"
+  ///
+  /// If [onUploadProgress] callback is provided and length is computable,
+  /// [onUploadProgress] will execute for each chunk was sent.
+  ///
+  /// lengthComputable :
+  ///     library.html : xhr.lengthComputable
+  ///     library.io : content-length is provided (MultipartRequest provide)
+  final OnUploadProgress? onUploadProgress;
```
lrhn, line 116: Don't use a typedef here, just use the function type.

pkgs/http/lib/src/base_request.dart
lrhn: Or make `total` nullable and don't provide it when you don't have one.

If the `loaded` is defined as being bytes, at least when total is `null`, then the user can still be shown a meaningful progress, it'll just be "49 MB" instead of "37%".


I'd never read the entire stream and then chunk it afterwards. That's just useless overhead. If everything comes in one big chunk, that's just what you have to deal with.

pkgs/http/lib/src/io_client.dart
```diff
@@ -53,17 +56,33 @@ class IOClient extends BaseClient {
...
+    ByteStream? handledStream;
+
+    var contentLength = request.contentLength;
+    if (request.onUploadProgress != null && contentLength != null) {
+      var load = 0;
+      handledStream =
+          ByteStream(stream.transform(StreamTransformer.fromBind((d) async* {
+        await for (var data in d) {
+          load += data.length;
+          request.onUploadProgress!.call(contentLength, load);
```
lrhn, line 68: No need for `.call` here, just call the function directly.

pkgs/http/lib/src/utils.dart
```diff
@@ -74,3 +74,8 @@ Stream<T> onDone<T>(Stream<T> stream, void Function() onDone) =>
       sink.close();
       onDone();
     }));
+
+/// On upload progress callback.
+///
+/// See : "MultipartRequest"
+typedef OnUploadProgress = void Function(int uploaded, int total);
```
lrhn, line 81: I'd drop the typedef and just inline the function type.

No need to pollute the public, exported, API with an extra name.

pkgs/http/lib/src/utils.dart
```diff
@@ -74,3 +74,8 @@ Stream<T> onDone<T>(Stream<T> stream, void Function() onDone) =>
       sink.close();
       onDone();
     }));
+
+/// On upload progress callback.
```
lrhn, line 78: (Would drop the "On" here, just "Upload progress callback", it's cleaner. The upload progress callback is called on upload progress.)

pkgs/http/lib/src/base_request.dart
natebosch: The `xhr.upload.onLoad` event will fire exactly once. This will not show incremental progress.

The `xhr.upload.progress` event may fire multiple times.

Mehmetyaz: Done

Mehmetyaz: `MultipartRequest` calculates the length value before the stream starts. If the length value is a non-calculated request (which I just added to the MultipartRequest), `onUploadProgress` will never be triggered.

Documentation will be extended

pkgs/http/lib/src/base_request.dart
Mehmetyaz: Done

pkgs/http/lib/src/io_client.dart
Mehmetyaz: Done

pkgs/http/lib/src/utils.dart
Mehmetyaz: Done

pkgs/http/lib/src/utils.dart
Mehmetyaz: typedef removed

pkgs/http/lib/src/base_request.dart
Mehmetyaz: Making `total` nullable does not fix the problem. Instead, the `ByteStream .fromBytes` consturctor simply splits the List<int>(bytes) into parts.

If MultipartFile is created with a stream, meaningful data is now obtained. But when created with fromBytes it pipes it as one piece. This prevents the rates from becoming meaningful. 

Also at the moment onUploadProgress is not called with a rate anyway. There are loaded and total. Since the callback is triggered only where the total is calculated, the developer can calculate the rate.

pkgs/http/lib/src/base_request.dart
```diff
@@ -89,14 +89,33 @@ abstract class BaseRequest {
...
+  /// yielded by the stream at any given moment is "uploaded" and the total
+  /// length of the stream is "total"
+  ///
+  /// If [onUploadProgress] callback is provided and length is computable,
+  /// [onUploadProgress] will execute for each chunk was sent.
+  ///
+  /// lengthComputable :
+  ///     library.html : xhr.lengthComputable
+  ///     library.io : content-length is provided (MultipartRequest provide)
+  final void Function(int uploaded, int total)? onUploadProgress;
```
AriasBros, line 116: Hi @Mehmetyaz, thanks for this PR, it is very useful! We are using your fork in our app overriding the dependency of the `http` package until this get deployed in a new version of the package.

Said this, we noted that this signature could be wrong and has the parameters in the wrong order.
In the rest of the code you are calling the `onUploadProgress` callback using the `total` as the first parameter and the `uploaded` bytes as the second one.

Just that. Thanks a lot again.

mehcode:
Is there anything needed here before this is merged? I'd really love to see this merged and support added to cupertino_http.

kevmoo:
@brianquinlan @lrhn ?

natebosch:
I'd still like to understand how meaningful these numbers are.
Moving the http implementation to the `progress` event helps - it feels like we're adding specific value in that case. If we feel like the stream watching is high enough fidelity for the IO case we can support both through the same interface. Are we confident in that fidelity @brianquinlan @lrhn ?
Is there any buffering in between this Stream that we can account for better with an API in the SDK? Will watching the Stream consumption be meaningful for the cronet and cupertino implementations?


If what we're offering is implemented as a callback that gets invoked as the Stream is consumed - does that need to be shipped as part of `package:http`?

pkgs/http/lib/src/base_request.dart
```diff
@@ -89,14 +89,33 @@ abstract class BaseRequest {
...
+
   static String _validateMethod(String method) {
     if (!_tokenRE.hasMatch(method)) {
       throw ArgumentError.value(method, 'method', 'Not a valid method');
     }
     return method;
   }
 
-  BaseRequest(String method, this.url)
+  /// On upload progress callback.
```
lrhn, line 100: Could we provide a (broadcast) stream instead, instead of putting the callback directly into the class.

That's the traditional way to provide events, and the `on` name would be traditional for such a stream.
```dart
/// Upload progress notification
///
/// Sends events when upload progress information is
/// available.
///
/// ....
Stream<Progress> get onUploadProgress;
```
with a 
```dart
abstract class Progress {
  int progress;
  int? get total;
}
```
(Which I'd make a record when we get records, and document as something users *must not* implement or extend.)

Mehmetyaz: If you agree, I can commit to implement this solution.

Mehmetyaz: Unfortunately I found that this is not a viable solution. What will I return in the getter you defined to BaseRequest? `StreamController.stream`. 

A StreamController is required and if the developer does not want to listen to this progress, it creates an unnecessary load. In addition, since I need to know if the user has decided to listen for progress and this controller must be nullable.

If I could access the private values of `BaseRequest` from within the client (e.g `IOClient`), maybe I could come up with suitable solutions. Otherwise, two more properties are required, such as `BaseRequest.onUploadProgress` and `BaseRequest.uploadProgressController?`. I can guess that you don't want this visible to the developer.

lrhn: You can always create the controller when requested, and store it in a private variable:
```dart
  StreamController<Progress>? _controller;
  Stream<Progress> get onProgress => (_controller ??= StreamController(onCancel: () {
    _controller = null;
  })).stream;

  void _addProgress(int bytes, int? total) => _controller?.add(Progress(bytes, total));
```

This code has very little overhead when it's not used (a single `null` field and a single `null` check of that field whenever there is progress).
You can check whether `_controller` is `null` anywhere you currently check whether `onProgressCallback` is set.
 
If you need to access the internals inside `IOClient`, then you should just make the `onProgress` getter abstract and implement it in all the concrete subclasses. (I admit that the lack of `protected` access is annoying.)

Or add a 
```dart
void addProgressToRequest(Progress progress, BaseRequest request) {
  request._addProgress(progress);
}
```
to `base_progress.dart`, use it from the other libraries, but don't export it as part of the public interface of the package.
(That's a traditional way to share privates between libraries in the same package.)

brianquinlan: AFAIK, we are waiting on a `package:http` major version bump to review changes like this.

Is that true @natebosch ?

natebosch: > AFAIK, we are waiting on a `package:http` major version bump to review changes like this.

I haven't evaluated how difficult it will be to roll this internally. I do think it would need to wait for a major version bump.

I'm still a little unsure how meaningful these numbers are, or why it's important that we provide this API in this package instead of having the caller watch the stream progress.

Mehmetyaz: You are indeed right.  In my implementation for Altogic, onUploadProgress was useful because I split the bytes into chunks even if bytes weren't streamed(e.g. multipartfile.fromBytes). 

If it is not a problem for you, a solution is also possible, such as automatically increasing the number of stream parts in certain situations.

pkgs/http/lib/src/base_request.dart
brianquinlan: > @brianquinlan, @lrhn - how meaningful do we expect this to be? Is watching a transform step in a stream being read by `dart:io` HTTP handling an accurate way to understand upload progress?

Probably not. When you `write` to a dart socket it doesn't mean that the data has actually left the local machine (or even been copied to OS memory).

brianquinlan: Of the 6 implementations that I know about, you could get good numbers for:
- browser
- cupertino_http

and bad numbers for:
- fetch_client (though this should improve with time as the JavaScript streaming API improves)
- cronet_http
- io
- java_http (not sure about this, based on Java HttpURLConnection)

pkgs/http/lib/src/base_request.dart
brianquinlan: @lrhn I actually prefer the original API (using a callback rather than a `Stream`) because it means that the `Client` implementation can easily check whether the caller is interesting in progress notification one at the beginning of their send implementation. In `package:cupertino_http`, I'd probably implement something like:

```dart
  @override
    ...
    if (request.onUploadProgress != null) {
      // This `Timer` is an unnecessary expense if no one is listening for notifications.
      final timer = Timer.periodic(const Duration(milliseconds: 100), (timer) { 
      if (taskTracker.isDone) {
        timer.cancel();
      } else {
        progress = Progress(task.countOfBytesExpectedToSend,  task.countOfBytesSent);
        request.onUploadProcess(progress);
      });
    }
```

If the `Stream/StreamController` approach were used, it seems like the `StreamController` would have to be public and so the `send` implementation can do something like:

```
  @override
    void setupTimer() {
    
    }

    if (request.controller.isListening) {
      setupTimer();
    } else {
      request.onListen = setupTimer;
   }
```

brianquinlan:
I patched this change locally and did some testing.

For `IOClient`, the progress seemed useful for large requests (assuming that the stream is split into relatively small lists). On my Mac, ~500KiB of progress would be indicated before the first byte arrives on the server. So presumably that is approximately when the TCP write buffer is full [1].

For `BrowserClient`, I was not able go get any progress messages when running on Chrome - @Mehmetyaz how did you test this?

[1] I think that this can be verified by checking [`availableSendBuffer`](https://developer.apple.com/documentation/network/nwprotocoltcp/metadata/2998771-availablesendbuffer)

Mehmetyaz:
Moved to #1071

PR #731, opened 2022-07-12
Fix a bug where strongToStrongObjectsMapTable was not retained 

brianquinlan:


aam (approved):
how do you find this kind of bugs?

brianquinlan:
> how do you find this kind of bugs?

I was running this in Flutter and get an exception from the objective-C runtime saying that I was trying to release an already released object.

I should probably figure out a way to force GC during the unit tests so I can catch these sorts of issues there.

PR #732, opened 2022-07-13
Add the ability to download data into a file.

brianquinlan:


pkgs/cupertino_http/lib/cupertino_http.dart
```diff
@@ -697,6 +712,28 @@ void _setupDelegation(
...
         }
         break;
+      case ncb.MessageType.FinishedDownloading:
+        final finishedDownloading =
+            ncb.CUPHTTPForwardedFinishedDownloading.castFrom(forwardedDelegate);
+        try {
+          if (onFinishedDownloading == null) {
+            break;
+          }
+          try {
```
aam, line 722: What is the meaning of outer try-finally? it doesn't look like `if (onFinishedDownloading == null)` could throw.

aam (approved):

pkgs/cupertino_http/lib/cupertino_http.dart
brianquinlan: It can't. I just want to ensure that `finishedDownloading.finish()` is called because, if it isn't, then `cupertino_http` will deadlock.

pkgs/cupertino_http/lib/cupertino_http.dart
```diff
@@ -875,4 +925,23 @@ class URLSession extends _ObjectHolder<ncb.NSURLSession> {
 
     return URLSessionTask._(task._nsObject);
   }
+
+  /// Creates a [URLSessionDownloadTask] that downloads the data from a server
+  /// URL.
+  ///
+  /// Provide a `onFinishedDownloading` handler in the [URLSession] factory to
+  /// received notifications when the data has completed downloaded.
```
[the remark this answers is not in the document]
brianquinlan: Fixed.

PR #737, opened 2022-07-24
Add support for flutter plugins

brianquinlan:


pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -20,7 +21,8 @@ void testRedirect(Client client, {bool redirectAlwaysAllowed = false}) async {
     late final StreamQueue<Object?> httpServerQueue;
 
     setUpAll(() async {
-      httpServerChannel = await startServer('redirect_server.dart');
+      httpServerChannel =
+          await startServer('redirect_server.dart', server.hybridMain);
```
aam, line 25: It feels unfortunate how references to target file (redirect_server.dart) are duplicated as a runtime parameter to a startServer function as well as static import as a target for `server.hybridMain` target function.
Why generic `spawnHybridUri` approach doesn't work for flutter plugin?

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
```diff
@@ -7,6 +7,7 @@ import 'package:http/http.dart';
 import 'package:stream_channel/stream_channel.dart';
 import 'package:test/test.dart';
 
+import 'redirect_server.dart' as server;
```
natebosch, line 10: I would expect these imports to cause problems on the web.

brianquinlan: I would expect that too - but it seems to work in practice - the CI passes and running `dart test -p chrome` on my workstation also works.

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
brianquinlan: The problem with flutter plugins is that the integration tests are run either on a device or on the emulator. But spawnHybridUri will start the server on the computer running the test and the two might not share the same `localhost`. So my change runs the server on the device itself, except in the web case where running the server in the browser is not possible.

If we don't like this approach, it might not be possible to have special logic for the android emulator where we hardcode the address of the host device (it's always `10.0.2.2` [1]). The iOS emulator uses the same network as the host. And we could not support running the tests on read devices.

[1] https://developer.android.com/studio/run/emulator-networking

pkgs/http_client_conformance_tests/lib/src/redirect_tests.dart
natebosch: We might have broken our safeguards against importing `dart:io` when compiling through the test runner. I best we haven't broken them in google3 though. This is not going to work internally.

natebosch: I filed https://github.com/dart-lang/test/issues/1748 to make it easier to catch these eagerly in the test runner.

brianquinlan: Do I need to care if this works in google3? It seems like it is useful that this actually works.

But, if we want to change it, these approaches come to mind:
1. Try to always use `spawnHybridUri` and try to deal with emulators as a special chase (as I wrote @aam above) - the downside is 
2. Only use `spawnHybridUri` for the web and solve the problem using conditional imports (I think that I'd need 3 files per server) like:

`redirect_server.dart` - existing file
`redirect_server_web.dart` - calls `spawnHybridUri`
`redirect_server_vm.dart` - calls `hybridMain` directly

`redirect_test.dart` would then do:

```dart
import 'redirect_server_vm.dart'
    if (dart.library.html) 'redirect_server_web.dart';
```

3: Remove browser support for the tests.

I'd go with (2) - it is verbose but it should work. Any thoughts?

natebosch: > Do I need to care if this works in google3?

Yes. We want to have `package:http` tests running in google3.

> 2\. Only use `spawnHybridUri` for the web and solve the problem using conditional imports (I think that I'd need 3 files per server)

I think this is the best approach. We could probably write a little script in Dart or bash to create a `_web.dart` and `_vm.dart` file for each `_server.dart` file in a directory.

@jakemac53 - can you think of any other patterns we could use?

jakemac53: conditional imports is going to be the best way I think (and will work internally, with some special config)

brianquinlan: Done.

natebosch (approved):

brianquinlan:
@aam Any more feedback here?

pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart
```diff
@@ -0,0 +1,46 @@
...
+
+const web = '''// Generated by generate_server_wrappers.dart. Do not edit.
+
+import 'package:stream_channel/stream_channel.dart';
+import 'package:test/test.dart';
+
+/// Starts the redirect test HTTP server out-of-process.
+Future<StreamChannel<Object?>> startServer() async => spawnHybridUri(Uri(
+    scheme: 'package',
+    path: 'http_client_conformance_tests/src/<server_file_placeholder>'));
```
aam, line 32: Can something be done to make sure this generated line is not too long, won't require reformatting before checking-in?
Or even better solution would be not to check them in at all and use https://pub.dev/packages/source_gen instead to generate them as part of the build.

natebosch: Not all packages that consume this will be using the build system already. I don't think we should impose it for now.

> won't require reformatting before checking-in?

We can add a `dart format` step in this script if this situation comes up. I don't think we need to be too particular about this - it isn't a process we're publishing for others to adopt, just a quick hack to make this pattern painless locally.

brianquinlan: I'm happy to use Dart format now... it seems trivial to do. I'll report back when done.

brianquinlan: Done.

aam (approved):

PR #745, opened 2022-08-03
Upgrade to the latest package:lints

natebosch:


pkgs/http/pubspec.yaml
```diff
@@ -16,7 +16,7 @@ dev_dependencies:
   fake_async: ^1.2.0
   http_client_conformance_tests:
     path: ../http_client_conformance_tests/
-  lints: ^1.0.0
+  lints: ^2.0.0
```
devoncarew, line 19: If this have trouble passing as is (since we currently provisioning `2.14.0` on the bots), you could try `>=1.0.0 <3.0.0`. That's what I've done in other situations; locally you'll see 2.0.0, but the older sdk bots can still run.

kevmoo: or bump the min SDK to 2.17?

devoncarew (approved):

PR #755, opened 2022-08-12
Add note about state in README

mit-mit:


pkgs/cupertino_http/README.md
```diff
@@ -4,6 +4,14 @@
 A macOS/iOS Flutter plugin that provides access to the
 [Foundation URL Loading System][].
 
+## Experiment warning
+
+**NOTE**: This package is experimental, and shared for feedback.
+Breaking changes may be made without announcement.
+The whole package may get discontinued.
```
brianquinlan, line 11: Is this true? At what point do we make a commitment to a package?

devoncarew: It's true for packages in the labs.dart.dev publisher. We either graduate them to a more supported publisher or discontinue them.

My guess is that you'd likely decide on the timeframe for when to officially support the package (we're happy with the current API, and this is the direction we want to go).

brianquinlan: I'm a bit concerned that this wording understates our commitment and will make it hard to get traction. 

Looking at other labs.dart.dev packages, only one active project has this disclaimer (`dart_internal`).

Others (i.e. `wasm`, `gcloud`, `appengine`,  `grpc_cronet`) do not.

mit-mit: That's simply because I haven't gotten to them yet.

It would be good for us to agree on standard verbiage for all labs packages, and then update all the READMEs.

brianquinlan: @a-siva Do we have a higher level of commitment than "The whole package may get discontinued."

a-siva: At least for the http plugin packages I would like to avoid adding the phrase "The whole package may get discontinued", we do not plan on discontinuing the package, my hope is these packages would graduate into the officially supported category once we feel the API has accounted for all the feedback we get.

mit-mit: I'd like to clarify what my statement is not specific to this package, but general to all of our labs packages. And we already have similar language for those here:
https://dart.dev/dart-team-packages

We by-design don't have a support level that is "this is an experiment, but we're pretty sure that this will graduate". We don't have that, as the nature of any experiment is that we may learn that something does not carry its weight. For example, we may learn that something isn't as general as we thought it might be, or that something ends up getting so little usage, that it's not worth moving forward.

mit-mit:
@brianquinlan I updated this with the new verbiage we discussed

pkgs/cupertino_http/README.md
```diff
@@ -4,11 +4,19 @@
...
+[labs.dart.dev](https://dart.dev/dart-team-packages) pub publisher in order to
+solicit feedback. 
+
+For packages in the labs.dart.dev publisher we generally plan to either graduate
+the package into a supported publisher (dart.dev, tools.dart.dev) after a period
+of feedback and iteration, or discontinue the package. These packages have a
+much higher expected rate of API and breaking changes.
+
+Your feedback is valuable and will help us evolve this package. For general
+feedback, suggestions, and comments, please file an issue in the bug tracker.
```
brianquinlan, line 19: Could you make "bug tracker" a link to the bug tracker?

brianquinlan (approved):

pkgs/cupertino_http/README.md
mit-mit: Certainly, done!

Issue #618, opened 2021-09-22
Support for OPTIONS method


karina-klinkeviciute:
It would be good if this library would support OPTIONS HTTP method.

It would be especially helpful for multilanguage apps. 

Please consider this example: 

truncated

 
```json
  "actions": {
    "POST": {
      "year_of_birth": {
        "type": "integer",
        "required": false,
        "read_only": false,
        "label": "Gimimo metai",
        "min_value": -2147483648,
        "max_value": 2147483647
      },
      "uid": {
        "type": "string",
        "required": false,
        "read_only": true,
        "label": "Uid"
      },
      "email": {
        "type": "email",
        "required": false,
        "read_only": true,
        "label": "El. pašto adresas"
      },
      "gender": {
        "type": "choice",
        "required": false,
        "read_only": false,
        "label": "Lytis",
        "choices": [
          {
            "value": "woman",
            "display_name": "moteris"
          },
          {
            "value": "man",
            "display_name": "vyras"
          },
          {
            "value": "non-binary",
            "display_name": "nebinarinė"
          },
          {
            "value": "other",
            "display_name": "kita"
          },
          {
            "value": "prefer_not_to_say",
            "display_name": "nenoriu nurodyti"
          }
        ]
      },
```

The API (it's Django with Django Rest Framework, but I think other frameworks might give something as valuable too). If you see, it returns all translations in labels. I have worked on projects where options was also useful in other cases for a frontend. 

If there is a way to get options without this method, could you then document it?

natebosch:
> If there is a way to get options without this method, could you then document it?

You can construct a [`Request`](https://pub.dev/documentation/http/latest/http/Request/Request.html) with an arbitrary `method` and then [`send()`](https://pub.dev/documentation/http/latest/http/Client/send.html) it with a `Client`

dvaldivia:
The problem with a manual `Request` is that we are not allowed to set headers because these are final, is the only solution to add it as a method @natebosch ?

dvaldivia:
I created PR https://github.com/dart-lang/http/pull/756 to add support for `OPTIONS` method with headers

natebosch:
> we are not allowed to set headers because these are final

The field is a mutable map. For better or worse, the design is to add to that map.

```dart
Request optionsRequest(Uri url, Map<String, String> headers) => Request('OPTIONS', url)..headers.addAll(headers);
```

PR #756, opened 2022-08-14
Add OPTIONS Method to Client Interface and Client Implementation

dvaldivia:
Adds the `OPTIONS` method call to the `Client` interface so that the call can be invoked with the option to pass headers as opposed to trying to do it via raw `Request` which doesn't allow to set headers upon instantiation.


natebosch:
Thanks for the PR. I think an OPTIONS request is an unusual enough use case that it can have a little boilerplate.

See https://github.com/dart-lang/http/issues/618#issuecomment-1217224516 where I show how.

PR #758, opened 2022-08-15
[cupertino_http] A bit of refactoring of the example

mit-mit:
- Give a bit of user feedback for case where there is no search query
- Use async/await per Dart style
- Split code across a few files
- Improve indentation a bit


pkgs/cupertino_http/example/lib/get_books.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:convert';
+import 'package:http/http.dart';
+
+import 'book.dart';
+
+Future<List<Book>> getBooks(String query) async {
+  final response = await get(
```
brianquinlan, line 7: You don't think that it is valuable so have this comment?

// `get` will use the `Client` configured in main.

pkgs/cupertino_http/example/lib/get_books.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:convert';
```
brianquinlan, line 1: Don't you only see `main.dart` on pub (i.e. https://pub.dev/packages/cupertino_http/example) so the network code will disappear from the example.

pkgs/cupertino_http/example/lib/get_books.dart
mit-mit: Yes, good idea, added

pkgs/cupertino_http/example/lib/get_books.dart
mit-mit: Good point, moved the JSON code to book.dart and then the networking code back to main.dart.

mit-mit:
Updated, PTAL @brianquinlan

pkgs/cupertino_http/example/lib/main.dart
```diff
@@ -49,88 +45,71 @@ class HomePage extends StatefulWidget {
...
+    final response = await get(
+      Uri.https(
+        'www.googleapis.com',
+        '/books/v1/volumes',
+        {'q': query, 'maxResults': '40', 'printType': 'books'},
+      ),
+    );
+
+    final json = jsonDecode(utf8.decode(response.bodyBytes)) as Map;
+    return Book.listFromJson(json);
```
brianquinlan, line 67: Nice! I was annoying how much tedious parsing was getting in the way of my example ;-)

pkgs/cupertino_http/example/lib/book.dart
```diff
@@ -0,0 +1,32 @@
+class Book {
```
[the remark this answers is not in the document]
mit-mit: Yes!

PR #767, opened 2022-08-20
Make it possible to configure CronetEngine settings

brianquinlan:


pkgs/cronet_http/lib/cronet_client.dart
```diff
@@ -34,10 +122,27 @@ import 'src/messages.dart';
...
+  CronetEngine? _engine;
+  final bool _ownedEngine;
+
+  CronetClient([CronetEngine? engine])
+      : _engine = engine,
+        _ownedEngine = engine == null;
+
+  @override
+  void close() {
+    if (_ownedEngine) {
```
aam, line 134: would `_engine != null` be sufficient here making _ownedEngine property unnecessary?

pkgs/cronet_http/lib/src/messages.dart
```diff
@@ -219,6 +332,60 @@ class HttpApi {
...
+        code: (error['code'] as String?)!,
+        message: error['message'] as String?,
+        details: error['details'],
+      );
+    } else if (replyMap['result'] == null) {
+      throw PlatformException(
+        code: 'null-error',
+        message: 'Host platform returned null value for non-null return value.',
+      );
+    } else {
```
aam, line 360: You could drop `else` block reducing nesting/indentation.

aam: here and below.

pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart
```diff
@@ -0,0 +1,131 @@
...
+          request.response.headers.set('etag', '12345');
+          await request.response.close();
+        });
+    });
+    tearDown(() {
+      server.close();
+    });
+
+    test('disabled', () async {
+      final builder = await CronetEngine.build(cacheMode: CacheMode.disabled);
```
aam, line 36: I think these are engines, not builders (here and below).

aam (approved):

pkgs/cronet_http/lib/cronet_client.dart
brianquinlan: `_engine` gets assigned in `send` if not set in the constructor.

pkgs/cronet_http/lib/src/messages.dart
brianquinlan: This is generated code.

pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart
brianquinlan: Done

Issue #769, opened 2022-08-22
Allow the Client passed to runWithClient to return null


brianquinlan:
In which case the default Client will be used.

brianquinlan:
@mit-mit

PR #772, opened 2022-08-23
Better document the runWithClient(() => ..., client.call) pattern

brianquinlan:
Fixes https://github.com/dart-lang/http/issues/769

### DIFF
--- a/pkgs/cronet_http/example/lib/main.dart
+++ b/pkgs/cronet_http/example/lib/main.dart
@@ -14,6 +14,7 @@ import 'book.dart';
 void main() async {
   Client? client;
   if (Platform.isAndroid) {
+    WidgetsFlutterBinding.ensureInitialized();
     client = CronetClient(await CronetEngine.build(
         cacheMode: CacheMode.memory, cacheMaxSize: 1024 * 1024));
   }
--- a/pkgs/cronet_http/example/lib/main.dart
+++ b/pkgs/cronet_http/example/lib/main.dart
@@ -11,16 +11,13 @@ import 'package:http/http.dart';
 
 import 'book.dart';
 
-void main() async {
-  Client? client;
+void main() {
+  var clientFactory = Client.new; // Constructs the default client.
   if (Platform.isAndroid) {
     WidgetsFlutterBinding.ensureInitialized();
-    client = CronetClient(await CronetEngine.build(
-        cacheMode: CacheMode.memory, cacheMaxSize: 1024 * 1024));
+    clientFactory = CronetClient.new;
   }
-
-  // Use `client` if set, otherwise use the default Client.
-  runWithClient(() => runApp(const BookSearchApp()), () => client ?? Client());
+  runWithClient(() => runApp(const BookSearchApp()), clientFactory);
 }
 
 class BookSearchApp extends StatelessWidget {
--- a/pkgs/cronet_http/example/lib/main.dart
+++ b/pkgs/cronet_http/example/lib/main.dart
@@ -8,20 +8,17 @@ import 'dart:io';
 import 'package:cronet_http/cronet_client.dart';
 import 'package:flutter/material.dart';
 import 'package:http/http.dart';
-import 'package:http/io_client.dart';
 
 import 'book.dart';
 
 void main() {
-  late Client client;
+  Client? client;
   if (Platform.isAndroid) {
     client = CronetClient();
-  } else {
-    client = IOClient();
   }
 
-  // Run the app with the default `client` set to the one assigned above.
-  runWithClient(() => runApp(const BookSearchApp()), () => client);
+  // Use `client` if set, otherwise use the default Client.
+  runWithClient(() => runApp(const BookSearchApp()), () => client ?? Client());
 }
 
 class BookSearchApp extends StatelessWidget {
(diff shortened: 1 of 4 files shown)


pkgs/http/lib/src/client.dart
```diff
@@ -154,7 +154,7 @@ abstract class Client {
 @internal
 Client? get zoneClient {
   final client = Zone.current[#_clientToken];
-  return client == null ? null : (client as Client Function())();
+  return client == null ? null : (client as Client? Function())();
```
natebosch, line 157: [optional] If you wanted you could use `?.call` over the `?:`. This could be shortened all the way to a `=>`

```dart
Client? get zoneClient =>
    (Zone.current[#_clientToken] as Client? Function()?)?.call();
```

natebosch (commented):
Can you expand a bit on the motivation? Is this specifically to write tests in `package:http` and be able to run run the same `_test.dart` on multiple platforms?

pkgs/http/lib/src/client.dart
brianquinlan: I find the additional question marks harder to read.

brianquinlan:
> Can you expand a bit on the motivation? Is this specifically to write tests in `package:http` and be able to run run the same `_test.dart` on multiple platforms?

No, the motivation is to make it possible to use `runWithZone` where you only care about certain platforms e.g.

```dart
Client? client;
if (Platform.isIOS) {
  client = ...
}
runWithZone(MyWidget, () => client); // will get BrowserClient on web, IOClient on non-IOS VM.
```

natebosch:
Hmm. I would write that as

```dart
if (Platform.isIOS) {
  final client = ...
  runWithZone(something, () => client);
} else {
  something();
}
```

Or, if you really like the pattern, you could use

```dart
Client? client;
if (Platform.isIOS) {
  client = ...
}
runWithZone(MyWidget, () => client ?? Client()); // will get BrowserClient on web, IOClient on non-IOS VM.
```

I don't think I see the value in a nullable signature.

brianquinlan:
> Hmm. I would write that as
> 
> ```dart
> if (Platform.isIOS) {
>   final client = ...
>   runWithZone(something, () => client);
> } else {
>   something();
> }
> ```
> 
> Or, if you really like the pattern, you could use
> 
> ```dart
> Client? client;
> if (Platform.isIOS) {
>   client = ...
> }
> runWithZone(MyWidget, () => client ?? Client()); // will get BrowserClient on web, IOClient on non-IOS VM.
> ```
> 
> I don't think I see the value in a nullable signature.

@mit-mit Any opinion here?

mit-mit:
My concern is that the `else` part currently isn't that easy. What would I write under `something()` here?


```dart
if (Platform.isIOS) {
  final client = ...
  runWithZone(something, () => client);
} else {
  something();
}
```

I can't just write `client = IOClient();` or similar as that won't work on the web. I'd have to use config specific imports, which are hard.

But could I use something like this?

```dart
if (Platform.isIOS) {
  final client = ...
  runWithZone(something, () => client);
} else {
  run(something);
}
```

Or perhaps like this?

```dart
if (Platform.isIOS) {
  final client = ...
  runWithZone(something, () => client);
} else {
  runWithZone(something, http.defaultClient => client);
}
```

brianquinlan:
> Or perhaps like this?
> 
> ```dart
> if (Platform.isIOS) {
>   final client = ...
>   runWithZone(something, () => client);
> } else {
>   runWithZone(something, http.defaultClient => client);
> }
> ```

You'd write that as:

```dart
Client? client;
if (Platform.isIOS) {
  final client = ...
}
runWithZone(MyWidget, () => client ?? Client());  // The Client() constructor will select the correct default for the platform.
```

This seems reasonable to me and I can document that pattern.

mit-mit:
OK, that sounds good. I guess when no config of the client is needed, you can also just use this?

```dart
final client = Platform.isIOS ? CupertinoClient.defaultSessionConfiguration() : Client();
runWithClient(myFunction, () => client);
```

natebosch:
> OK, that sounds good. I guess when no config of the client is needed, you can also just use this?
> 
> ```dart
> final client = Platform.isIOS ? CupertinoClient.defaultSessionConfiguration() : Client();
> runWithClient(myFunction, () => client);
> ```

Yeah, that would be a good way to reuse a consistent `Client` instance too, which a `null` return couldn't achieve.

brianquinlan:
OK, I reframed this as a documentation fix. PTAL.

natebosch (approved):

pkgs/cronet_http/example/lib/main.dart
```diff
@@ -8,21 +8,19 @@ import 'dart:io';
...
 
 void main() async {
-  late Client client;
+  Client? client;
   if (Platform.isAndroid) {
+    WidgetsFlutterBinding.ensureInitialized();
     client = CronetClient(await CronetEngine.build(
         cacheMode: CacheMode.memory, cacheMaxSize: 1024 * 1024));
-  } else {
-    client = IOClient();
```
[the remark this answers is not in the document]
brianquinlan: This is a great point! I think that my version is incorrect because it would not work if you wrote:

```dart
final c = Client();
c.get(...)
c.close();
final d = Client()
d.get(...)  # Exception because the client has been closed!
```

This formulation seems better.

```dart
void main() async {
  var client = Client.new; // The default Client.
  if (Platform.isAndroid) {
    WidgetsFlutterBinding.ensureInitialized();
    client = CronetClient.new;
  }
  runWithClient(() => runApp(const BookSearchApp()), client);
}
```

natebosch (approved):

PR #786, opened 2022-09-07
Update pubspec.yaml version is preparation for publishing

brianquinlan:


natebosch (approved):
Do we have corresponding versions in the CHANGELOG?

brianquinlan:
> Do we have corresponding versions in the CHANGELOG?

Yes. I forgot to update pubspec.yaml when I updated the CHANGELOGs.

PR #794, opened 2022-09-14
Fix disappearing issue templates

brianquinlan:


kevmoo (approved):
Maybe delete the extra trailing newlines at the end of the files, but not a big deal....

brianquinlan:
> Maybe delete the extra trailing newlines at the end of the files, but not a big deal....

These files are generated but I'll try in a follow-up PR.

PR #799, opened 2022-09-19
http_client_conformance_tests: no plan to publish

kevmoo:
We can update this as needed.

### DIFF
--- a/pkgs/http_client_conformance_tests/pubspec.yaml
+++ b/pkgs/http_client_conformance_tests/pubspec.yaml
@@ -1,8 +1,8 @@
 name: http_client_conformance_tests
-description: >
+description: >-
   A library that tests whether implementations of package:http's `Client` class
   behave as expected.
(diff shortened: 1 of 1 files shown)


brianquinlan (approved):

lucavenir:
May I ask: why, tho? `http`'s [pub package description](https://pub.dev/packages/http) mentions this package being "composable". Therefore I try to do exactly that; but I took these tests for granted.

After all, it would be useful to have a pre-made suite of tests that can be run against any new wrapper around `BaseClient`. Isn't that the purpose of this lib?

brianquinlan:
That's the purpose but there is also effort involved in publishing and even coming up with a publishing policy. For example, ever test addition or change is likely to be breaking so does every change require a semver major bump? It is already possible to use the package through the [git packages mechanism](https://dart.dev/tools/pub/dependencies#git-packages). Let me know if that isn't sufficient.

PR #800, opened 2022-09-20
Fix a bug where cronet_http did not set content-length correctly

brianquinlan:
Fixes https://github.com/dart-lang/http/issues/801

### DIFF
--- a/pkgs/http_client_conformance_tests/lib/src/response_body_server.dart
+++ b/pkgs/http_client_conformance_tests/lib/src/response_body_server.dart
@@ -20,6 +20,10 @@ void hybridMain(StreamChannel<Object?> channel) async {
       await request.drain<void>();
       request.response.headers.set('Access-Control-Allow-Origin', '*');
       request.response.headers.set('Content-Type', 'text/plain');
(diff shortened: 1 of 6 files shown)


pkgs/http_client_conformance_tests/lib/src/response_body_server.dart
```diff
@@ -20,6 +20,10 @@ void hybridMain(StreamChannel<Object?> channel) async {
       await request.drain<void>();
       request.response.headers.set('Access-Control-Allow-Origin', '*');
       request.response.headers.set('Content-Type', 'text/plain');
+      if (request.requestedUri.pathSegments.isNotEmpty &&
+          request.requestedUri.pathSegments.last == 'length') {
+        request.response.contentLength = 'Hello World!'.length;
```
natebosch, line 25: [nit] how about adding a variable instead of repeating the content string.

natebosch (approved):

pkgs/http_client_conformance_tests/lib/src/response_body_server.dart
brianquinlan: Done.

Issue #609, opened 2021-08-17
Add support for `preserveHeaderCase`


natebosch:
On the SDK we now have an option to pass `preserveHeaderCase` when setting Http headers to work around the default behavior of converting them all to lowercase. Some (non spec compliant?) servers expect headers with specific capitalization.

I don't know if there is any equivalent behavior or concern on the web.

Some design options:
- An argument when constructing an `IOClient` to make the decision once at a client level.
  - **Pro:** Non breaking.
  - **Con:** Could be clumsy to use - cross platform code can't reference `IOClient`, may need to construct a specific client for some requests and a different client for others.
- Add a named argument to all APIs.
  - **Pro:** fine-grained control by request.
  - **Con:** If the behavior can't be supported on the web we'd have to ignore the argument.
- Preserve header case on the VM by default with no option to switch.
  - **Con:** Potentially breaking behavior change for existing users.

natebosch:
We could also consider flipping the default in the next major version since it seems that most folks who care would prefer to preserve case.

PiotrWpl:
Hi, any progress with this one?

natebosch:
@brianquinlan - WDYT about using `preserveHeaderCase` by default, and/or adding an option? I don't think this is something we can control on the web - is it possible on the other clients?

brianquinlan:
Setting `preserveHeaderCase: true` for `IOClient` seems reasonable to me. But this isn't something that we should require of clients because:
1. case insensitivity is required by RFC-2616
2. some clients (e.g. `package:cupertino_http`) cannot promise to to preserve case because their underlying implementations are not case preserving

natebosch:
> 1. case insensitivity is required by RFC-2616

Yeah - it feels weird to have to add features that explicitly go off-spec, but buggy servers do exist so we may as well try to make the client as compatible as possible.

It sounds like the best path forward may be to default to preserving case for the `IOClient`, but to keep the option out of the interface. At most we can add an IO specific constructor arg if it becomes necessary.

I don't know how breaking we should consider this change if we do just flip the default.

maikolb:
> > 1. case insensitivity is required by RFC-2616
> 
> Yeah - it feels weird to have to add features that explicitly go off-spec, but buggy servers do exist so we may as well try to make the client as compatible as possible.
> 
> It sounds like the best path forward may be to default to preserving case for the `IOClient`, but to keep the option out of the interface. At most we can add an IO specific constructor arg if it becomes necessary.
> 
[...]

PR #805, opened 2022-09-30
Changed preserveHeaderCase from false to true for request headers.

Mohammed3194:
I have changed `preserveHeaderCase` for headers from `false` to `true`. The custom headers defined in the back end server were not accepting the case-sensitive header (`X-Auth-Signature`).
The result was that I couldn't make API requests and was getting 401 unauthorized responses all the time.

Once after setting `preserveHeaderCase` to true API has started working. Consider this change or is that it is possible to change `preserveHeaderCase` from `post/get/put/delete` methods, where actual API requests are made.

This is a problem that most others are experiencing as well - #[Reference Link](https://stackoverflow.com/questions/73133740/401-unauthorized-error-in-flutter-but-works-fine-in-postman)

@brianquinlan, @kevmoo, @mit-mit, @natebosch, @devoncarew, @osa1, @Hamaad-Siddiqui, @aam, @guyluz11 

Thank You

### DIFF
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -60,7 +60,7 @@ class IOClient extends BaseClient {
         ..contentLength = (request.contentLength ?? -1)
         ..persistentConnection = request.persistentConnection;
       request.headers.forEach((name, value) {
-        ioRequest.headers.set(name, value);
+        ioRequest.headers.set(name, value, preserveHeaderCase: true);
       });
 
       var response = await stream.pipe(ioRequest) as HttpClientResponse;


natebosch:
I'm not sure if we need to wait to release this as a major version change with other breaking changes.

https://github.com/dart-lang/http/issues/609

brianquinlan:
Would this change cause a semantic difference between `browser_client` and `io_client`?

natebosch:
> Would this change cause a semantic difference between `browser_client` and `io_client`?

I don't think so. I think there is currently a semantic difference and this would remove it. AFAIK the browser does not change the case of headers.

natebosch:
This change does break some tests, but not many IIRC.

mosuem:
Closing this as it is stale and the CLA is missing. Feel free to reopen in case you still want to land it!

PR #817, opened 2022-11-09
blast_repo fixes

devoncarew:
Dependabot


devoncarew:
Ah, this is a mono_repo managed repo. @kevmoo - https://github.com/google/mono_repo.dart/pull/420 would let us use both mono_repo and dependabot together.

PR #843, opened 2022-12-29
Make it possible to use a custom CronetEngine with runWithClient

brianquinlan:


pkgs/cronet_http/lib/cronet_client.dart
```diff
@@ -138,11 +177,22 @@ class CronetClient extends BaseClient {
...
-    try {
-      _engine ??= await CronetEngine.build();
-    } catch (e) {
-      throw ClientException(e.toString(), request.url);
+    if (_engine == null) {
+      try {
+        if (_engineFuture == null) {
+          // Create the future here rather than in the [fromCronetEngineFuture]
+          // factory so that [close] does not have to await the future just to
+          // close it in the case where [send] is never called.
```
natebosch, line 185: Is there still a gap where `send` has been called, but the future has not resolved?

Maybe we could add an `_isClosed` bool field, and after the `await` here check for this rare situation?

pkgs/cronet_http/lib/cronet_client.dart
```diff
@@ -138,11 +177,22 @@ class CronetClient extends BaseClient {
 
   @override
   Future<StreamedResponse> send(BaseRequest request) async {
-    try {
-      _engine ??= await CronetEngine.build();
-    } catch (e) {
-      throw ClientException(e.toString(), request.url);
+    if (_engine == null) {
+      try {
+        if (_engineFuture == null) {
```
natebosch, line 182: It's possible to arrive here twice, and construct two different engines (one gets lost), if there are two non-serial calls to `send`.

Consider using `AsyncMemoizer` https://pub.dev/documentation/async/latest/async/AsyncMemoizer-class.html

```dart
  late final _ownEngineMemo = AsyncMemoizer<CronetEngine>();

          _engine = await _ownEngineMemo.runOnce(CronetEngine.build);
```

natebosch (approved):
Overall approach looks good. I think there are still some narrow async gaps with opportunities for incorrect behavior.

pkgs/cronet_http/lib/cronet_client.dart
brianquinlan: I don't understand the scenario. I don't think that there is a path through the initialization where a `CronetEngine` is created but not assigned to `_engine`.

natebosch: The situation I'm concerned about is where it is not _yet_ assigned to `_engine`.

```dart
var client = CronetClient.defaultCronetEngine();
unawaited(client.send(someRequest));
// We have called `CronetEngine.build()` and assigned `_engineFuture`, but _not_ assigned `_engine`
client.close();
```

brianquinlan: I think that I understand...

In that scenario `close()` would end up doing nothing, and the `send(someRequest)` would succeed. Which I think would be fine (I think that [`IOClient`](https://github.com/dart-lang/http/blob/master/pkgs/http/lib/src/io_client.dart) has the same behavior).

But future calls to `send()` would also succeed... but that is true of [`BrowserClient`](https://github.com/dart-lang/http/blob/master/pkgs/http/lib/src/browser_client.dart) as well.

I like the `IOClient` behavior better. I'll send you a CL to add `_isClosed`. If we decide to standardize the behavior, I can change `BrowserClient` as well.

PR #846, opened 2022-12-30
Add usage instructions for runWithClient and Flutter

brianquinlan:


natebosch (approved):
Do you know if there is any canonical Flutter doc that dicusses the way that `runApp` uses the zone from `ensureInitialized`?

Ideally it would be nice to avoid repeating this description in each API that uses zones, and to instead link to a single description.

This looks fine for now if we don't have a canonical link.

natebosch (approved):

PR #851, opened 2023-01-13
Add consistent implementations for `close`.

brianquinlan:


pkgs/cronet_http/lib/cronet_client.dart
```diff
@@ -170,13 +171,19 @@ class CronetClient extends BaseClient {
...
-    if (_ownedEngine) {
+    if (!_isClosed && _ownedEngine) {
       _engine?.close();
     }
+    _isClosed = true;
   }
 
   @override
   Future<StreamedResponse> send(BaseRequest request) async {
+    if (_isClosed) {
```
natebosch, line 182: Do we need to worry about the `CronetEngine` that never gets a `.close()` call in that narrow case where `.close` is called synchronously following `send`?

brianquinlan: No, the _CronetEngine should eventually get GCed.

natebosch (approved):

PR #853, opened 2023-01-18
✨ Add Cronet embedded tool

### LINKED ISSUE #807: cronet_http: depend on org.chromium.net:cronet-api instead

AlexV525:
Resolves #807.

Creates a new tool for publishing the non-GMS-based Cronet package. The tool is implemented with these concerns:
* Not creating symlinks because the structure can be changed easily.
* Not to restrict running on a single platform, so it's been written in Dart.
* Only modify files, but do not publish in this tool, leave the process to the operator, since the latest stable version can have breaking changes and require further CI tests (more jobs in `.github/workflows`).
* As easy as possible to maintain the tool.

### DIFF
--- a/pkgs/cronet_http/README_EMBEDDED.md
+++ b/pkgs/cronet_http/README_EMBEDDED.md
@@ -1,0 +1,23 @@
+An Android Flutter plugin that provides access to the
+[Cronet](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/package-summary)
+HTTP client.
+
+Cronet embedded is available as a standalone implementation
+which will brought additional 8MB for apps approximately.
+
+## Status: Experimental
+
+**NOTE**: This package is currently experimental and published under the
+[labs.dart.dev](https://dart.dev/dart-team-packages) pub publisher in order to
+solicit feedback. 
+
+For packages in the labs.dart.dev publisher we generally plan to either graduate
+the package into a supported publisher (dart.dev, tools.dart.dev) after a period
+of feedback and iteration, or discontinue the package. These packages have a
+much higher expected rate of API and breaking changes.
+
+Your feedback is valuable and will help us evolve this package. 
+For general feedback and suggestions please comment in the
+[feedback issue](https://github.com/dart-lang/http/issues/764).
+For bugs, please file an issue in the 
+[bug tracker](https://github.com/dart-lang/http/issues).
--- a/pkgs/cronet_http/README_EMBEDDED.md
+++ b/pkgs/cronet_http/README_EMBEDDED.md
@@ -2,22 +2,11 @@ An Android Flutter plugin that provides access to the
 [Cronet](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/package-summary)
 HTTP client.
 
-Cronet embedded is available as a standalone implementation
-which will brought additional 8MB for apps approximately.
+This package is identical to [`package:cronet_http`](https://pub.dev/packages/cronet_http)
+except that it embeds
+[Cronet](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/package-summary)
+rather than using the version included with
+[Google Play Services](https://developers.google.com/android/guides/overview).
+This increases the uncompressed size of the application by approximately 8MB.
 
-## Status: Experimental
-
-**NOTE**: This package is currently experimental and published under the
-[labs.dart.dev](https://dart.dev/dart-team-packages) pub publisher in order to
-solicit feedback. 
-
-For packages in the labs.dart.dev publisher we generally plan to either graduate
-the package into a supported publisher (dart.dev, tools.dart.dev) after a period
-of feedback and iteration, or discontinue the package. These packages have a
-much higher expected rate of API and breaking changes.
-
-Your feedback is valuable and will help us evolve this package. 
-For general feedback and suggestions please comment in the
-[feedback issue](https://github.com/dart-lang/http/issues/764).
-For bugs, please file an issue in the 
-[bug tracker](https://github.com/dart-lang/http/issues).
+See more details about the cronet_http at: https://pub.dev/packages/cronet_http.
--- a/pkgs/cronet_http/README_EMBEDDED.md
+++ b/pkgs/cronet_http/README_EMBEDDED.md
@@ -9,4 +9,4 @@ rather than using the version included with
 [Google Play Services](https://developers.google.com/android/guides/overview).
 This increases the uncompressed size of the application by approximately 8MB.
 
-See more details about the cronet_http at: https://pub.dev/packages/cronet_http.
+See more details about cronet_http at: https://pub.dev/packages/cronet_http.
--- a/pkgs/cronet_http/pubspec.yaml
+++ b/pkgs/cronet_http/pubspec.yaml
@@ -17,6 +17,7 @@ dev_dependencies:
   lints: ^1.0.0
   pigeon: ^3.2.3
   xml: ^6.1.0
+  yaml_edit: ^2.0.3
 
 flutter:
   plugin:
--- a/pkgs/cronet_http/pubspec.yaml
+++ b/pkgs/cronet_http/pubspec.yaml
@@ -16,6 +16,7 @@ dependencies:
 dev_dependencies:
   lints: ^1.0.0
   pigeon: ^3.2.3
+  xml: ^6.1.0
 
 flutter:
   plugin:
--- a/pkgs/cronet_http/tool/prepare_for_embedded.dart
+++ b/pkgs/cronet_http/tool/prepare_for_embedded.dart
@@ -47,9 +47,11 @@ void _writeImplementationToTheFile(String latestVersion) {
     ':\\d+.\\d+.\\d+[\'"]',
     multiLine: true,
   );
+  final newImplementation = 'org.chromium.net:cronet-embedded:$latestVersion';
+  print('Patching $newImplementation');
   final newGradleContent = gradleContent.replaceAll(
     implementationRegExp,
-    '    implementation "org.chromium.net:cronet-embedded:$latestVersion"',
+    '    implementation $newImplementation',
   );
   fBuildGradle.writeAsStringSync(newGradleContent);
   // Update pubspec.yaml
--- a/pkgs/cronet_http/tool/prepare_for_embedded.dart
+++ b/pkgs/cronet_http/tool/prepare_for_embedded.dart
@@ -1,6 +1,7 @@
 // Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+// ignore_for_file: lines_longer_than_80_chars
 
 import 'dart:io';
 
@@ -8,18 +9,49 @@ import 'package:http/http.dart' as http;
 import 'package:xml/xml.dart';
 import 'package:yaml_edit/yaml_edit.dart';
 
+/// The cronet_http directory is used to produce two packages:
+/// - `cronet_http`, which uses the Google Play Services version of Cronet.
+/// - `cronet_http_embedded`, which embeds Cronet.
+///
+/// The default configuration of this code is to use the Google Play Services version of Cronet.
+///
+/// The script transforms the configuration into one that embeds Cronet by:
+/// 1. Modifying the Gradle build file to reference the embedded version of Cronet.
+/// 2. Modifying the *name* and *description* in `pubspec.yaml`.
+/// 3. Replaying `README.md` with `README_EMBEDDED.md`.
+///
+/// After running this script, `flutter pub publish` can be run to update package:cronet_http_embedded.
+///
+/// NOTE: This script modifies the above files in place.
+late final Directory _directory;
+
+const String _gmsDependencyName = 'com.google.android.gms:play-services-cronet';
+const String _embeddedDependencyName = 'org.chromium.net:cronet-embedded';
+const String _packageName = 'cronet_http_embedded';
+const String _packageDescription = 'An Android Flutter plugin that '
+    'provides access to the Cronet HTTP client. '
+    'Identical to package:cronet_http except that it embeds Cronet '
+    'rather than relying on Google Play Services.';
+final Uri _cronetVersionUri = Uri.https(
+  'dl.google.com',
+  'android/maven2/org/chromium/net/group-index.xml',
+);
+
 void main() async {
-  final latestVersion = await _getLatestVersion();
-  _writeImplementationToTheFile(latestVersion);
-  _replaceREADME();
+  if (Directory.current.path.endsWith('tool')) {
+    _directory = Directory.current.parent;
+  } else {
+    _directory = Directory.current;
+  }
+
+  final latestVersion = await _getLatestCronetVersion();
+  updateCronetDependency(latestVersion);
+  updatePubSpec();
+  updateReadme();
 }
 
-Future<String> _getLatestVersion() async {
-  final url = Uri.https(
-    'dl.google.com',
-    'android/maven2/org/chromium/net/group-index.xml',
-  );
-  final response = await http.get(url);
+Future<String> _getLatestCronetVersion() async {
+  final response = await http.get(_cronetVersionUri);
   final parsedXml = XmlDocument.parse(response.body);
   final embeddedNode = parsedXml.children
       .singleWhere((e) => e is XmlElement)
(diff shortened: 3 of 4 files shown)


AlexV525:
@brianquinlan @natebosch

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,62 @@
...
+    'dl.google.com',
+    'android/maven2/org/chromium/net/group-index.xml',
+  );
+  final response = await http.get(url);
+  final parsedXml = XmlDocument.parse(response.body);
+  final embeddedNode = parsedXml.children
+      .singleWhere((e) => e is XmlElement)
+      .children
+      .singleWhere((e) => e is XmlElement && e.name.local == 'cronet-embedded');
+  final stableVersionReg = RegExp(r'^(\d+).(\d+).(\d+)$');
```
AlexV525, line 26:
```suggestion
  final stableVersionReg = RegExp(r'^\d+.\d+.\d+$');
```

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,62 @@
...
+void _writeImplementationToTheFile(String latestVersion) {
+  var dir = Directory.current;
+  if (dir.path.endsWith('tool')) {
+    dir = dir.parent;
+  }
+  // Update android/build.gradle
+  final fBuildGradle = File('${dir.path}/android/build.gradle');
+  final gradleContent = fBuildGradle.readAsStringSync();
+  final implementationRegExp = RegExp(
+    '^(\\s*)implementation [\'"]'
```
AlexV525, line 44:
```suggestion
    '^\\s*implementation [\'"]'
```

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,62 @@
...
+    'com.google.android.gms:play-services-cronet'
+    ':(\\d+.\\d+.\\d+)[\'"]',
```
AlexV525, line 46:
```suggestion
    ':\\d+.\\d+.\\d+[\'"]',
```

brianquinlan:
@natebosch Do you know how the upcoming auto-publish feature works? Will it be possible to trigger a script like this as part of the publishing process?

AlexV525:
FYI, previously when we do automatically publish on GitHub, it's easy to run any scripts in jobs. e.g. https://github.com/fluttercandies/flutter_photo_manager/blob/361773c98c4f9665ee61ef25f1a6aa003d7692ec/.github/workflows/publish.yml#L14

natebosch:
@devoncarew is setting up autopublish on some repos. I don't think we've set up any mono repos yet - so we may need some work in the `mono_repo` tool.

devoncarew:
From my understanding, you want to run a custom dart script before publishing? It'll perform a native build or provision native artifacts?

The new CI publishing feature has a few different ways it can operate. The way I've used it is configure the package on pub.dev to be published from a github action. Then, when a github action is triggered by a tag on a commit, running `dart pub publish --force` will recognize that it's running in a github action, has an 'OIDC' token, and pass that as authentication to pub.dev when publishing.

The tooling I've written for publishing automation - that does validation for PRs and publishing for tagged commits on main - is likely too purpose built to re-use here (it does however support mono-repos :) ).

If you think you want to use some publishing automation here, you could:
- add a `.github/workflows/publish_cronet_http.yaml` file, loosely based on https://github.com/devoncarew/firehose/blob/main/.github/workflows/publish_internal.yaml
- use the pattern `cronet_http-v[0-9]+.[0-9]+.[0-9]+*` as the tag's glob pattern to trigger on
- in that workflow file, you want to make sure to have the right cwd for your job steps (`working-directory: foo`)
- run whatever arbitrary script you want
- finish with a call to `dart pub publish --force`

AlexV525:
Okay according to the above discussions and advises from @devoncarew, we can make an auto-publish workflow, and also allow it to be done manually using the simple Dart tool. I'll go for this goal then, please vote 👍 if you guys think it's valid.

brianquinlan:
@AlexV525 Your solution will work but it feels a bit hacky to me. 

Let's take a step make and think about options:
1. Fork this repo, make the needed changes, and periodically sync the fork:
    - The sync can be [automated](https://github.com/marketplace/actions/fork-sync)
    - There would be no coordination in releases i.e. a release of `package:cronet_http` would not automatically release `package:cronet_http_embedded`
    -  The fork would include unnecessary packages like `cupertino_http`
2. Create a script that modifies this repository in-place (this approach): 
    - Error prone if used done manually (but reasonable if done through automation)
    - Will double CI time if also runs with the script-modified changes
3. Create a branch, make the needed changes, and periodically sync from main
    -  Similar to (1) except the Dart team would have to manage it

Is there some other approach?

AlexV525:
The main guideline I based on when I was developing ideas is **Avoid manipulation by humans as possible, let the automation rolls**. The guideline will avoid exceptions made by human mistakes, such as forgetting to apply patches, forgetting to sync the code base, merging leads to code loss, etc.

I'll base on this guideline to discuss the above options.

### Fork this repo, make the needed changes, and periodically sync the fork

>    * The sync can be [automated](https://github.com/marketplace/actions/fork-sync)
>    * There would be no coordination in releases i.e. a release of `package:cronet_http` would not automatically release `package:cronet_http_embedded`

Not only the sync can be done automatically, as we discussed, but the publishing can also do it as well. The coordination can be bridged using GitHub Actions too, but it'll involve crossing repo calls, which seems too much for the proposal.

>    * The fork would include unnecessary packages like `cupertino_http`

That is correct, but the impact is not as much as we thought.

### Create a script that modifies this repository in-place (this approach)

>    * Error prone if used done manually (but reasonable if done through automation)

Given the complexity and the corresponding change made by the tool, I'll say it'll barely meet exceptions during regular development. The idea is to get rid of manual publishing, so if we think too much about doing this manually, we'll come to another path eventually.

>    * Will double CI time if also runs with the script-modified changes

The time cost can be replaced with space cost (running two jobs parallel). And if new exceptions were thrown during the publishing of the embedded package, it can be easily solved during PRs. It actually saves time for debugging incompatible changes.

### Create a branch, make the needed changes, and periodically sync from main

>    * Similar to (1) except the Dart team would have to manage it

Any manual manipulation should be avoided when doing patches or publishing, which will make the package solid.

### Other approaches?

I don't come up with any at present.

AlexV525:
Frankly, the difference between the GMS-based and embedded is tiny in the code base, and the community as well as [the Dart team has already explored automatic publishing](https://dart.dev/tools/pub/automated-publishing). We should not have many issues with maintaining a publishing tool.

AlexV525:
Hi @brianquinlan, did you come up with more thoughts or considerations in these weeks?

brianquinlan:
Hey @AlexV525 ,

I'm patching in the code now to take a look.

brianquinlan:
Hey @AlexV525, 

I think that you need to make at least two changes:
1. you need to modify the `name` and `description` in `pubspec.yaml`. The name should probably be something like `cronet_http_embedded` and the `description` could be `An Android Flutter plugin that provides access to the Cronet HTTP client. Identical to package:cronet_http except that it embeds Cronet rather than relying on Google Play Services.
2. we probably need a new readme (maybe `README_EMBEDDED.md`) that explains that the difference between it and `package:cronet_http` is and then links to `package:cronet_http` (so we don't need to keep the documentation in sync).
3. Your `prepare_for_embedded.dart` script should rename `README_EMBEDDED.md` to `README.md`

pkgs/cronet_http/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
 dev_dependencies:
   lints: ^1.0.0
   pigeon: ^3.2.3
+  xml: ^6.1.0
```
kevmoo, line 19: I got freaked out. It's only a dev_dependency. 😌

AlexV525:
@brianquinlan Updated accroding to your suggestions.

AlexV525:
Also please let me know if we can setup a GitHub action for this, maybe in separate PRs.

pkgs/cronet_http/README_EMBEDDED.md
```diff
@@ -0,0 +1,23 @@
+An Android Flutter plugin that provides access to the
+[Cronet](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/package-summary)
+HTTP client.
+
+Cronet embedded is available as a standalone implementation
+which will brought additional 8MB for apps approximately.
+
+## Status: Experimental
```
brianquinlan, line 8: I'd remove this line and everything after it. It will be annoying to keep this file in sync with README.md

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,79 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:http/http.dart' as http;
+import 'package:xml/xml.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+void main() async {
+  final latestVersion = await _getLatestVersion();
```
brianquinlan, line 12: I have a few code nits.

I'd create three functions that do the three things that you want to do here:
1. `updateCronetDependency`
2. `updatePubSpec`
3. `updateReadme`

You can keep `_getLatestVersion` as a helper but I'd call it `_getLatestCronetVersion`.

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,79 @@
...
+  final fBuildGradle = File('${dir.path}/android/build.gradle');
+  final gradleContent = fBuildGradle.readAsStringSync();
+  final implementationRegExp = RegExp(
+    '^\\s*implementation [\'"]'
+    'com.google.android.gms:play-services-cronet'
+    ':\\d+.\\d+.\\d+[\'"]',
+    multiLine: true,
+  );
+  final newImplementation = 'org.chromium.net:cronet-embedded:$latestVersion';
+  print('Patching $newImplementation');
```
brianquinlan, line 52: You can remove this.

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,79 @@
...
+    '    implementation $newImplementation',
+  );
+  fBuildGradle.writeAsStringSync(newGradleContent);
+  // Update pubspec.yaml
+  final fPubspec = File('${dir.path}/pubspec.yaml');
+  final yamlEditor = YamlEditor(fPubspec.readAsStringSync())
+    ..update(['name'], 'cronet_http_embedded')
+    ..update(
+      ['description'],
+      'An Android Flutter plugin that '
```
brianquinlan, line 64: Maybe make this a global constant?

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,79 @@
...
+  print('Patching $newImplementation');
+  final newGradleContent = gradleContent.replaceAll(
+    implementationRegExp,
+    '    implementation $newImplementation',
+  );
+  fBuildGradle.writeAsStringSync(newGradleContent);
+  // Update pubspec.yaml
+  final fPubspec = File('${dir.path}/pubspec.yaml');
+  final yamlEditor = YamlEditor(fPubspec.readAsStringSync())
+    ..update(['name'], 'cronet_http_embedded')
```
brianquinlan, line 61: Maybe make this a global constant?

pkgs/cronet_http/tool/run_pigeon.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 # Generate the platform messages used by cronet_http.
+cd ../
```
brianquinlan, line 4: Why did you change this?

brianquinlan:
I have a few code nits but overall it looks good!

I wouldn't worry about having a github action for now. I think that we are going to roll out automatic publishing soon and we can use this script when that rolls out. For now I'll just use it manually when I publish package:cronet_http.

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
brianquinlan, line 4: I think that we need a big comment here to explain what this does. Like:

/// The `cronet_http` directory is used to produce two packages:
/// 1. `package:cronet_http`, which uses the Google Play Services version of Cronet
/// 2. `package:cronet_http_embedded`, which embeds Cronet
///
/// The default configuration of this code is to use the Google Play Services version of Cronet.
///
/// This script transforms the default configuration into one that embeds Cronet by:
/// 1. modifying the Gradle build file to reference the embedded version of Cronet
/// 2. modifying the `name` and `description` in `pubspec.yaml`
/// 3. replaying `README.md` with `README_EMBEDDED.md`
///
/// After running this script, `flutter pub publish` can be run to update `package:cronet_http_embedded`.
///
/// **NOTE:*** This script modifies the above files **in place**.

brianquinlan (changes requested):

pkgs/cronet_http/tool/prepare_for_embedded.dart
AlexV525: I thought this would be helpful to let you know which version of the embedded cronet you're using. WDYT?

pkgs/cronet_http/README_EMBEDDED.md
AlexV525: I think we can leave at least a *more details* link like:

> See more details about the cronet_http at: https://pub.dev/packages/cronet_http.

pkgs/cronet_http/tool/run_pigeon.sh
AlexV525: It's under the root path previously, now it's under `/tool`.

pkgs/cronet_http/README_EMBEDDED.md
```diff
@@ -0,0 +1,12 @@
...
+HTTP client.
+
+This package is identical to [`package:cronet_http`](https://pub.dev/packages/cronet_http)
+except that it embeds
+[Cronet](https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/package-summary)
+rather than using the version included with
+[Google Play Services](https://developers.google.com/android/guides/overview).
+This increases the uncompressed size of the application by approximately 8MB.
+
+See more details about the cronet_http at: https://pub.dev/packages/cronet_http.
```
brianquinlan, line 12: Remove "the"

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,102 @@
...
+// BSD-style license that can be found in the LICENSE file.
+// ignore_for_file: lines_longer_than_80_chars
+
+import 'dart:io';
+
+import 'package:http/http.dart' as http;
+import 'package:xml/xml.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+/// The cronet_http directory is used to produce two packages:
```
brianquinlan, line 12: I think that convention is that you put the file comment before the imports.

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,102 @@
...
+/// - `cronet_http`, which uses the Google Play Services version of Cronet.
+/// - `cronet_http_embedded`, which embeds Cronet.
+///
+/// The default configuration of this code is to use the Google Play Services version of Cronet.
+///
+/// The script transforms the configuration into one that embeds Cronet by:
+/// 1. Modifying the Gradle build file to reference the embedded version of Cronet.
+/// 2. Modifying the *name* and *description* in `pubspec.yaml`.
+/// 3. Replaying `README.md` with `README_EMBEDDED.md`.
```
brianquinlan, line 21: Replacing

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,102 @@
...
+///
+/// After running this script, `flutter pub publish` can be run to update package:cronet_http_embedded.
```
brianquinlan, line 23: Wrap at 80 columns please.

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,102 @@
...
+///
+/// NOTE: This script modifies the above files in place.
+late final Directory _directory;
+
+const String _gmsDependencyName = 'com.google.android.gms:play-services-cronet';
```
brianquinlan, line 28: All of these type declarations are redundant, please remove them i.e. just write
const _gmsDependencyName = "...";

pkgs/cronet_http/tool/prepare_for_embedded.dart
```diff
@@ -0,0 +1,102 @@
...
+///
+/// The script transforms the configuration into one that embeds Cronet by:
+/// 1. Modifying the Gradle build file to reference the embedded version of Cronet.
+/// 2. Modifying the *name* and *description* in `pubspec.yaml`.
+/// 3. Replaying `README.md` with `README_EMBEDDED.md`.
+///
+/// After running this script, `flutter pub publish` can be run to update package:cronet_http_embedded.
+///
+/// NOTE: This script modifies the above files in place.
+late final Directory _directory;
```
brianquinlan, line 26: Maybe `_packageDirectory`?

pkgs/cronet_http/tool/prepare_for_embedded.dart
brianquinlan: OK.

pkgs/cronet_http/tool/run_pigeon.sh
brianquinlan: OK.

brianquinlan (changes requested):

pkgs/cronet_http/tool/prepare_for_embedded.dart
AlexV525: Used to write with Flutter. :)

brianquinlan:
Hey @AlexV525,

Thanks for your contribution and your patience!

I will try to do a release tomorrow.

brianquinlan (approved):

AlexV525:
> I will try to do a release tomorrow.

Is it working as expected?

brianquinlan:
@AlexV525 The release is https://pub.dev/packages/cronet_http_embedded

You could check to see if it works as you expected ;-)

DemoJameson:
build failed with cronet_http_embedded on android, since `implementation com.google.android.gms:play-services-cronet:18.0.1` lack of double quotes
```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\DemoJameson\AppData\Local\Pub\Cache\hosted\pub.flutter-io.cn\cronet_http_embedded-0.2.0\android\build.gradle' line: 49

* What went wrong:
Could not compile build file 'C:\Users\DemoJameson\AppData\Local\Pub\Cache\hosted\pub.flutter-io.cn\cronet_http_embedded-0.2.0\android\build.gradle'.
> startup failed:
  build file 'C:\Users\DemoJameson\AppData\Local\Pub\Cache\hosted\pub.flutter-io.cn\cronet_http_embedded-0.2.0\android\build.gradle': 49: Unexpected input: '{' @ line 49, column 14.
     dependencies {
                  ^

  1 error
```

brianquinlan:
I just released version 0.2.1 - could you check to see if that fixes the problem?

PR #865, opened 2023-02-07
Remove dependency on package:path

sigurdm:
Outside `test/` it was only used for the `basename` functionality. We can use Uri.pathsegments.last for that.

### DIFF
--- a/pkgs/http/lib/src/multipart_file_io.dart
+++ b/pkgs/http/lib/src/multipart_file_io.dart
@@ -11,7 +11,7 @@ import 'multipart_file.dart';
 
 Future<MultipartFile> multipartFileFromPath(String field, String filePath,
     {String? filename, MediaType? contentType}) async {
-  late var segments = File(filePath).uri.pathSegments;
+  late var segments = Uri.file(filePath).pathSegments;
   filename ??= segments.isEmpty ? '' : segments.last;
   var file = File(filePath);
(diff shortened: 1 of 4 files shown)


pkgs/http/lib/src/multipart_file_io.dart
```diff
@@ -5,14 +5,14 @@
...
 import 'package:http_parser/http_parser.dart';
-import 'package:path/path.dart' as p;
 
 import 'byte_stream.dart';
 import 'multipart_file.dart';
 
 Future<MultipartFile> multipartFileFromPath(String field, String filePath,
     {String? filename, MediaType? contentType}) async {
-  filename ??= p.basename(filePath);
+  late var segments = File(filePath).uri.pathSegments;
```
jakemac53, line 14: Can you go directly to the uri instead of through File?

jakemac53: also curious why this is late?

sigurdm: > Can you go directly to the uri instead of through File?
Yes! That is better.

> why this is late?
It only has to be computed if filename is `null`.

jakemac53: oh I see just to make it lazy, sgtm then

jakemac53 (approved):

PR #868, opened 2023-02-08
Add a flag to allow the default Client to be tree shaken away.

brianquinlan:
Currently this can save 128,624 bytes of disk space if you use a non-dart:io HttpClient.

### DIFF
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -15,7 +15,13 @@ import 'streamed_response.dart';
 /// Create a [BrowserClient].
 ///
 /// Used from conditional imports, matches the definition in `client_stub.dart`.
-BaseClient createClient() => BrowserClient();
+BaseClient createClient() {
+  if (const bool.fromEnvironment('no_default_http_client')) {
+    throw StateError('no_default_http_client was defined but runWithClient '
+        'was not used to configure a Client implementation.');
+  }
+  return BrowserClient();
+}
 
 /// A `dart:html`-based HTTP client that runs in the browser and is backed by
 /// XMLHttpRequests.
--- a/pkgs/http/lib/src/client.dart
+++ b/pkgs/http/lib/src/client.dart
@@ -217,6 +217,10 @@ Client? get zoneClient {
 /// $ flutter build appbundle --dart-define=no_default_http_client=true ...
 /// $ dart compile exe --define=no_default_http_client=true ...
 /// ```
+///
+/// If `no_default_http_client=true` is set then any call to the [Client]
+/// factory (i.e. `Client()`) outside of the [Zone] created by [runWithClient]
+/// will throw [StateError].
 R runWithClient<R>(R Function() body, Client Function() clientFactory,
         {ZoneSpecification? zoneSpecification}) =>
     runZoned(body,
--- a/pkgs/http/lib/src/client.dart
+++ b/pkgs/http/lib/src/client.dart
@@ -210,6 +210,13 @@ Client? get zoneClient {
 ///  runWithClient(() => runApp(const MyApp()), clientFactory);
 /// }
 /// ```
+///
+/// If [runWithClient] is used and the environment defines
+/// `no_default_http_client=true` then generated binaries may be smaller e.g.
+/// ```shell
+/// $ flutter build appbundle --dart-define=no_default_http_client=true ...
+/// $ dart compile exe --define=no_default_http_client=true ...
+/// ```
 R runWithClient<R>(R Function() body, Client Function() clientFactory,
         {ZoneSpecification? zoneSpecification}) =>
     runZoned(body,
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -12,7 +12,13 @@ import 'io_streamed_response.dart';
 /// Create an [IOClient].
 ///
 /// Used from conditional imports, matches the definition in `client_stub.dart`.
-BaseClient createClient() => IOClient();
+BaseClient createClient() {
+  if (const bool.fromEnvironment('no_default_http_client')) {
+    throw StateError('no_default_http_client was defined but runWithClient '
+        'was not used to configure a Client implementation.');
+  }
+  return IOClient();
+}
 
 /// Exception thrown when the underlying [HttpClient] throws a
 /// [SocketException].
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -157,6 +157,41 @@ jobs:
         if: "always() && steps.pkgs_http_client_conformance_tests_pub_upgrade.conclusion == 'success'"
         working-directory: pkgs/http_client_conformance_tests
   job_005:
+    name: "unit_test; Dart 2.14.0; PKG: pkgs/http; `dart test --platform chrome --define=no_default_http_client=true test/no_default_http_client_test.dart`"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cache Pub hosted dependencies
+        uses: actions/cache@627f0f41f6904a5b1efbaed9f96d9eb58e92e920
+        with:
+          path: "~/.pub-cache/hosted"
+          key: "os:ubuntu-latest;pub-cache-hosted;sdk:2.14.0;packages:pkgs/http;commands:test_3"
+          restore-keys: |
+            os:ubuntu-latest;pub-cache-hosted;sdk:2.14.0;packages:pkgs/http
+            os:ubuntu-latest;pub-cache-hosted;sdk:2.14.0
+            os:ubuntu-latest;pub-cache-hosted
+            os:ubuntu-latest
+      - name: Setup Dart SDK
+        uses: dart-lang/setup-dart@a57a6c04cf7d4840e88432aad6281d1e125f0d46
+        with:
+          sdk: "2.14.0"
+      - id: checkout
+        name: Checkout repository
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - id: pkgs_http_pub_upgrade
+        name: pkgs/http; dart pub upgrade
+        run: dart pub upgrade
+        if: "always() && steps.checkout.conclusion == 'success'"
+        working-directory: pkgs/http
+      - name: "pkgs/http; dart test --platform chrome --define=no_default_http_client=true test/no_default_http_client_test.dart"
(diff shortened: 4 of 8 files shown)


pkgs/http/lib/src/client.dart
```diff
@@ -210,6 +210,13 @@ Client? get zoneClient {
 ///  runWithClient(() => runApp(const MyApp()), clientFactory);
 /// }
 /// ```
+///
+/// If [runWithClient] is used and the environment defines
+/// `no_default_http_client=true` then generated binaries may be smaller e.g.
```
natebosch, line 215: Can we expand this? I would also describe the risk of using `no_defualt_http_client=true` any usage of `Client()` outside of this zone will throw.

natebosch (commented):
Can you add a CHANGELOG entry?

pkgs/http/lib/src/client.dart
brianquinlan: Done

brianquinlan:
> Can you add a CHANGELOG entry?

There already is one. Are there details that I'm missing?

brianquinlan:
@mkustermann FYI

pkgs/http/lib/src/client.dart
```diff
@@ -210,6 +210,17 @@ Client? get zoneClient {
...
+/// ```shell
+/// $ flutter build appbundle --dart-define=no_default_http_client=true ...
+/// $ dart compile exe --define=no_default_http_client=true ...
+/// ```
+///
+/// If `no_default_http_client=true` is set then any call to the [Client]
+/// factory (i.e. `Client()`) outside of the [Zone] created by [runWithClient]
+/// will throw [StateError].
```
natebosch, line 223: [optional] I think you can link directly to the constructor in DartDoc with [Client.new].

```suggestion
/// If `no_default_http_client=true` is set then any call to [Client.new]
/// outside of the [Zone] created by [runWithClient] will throw [StateError].
```

natebosch (approved):

natebosch:
> There already is one. Are there details that I'm missing?

Huh, maybe I misclicked and collapsed the diff and didn't notice...

pkgs/http/lib/src/io_client.dart
```diff
@@ -12,7 +12,13 @@ import 'io_streamed_response.dart';
 /// Create an [IOClient].
 ///
 /// Used from conditional imports, matches the definition in `client_stub.dart`.
-BaseClient createClient() => IOClient();
+BaseClient createClient() {
+  if (const bool.fromEnvironment('no_default_http_client')) {
+    throw StateError('no_default_http_client was defined but runWithClient '
+        'was not used to configure a Client implementation.');
+  }
+  return IOClient();
```
mkustermann, line 20: This defaults to false. So during development in `flutter-debug` mode / JIT mode things work fine and people may start getting exceptions when building & running production mode (if the flag is set to true).

It's also unclear where one would set this environment define and how someone setting it would know that's safe: If one uses some 3rd party package that happens to use this API, then how should app developer know?

I think there's two other options:
- no concept of default client: people have to create the specific client they want and pass it around (directly or via zone)
- we have a platform-specific default.

The first one would be API breaking. So the second one seems appealing. Why can't we change this instead to

```dart
BaseClient createClient() {
  if (Platform.isAndroid) return CornetHttpClient();
  if (Platform.isMac || Platform.isIOS) return CupertinoHttpClient();
  return IOClient();
}
```

This way it will affect the
- the main app
- any transitive 3rd party package that's used in the app
- works in any isolate

Especially the ladder part is difficult: Even if one sets up a custom http client in `main()`, one may call third party code that may spawn isolates, which may talk to network and no nothing about the custom http client setup in `main()`.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -15,7 +15,13 @@ import 'streamed_response.dart';
 /// Create a [BrowserClient].
 ///
 /// Used from conditional imports, matches the definition in `client_stub.dart`.
-BaseClient createClient() => BrowserClient();
+BaseClient createClient() {
+  if (const bool.fromEnvironment('no_default_http_client')) {
```
kevmoo, line 19: Aren't these often SCREAMING_FOR_SOME_REASON?

No need for a foolish consistency, just checking.

brianquinlan: After a quick survey, they don't seem to be.

pkgs/http/lib/src/io_client.dart
brianquinlan: > This defaults to false. So during development in `flutter-debug` mode / JIT mode things work fine and people may start getting exceptions when building & running production mode (if the flag is set to true).
> 
> It's also unclear where one would set this environment define and how someone setting it would know that's safe: If one uses some 3rd party package that happens to use this API, then how should app developer know?

This is documented in [`runWithClient`](https://github.com/dart-lang/http/pull/868/files#diff-f37ec58d142d7b7a15575ec724e388c882e028968fb945451030aa1dd9924abfR214) because, as you said, setting this flag without configuring a flag is not safe.

I'd be happy to expand the documentation if you think it isn't clear or if it is missing details.

> I think there's two other options:
> 
> * no concept of default client: people have to create the specific client they want and pass it around (directly or via zone)

We discussed that in the last ecosystem meeting - the consensus was that is too large of a usability hit.

> * we have a platform-specific default.
> 
> The first one would be API breaking. So the second one seems appealing. Why can't we change this instead to
> 
> ```dart
> BaseClient createClient() {
>   if (Platform.isAndroid) return CornetHttpClient();
>   if (Platform.isMac || Platform.isIOS) return CupertinoHttpClient();
>   return IOClient();
> }
> ```

There is a practical reason why this won't work - the dependency tree is now a diamond because `package:http` now depends on `package:cronet_http` and vise-versa. `package:cronet_http` is also a flutter plugin so it can't be consumed by a pure-Dart package like `package:http`

> This way it will affect the
> 
> * the main app
> * any transitive 3rd party package that's used in the app
> * works in any isolate

Doesn't that just move the problem? Now the user is paying the disk cost of `package:cronet_http` on Android even if they don't use it.

> Especially the ladder part is difficult to do otherwise: Even if one sets up a custom http client in `main()` via zones, one may call third party code that may spawn isolates, which may talk to network and know nothing about the custom http client setup in `main()`.

Yeah, that really sucks. We already have that problem with `HttpOverrides` and friends. I don't know of a good solution.

Having a federated plugin with sensible defaults could solve this problem but, AFAIK, federated plugins do not allow differentiated interfaces so you can't provide implementation-specific configuration or functionality.

mkustermann: > There is a practical reason why this won't work - the dependency tree is now a diamond because package:http now depends on package:cronet_http and vise-versa. package:cronet_http is also a flutter plugin so it can't be consumed by a pure-Dart package like package:http

The diamond is easy to solve, we introduce `package:http_base`, put the `Client` interface in it, which cuppertino/... packages can use and then we re-export the interface from `package:http`.

The fact that `package:http` will have a dependency on flutter plugin and can't be used anymore outside flutter is an actual technical problem. Maybe it would be worthwhile discussing whether that can be fixed.

> Doesn't that just move the problem? Now the user is paying the disk cost of package:cronet_http on Android even if they don't use it.

One can extend this a little by 
```dart
BaseClient createClient() {
  // Our platform-specific code tree shaker will make this constant and the switch as well, allowing tree
  // shaking.
  final default = Platform.isAndroid ? 'cronet' : ...;
  switch (const String.fromEnvironment('package.http.backend', defaultTo: default)) {
    case 'cronet': return CronetHttpClient();
    case 'cuppertino': return CuppertinoHttpClient();
    ...
    default: return IOClient();
  }
}
```
Again, it solves isolate use case as well.

If we also wanted to support 3rd party clients, maybe a more flexible way to generate above code could work.

I think the solution proposed here may simply not work very well (i.e. people may not be able to flip on the constant to allow tree shaking) and it increases probability of using multiple different http clients if not all of the app and all isolates use the same zone-injected client.

brianquinlan: > > There is a practical reason why this won't work - the dependency tree is now a diamond because package:http now depends on package:cronet_http and vise-versa. package:cronet_http is also a flutter plugin so it can't be consumed by a pure-Dart package like package:http
> 
> The diamond is easy to solve, we introduce `package:http_base`, put the `Client` interface in it, which cuppertino/... packages can use and then we re-export the interface from `package:http`.
> 
> The fact that `package:http` will have a dependency on flutter plugin and can't be used anymore outside flutter is an actual technical problem. Maybe it would be worthwhile discussing whether that can be fixed.
> 
> > Doesn't that just move the problem? Now the user is paying the disk cost of package:cronet_http on Android even if they don't use it.
> 
> One can extend this a little by
> 
> ```dart
> BaseClient createClient() {
>   // Our platform-specific code tree shaker will make this constant and the switch as well, allowing tree
>   // shaking.
>   final default = Platform.isAndroid ? 'cronet' : ...;
>   switch (const String.fromEnvironment('package.http.backend', defaultTo: default)) {
>     case 'cronet': return CronetHttpClient();
>     case 'cuppertino': return CuppertinoHttpClient();
>     ...
>     default: return IOClient();
>   }
> }
> ```
> 
> Again, it solves isolate use case as well.
> 
> If we also wanted to support 3rd party clients, maybe a more flexible way to generate above code could work.

I do want to support 3rd party clients. There is already at least one 3rd party client and I'm hoping that, in the long run, the popular clients will be written by 3rd parties and we won't have to maintain them.

It also doesn't solve the configuration problem. `cupertino_http` offers [three basic configuration options and 14 sub options](https://pub.dev/documentation/cupertino_http/latest/cupertino_http/URLSessionConfiguration-class.html). Same with `cronet_http` and `IOClient` (you can configure all of the options available on `HttpClient`).

Losing configuration in isolates seems at least as bad as getting the wrong client implementation.

If we wanted to offer users a lowest common denominator set of configuration options and pick the `Client` for the user then we should use the approach designed for this: federated plugins.  

> I think the solution proposed here may simply not work very well (i.e. people may not be able to flip on the constant to allow tree shaking) and it increases probability of using multiple different http clients if not all of the app and all isolates use the same zone-injected client.

I agree that it sucks but I'm not sure how to fix it without reducing the capabilities of these clients. I can definitely mention the problem with isolates in the docs. Maybe suggesting a pattern like:

```
Client Function() getClient() {
  if (Platform.isIOS || Platform.isMacOS) {
    return CupertinoClient.defaultSessionConfiguration.call;
  } ...
}

void main() {
  runWithClient(() => runApp(const MyFlutterApp()), getClient());
}

void someFunction() {
  Isolate.run(() => runWithClient(<actual isolate function>, getClient()));
}
 ```

PR #869, opened 2023-02-11
Remove final for headers in BaseRequest

hlwhl:
Current headers in BaseRequest is a final porperty, users can not modify headers of a specfic request (especially to users who use send), because of this, we should make BaseRequest as a non final porperty to give users a chance to modify headers.


kevmoo:
Can you not read and write individual values?

hlwhl:
> Can you not read and write individual values?

@kevmoo Sorry, didn't noticed that, I'll close the PR.

PR #874, opened 2023-02-21
Add a link to cronet_http_embedded from cronet_http.

brianquinlan:


brianquinlan:
> Do we do any rewriting of the README during publish for the alternate package? Should we?

We have a separate README.md for the alternate package:
https://github.com/dart-lang/http/blob/master/pkgs/cronet_http/README_EMBEDDED.md

PR #889, opened 2023-03-09
Add conformances test that verify that the Client works in Isolates

brianquinlan:


pkgs/http_client_conformance_tests/lib/src/isolate_test.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:isolate';
```
natebosch, line 5: Does this import break test compatibility on the web? I would expect to need conditional imports for this.

natebosch: Apparently we don't block the import when we compile in the test runner, but we will block the import in google3 and with `build_web_compilers`.

I think we should fix that, but probably can't until we are ready for a major version bump. https://github.com/dart-lang/test/issues/1968

brianquinlan: Fixed.

natebosch (approved):

Issue #789, opened 2022-09-09
Issue Decoding Special Characters (response.body)


lcsvcn:
Issue decoding special characters such as:

ã --> didn't test with any other special chars, but should be happening too. Such as ç, õ, á, à, ó, í, and so on...

# Code: 

```
final response = await http.get(
      serverUrl,
      ),
      headers: {
         "accept": "application/json;",
      },
    );
  
   print(response.body);  // prints { "title":"NÃ£o encontrado" } -- should be {"title":"Não encontrado"}
```
# Expected:
`
  {  
  "title": "Não encontrado"
}
`

# Reality:
`  {  
"title":"NÃ£o encontrado"
}
`

lcsvcn:
Found a workaround:

# Use: 
 `   json.decode(utf8.decode(response.bodyBytes))}`
    
# Instead of:
` json.decode(response.body); `

It works like a charm!

utkudenis:
I have a problem in this function, it was working 0.13.4 but now, it cannot work , 
![image](https://user-images.githubusercontent.com/68633376/202812164-45652517-42c7-43f6-ab98-740abeb1e7c4.png)

medz:
https://github.com/dart-lang/http/blob/master/pkgs/http/lib/src/utils.dart#LL28C67-L28C73
<img width="658" alt="image" src="https://user-images.githubusercontent.com/5564821/226617817-fdcba068-bdfa-40f0-b641-33402b42342f.png">
<img width="1156" alt="image" src="https://user-images.githubusercontent.com/5564821/226618052-6e24bdc1-b59a-46b1-be52-606cf6b14b88.png">
so retarded

alperefesahin:
> 

thanks, it solves my Turkish character (ş, ı,ç etc) problems

alperefesahin:
> Found a workaround:
> 
> # Use:
> ` json.decode(utf8.decode(response.bodyBytes))}`
> 
> # Instead of:
> `json.decode(response.body);`
> 
> It works like a charm!
> 
> I still think that the `response.body` should be able to decode any UTF-8 characters, so the issue exists, maybe some of the http library developers can confirm.
> 
> Because in my understanding, even with request headers {"charset": "utf-8"} it doesn`t decode from utf-8 what is kind of strange behaviour.

thanks, it solves my Turkish character (ş, ı,ç etc) problems

abdoutech93:
the solution for me was this : https://pub.dev/packages/html_unescape

```
[...]

PR #895, opened 2023-03-21
fix: Default decode any UTF-8 characters

medz:
Refs: #789


natebosch:
This is a breaking behavior change.

PR #920, opened 2023-05-03
Add "base", "final" and "interface" modifiers

brianquinlan:
Fixes https://github.com/dart-lang/http/issues/918

### DIFF
--- a/pkgs/http/lib/retry.dart
+++ b/pkgs/http/lib/retry.dart
@@ -14,7 +14,7 @@ import 'http.dart';
 /// NOTE: [RetryClient] makes a copy of the request data in order to support
 /// resending it. This can cause a lot of memory usage when sending a large
 /// [StreamedRequest].
-class RetryClient extends BaseClient {
+base class RetryClient extends BaseClient {
   /// The wrapped client.
   final Client _inner;
 
--- a/pkgs/http/lib/retry.dart
+++ b/pkgs/http/lib/retry.dart
@@ -14,7 +14,7 @@ import 'http.dart';
 /// NOTE: [RetryClient] makes a copy of the request data in order to support
 /// resending it. This can cause a lot of memory usage when sending a large
 /// [StreamedRequest].
-class RetryClient extends BaseClient {
+final class RetryClient extends BaseClient {
   /// The wrapped client.
   final Client _inner;
 
--- a/pkgs/http/lib/retry.dart
+++ b/pkgs/http/lib/retry.dart
@@ -14,7 +14,7 @@ import 'http.dart';
 /// NOTE: [RetryClient] makes a copy of the request data in order to support
 /// resending it. This can cause a lot of memory usage when sending a large
 /// [StreamedRequest].
-base class RetryClient extends BaseClient {
+class RetryClient extends BaseClient {
   /// The wrapped client.
   final Client _inner;
 
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -31,7 +31,7 @@ BaseClient createClient() {
 /// [BaseRequest.followRedirects], and [BaseRequest.maxRedirects] fields. It is
 /// also unable to stream requests or responses; a request will only be sent and
 /// a response will only be returned once all the data is available.
-base class BrowserClient extends BaseClient {
+class BrowserClient extends BaseClient {
   /// The currently active XHRs.
   ///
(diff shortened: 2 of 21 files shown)


pkgs/http/lib/retry.dart
```diff
@@ -14,7 +14,7 @@ import 'http.dart';
 /// NOTE: [RetryClient] makes a copy of the request data in order to support
 /// resending it. This can cause a lot of memory usage when sending a large
 /// [StreamedRequest].
-class RetryClient extends BaseClient {
+final class RetryClient extends BaseClient {
```
brianquinlan, line 17: Interface is identical to client. There is no reason to implement or extend this.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -31,7 +31,7 @@ BaseClient createClient() {
 /// [BaseRequest.followRedirects], and [BaseRequest.maxRedirects] fields. It is
 /// also unable to stream requests or responses; a request will only be sent and
 /// a response will only be returned once all the data is available.
-class BrowserClient extends BaseClient {
+base class BrowserClient extends BaseClient {
```
brianquinlan, line 34: Interface is identical to Client

pkgs/http/lib/src/io_client.dart
```diff
@@ -43,7 +43,7 @@ class _ClientSocketException extends ClientException
 }
 
 /// A `dart:io`-based HTTP client.
-class IOClient extends BaseClient {
+base class IOClient extends BaseClient {
```
brianquinlan, line 46: Interface is identical to Client

pkgs/http/test/io/client_test.dart
```diff
@@ -14,14 +14,14 @@ import 'package:test/test.dart';
 
 import '../utils.dart';
 
-class TestClient extends http.BaseClient {
+base class TestClient extends http.BaseClient {
```
brianquinlan, line 17: Interface is identical to Client

pkgs/http/test/io/http_test.dart
```diff
@@ -10,7 +10,7 @@ import 'package:test/test.dart';
 
 import '../utils.dart';
 
-class TestClient extends http.BaseClient {
+base class TestClient extends http.BaseClient {
```
brianquinlan, line 13: Interface is identical to Client

brianquinlan:
I didn't add `base` to `BaseRequest`, `BaseClient` and `BaseResponse` but someone might want to `implement` (e.g. for mocking) an implementor of those classes.

pkgs/http/lib/src/browser_client.dart
natebosch: I wonder if this should be `final`?

brianquinlan: I could imagine people wanting to add methods to `Client` implementations. Like:

````dart
class DownloadingBrowserClient extends BrowserClient {
  Stream<UInt8List> downloadFile(Url) {
     ...
  }

}```

**Review by natebosch — APPROVED** [review]

File: pkgs/http/test/io/client_test.dart [hunk 6]
```diff
@@ -14,14 +14,14 @@ import 'package:test/test.dart';
...
   @override
   Future<http.StreamedResponse> send(http.BaseRequest request) {
     throw UnimplementedError();
   }
 }
 
-class TestClient2 extends http.BaseClient {
+base class TestClient2 extends http.BaseClient {
```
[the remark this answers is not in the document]
lrhn: Ditto. Just no modifier. It's not a public class, so we don't care what people can do with it.

pkgs/http/lib/src/byte_stream.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 /// A stream of chunks of bytes representing a single piece of data.
-class ByteStream extends StreamView<List<int>> {
+final class ByteStream extends StreamView<List<int>> {
```
lrhn, line 10: (This class should not need to exist. It should just be extension methods on `Stream<List<int>>`.)

lrhn (approved):
The `BaseClient` class is documented as "This is a mixin-style class" and it can be used as a mixin today, so I'd recommend making it a `mixin class`.

LGTM without the `base`s.

(The package could generally use some clean-up. A lot of the code seems overly complicated, and could use more modern features or just be optimized better.)

Issue #916, opened 2023-05-01
Support web sockets


brianquinlan:
See https://github.com/dart-lang/sdk/issues/41376#issuecomment-1525036848

PR #921, opened 2023-05-03
Support the NSURLSession WebSocket API

brianquinlan:
First step in fixing https://github.com/dart-lang/http/issues/916

### DIFF
--- a/pkgs/cupertino_http/lib/src/cupertino_api.dart
+++ b/pkgs/cupertino_http/lib/src/cupertino_api.dart
@@ -501,27 +501,44 @@ enum URLSessionTaskState {
   urlSessionTaskStateCompleted,
 }
 
-/// A WebSocket
+/// A WebSocket message.
 ///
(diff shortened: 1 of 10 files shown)


pkgs/cupertino_http/lib/src/cupertino_api.dart
```diff
@@ -664,6 +721,108 @@ class URLSessionDownloadTask extends URLSessionTask {
...
+    if (reason == null) {
+      return null;
+    } else {
+      return Data._(reason);
+    }
+  }
+
+  /// Sends a single WebSocket message.
+  ///
+  /// Throws an [Error] on failure.
```
liamappelbe, line 754: Nit: Strictly speaking, it completes the Future with an error. `completeError` only becomes a throw when the user awaits the Future. If they use the `then` method instead, they can add an `onError` handler.

liamappelbe (approved):

pkgs/cupertino_http/lib/src/cupertino_api.dart
brianquinlan: Done!

PR #965, opened 2023-06-16
Make StreameRequest.sink a StreamSink

### LINKED ISSUE #727: StreamedRequest.sink should be a StreamSink
Right now `StreamedRequest.sink` is declared as an `EventSink<List<int>>`:

```dart
  /// The sink to which to write data that will be sent as the request body.
  ///
  /// This may be safely written to before the request is sent; the data will be
  /// buffered.
  ///
  /// Closing this signals the end of the request.
  EventSink<List<int>> get sink => _controller.sink;
```

However, the sink of a `StreamController` is a child class `StreamSink`, which has useful methods like `addStream`: https://api.dart.dev/stable/2.17.5/dart-async/StreamConsumer/addStream.html

Being able to add a stream using this method is essential if you need the stream to be paused when the send buffer is full, e.g. if you are reading from disk faster than you can send it over the network.

Here's a simple example showing how a StreamController is notified when the send buffer is full:

```dart
void main() async {
  final server = await HttpServer.bind('127.0.0.1', 8456);
  server.listen((event) {
    print('server got client');
    final sub = event.listen((event) {});
    // Pause it on the server end, Dart will no longer read from the socket
    // causing new data to be buffered by the OS
    sub.pause();
  });

  final request = StreamedRequest('post', Uri.parse('http://127.0.0.1:8456'));

  final controller = StreamController<List<int>>(
    onPause: () {
      print('paused');
    },
    onResume: () {
      print('resumed');
    },
  );

  Timer.periodic(
    const Duration(seconds: 1),
    (timer) {
      if (controller.isPaused) {
        print('not sending (paused)');
        return;
      }
      print('sending 512K');
      controller.add(Uint8List(512 * 1024));
    },
  );

  // Cast request.sink to a StreamSink so we can call addStream
  (request.sink as StreamSink<List<int>>).addStream(controller.stream);

  await request.send();
}
```

Outputs:

```dart
paused
resumed
sending 512K
paused
server got client
resumed
sending 512K
paused
resumed
sending 512K
paused
resumed
sending 512K
paused
resumed
sending 512K
paused
resumed
sending 512K
paused
not sending - paused
not sending - paused
not sending - paused
not sending - paused
not sending - paused
```

My proposal is just a 1 line change to the getter:

```diff
-  EventSink<List<int>> get sink => _controller.sink;
+  StreamSink<List<int>> get sink => _controller.sink;
```

This would also allow us to get upload progress events without adding any callbacks to post / send:

```dart
final progress = ValueNotifier<int>(0);
request.sink.addStream(
  file.openRead().map((e) {
    progress.value += e.length;
    return e;
  }),
);
```

natebosch:
Closes #727

This class was not designed for extension, but there is at least one
extending implementation. I don't see any overrides of this field, so no
existing usage should be broken.

### DIFF
--- a/pkgs/http/lib/src/streamed_request.dart
+++ b/pkgs/http/lib/src/streamed_request.dart
@@ -24,13 +24,10 @@ import 'byte_stream.dart';
 ///
 /// // The sink must be closed to end the request.
 /// // The Future returned from `close()` may not complete until after the
-/// request is sent, an it generally should not be awaited.
+/// // request is sent, and it should not be awaited.
 /// unawaited(request.sink.close());
 /// final response = await request.send();
 /// ```
-///
-/// The `Future` returned from calling `sink.close()` will not complete until
-/// the request has been sent.
 class StreamedRequest extends BaseRequest {
   /// The sink to which to write data that will be sent as the request body.
   ///
--- a/pkgs/http/lib/src/streamed_request.dart
+++ b/pkgs/http/lib/src/streamed_request.dart
@@ -20,11 +20,17 @@ import 'byte_stream.dart';
 /// ```dart
 /// final request = http.StreamedRequest('POST', Uri.http('example.com', ''))
 ///     ..contentLength = 10
-///     ..sink.add([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
-///     ..sink.close();  // The sink must be closed to end the request.
+///     ..sink.add([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
 ///
+/// // The sink must be closed to end the request.
+/// // The Future returned from `close()` may not complete until after the
+/// request is sent, an it generally should not be awaited.
+/// unawaited(request.sink.close());
 /// final response = await request.send();
 /// ```
+///
+/// The `Future` returned from calling `sink.close()` will not complete until
+/// the request has been sent.
 class StreamedRequest extends BaseRequest {
   /// The sink to which to write data that will be sent as the request body.
   ///
--- a/pkgs/http/lib/src/streamed_request.dart
+++ b/pkgs/http/lib/src/streamed_request.dart
@@ -32,7 +32,7 @@ class StreamedRequest extends BaseRequest {
   /// buffered.
   ///
   /// Closing this signals the end of the request.
-  EventSink<List<int>> get sink => _controller.sink;
+  StreamSink<List<int>> get sink => _controller.sink;
 
   /// The controller for [sink], from which [BaseRequest] will read data for
   /// [finalize].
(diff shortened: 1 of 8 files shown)


pkgs/http/lib/src/streamed_request.dart
```diff
@@ -32,7 +32,7 @@ class StreamedRequest extends BaseRequest {
   /// buffered.
   ///
   /// Closing this signals the end of the request.
-  EventSink<List<int>> get sink => _controller.sink;
+  StreamSink<List<int>> get sink => _controller.sink;
```
brianquinlan, line 35: I think that you need to update the example in the documentation for this class.

pkgs/http/test/io/client_test.dart
```diff
@@ -81,9 +80,8 @@ void main() {
       ..headers[HttpHeaders.userAgentHeader] = 'Dart';
 
     var responseFuture = client.send(request);
-    request
-      ..sink.add('{"hello": "world"}'.codeUnits)
-      ..sink.close();
+    request.sink.add('{"hello": "world"}'.codeUnits);
+    await request.sink.close();
```
brianquinlan, line 84: None of the `await`s in the tests are necessary, right? Otherwise this would be a breaking change.

pkgs/http/lib/src/streamed_request.dart
brianquinlan, line 35: I think that you need to update the example in the documentation for this class.

brianquinlan:
> Closes #727
> 
> This class was not designed for extension, but there is at least one extending implementation. I don't see any overrides of this field, so no existing usage should be broken.

Yeah, this change is technically breaking. If we assume that it is non-breaking because no one implements this class then you could change it to `base` at the same time (I think).

natebosch:
cc @jakemac53 - this change isn't directly breaking, but with a typical linting setup that uses "unwaited_futures" the change  pushes for edits that would cause breakage - with the IO client the `sink.close()` and `request.send()` calls return cross-dependent futures. The call to `close()` returns a Future that cannot complete until the request is sent, and the call to `send()` returns a Future that cannot complete until the sink has been closed.

pkgs/http/lib/src/streamed_request.dart
```diff
@@ -20,19 +20,25 @@ import 'byte_stream.dart';
 /// ```dart
 /// final request = http.StreamedRequest('POST', Uri.http('example.com', ''))
 ///     ..contentLength = 10
-///     ..sink.add([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
-///     ..sink.close();  // The sink must be closed to end the request.
+///     ..sink.add([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
 ///
+/// // The sink must be closed to end the request.
+/// // The Future returned from `close()` may not complete until after the
+/// request is sent, an it generally should not be awaited.
```
brianquinlan, line 27: , and it should not...

brianquinlan (approved):

jakemac53 (approved):

PR #819, opened 2022-11-11
Add ability to close a StreamedRequest

jpopesculian:
Adds an extra `close` function to `StreamedRequest` and implements it in `IoStreamedRequest` as well as adds a `BrowserStreamedRequest` for implementing it

Issue #595, opened 2021-07-15
Migrate to the `fetch` API


natebosch:
dart2js will stop supporting internet explorer soon - so all supported browser should support `fetch` which allows for streaming. I'm not sure if streaming works on firefox so that will need to be tested.

https://api.dart.dev/stable/2.13.4/dart-html/Window/fetch.html

https://github.com/dart-lang/http/blob/f93c76fabdb03963303762e3fe16cbdf60799cff/lib/src/browser_client.dart#L27-L29

pedia:
How about this since one year past?

jtmcdole:
This would be a good feature request for us. Happy to take a look and see how well it maps.

natebosch:
Example of using `fetch` in Dart: https://github.com/dart-lang/dartdoc/blob/f0d2cbf5f517d936d01a61e7350147e3ad12edb9/web/search.dart#L35

pedia:
We need cross all platform `fetch`.  e.g long polling http request
It's called in package:http/http.

srix55:
Looks like we need to fallback on websockets. Flutter web's http (BrowserClient) has no streaming capability. EventSources have their own issues (no http headers, for starters). That's a huge minus. Fetch, QUIC, web-transport, http3 are still years away. websockets it is.

jtmcdole:
> Looks like we need to fallback on websockets. Flutter web's http (BrowserClient) has no streaming capability. EventSources have their own issues (no http headers, for starters). That's a huge minus. Fetch, QUIC, web-transport, http3 are still years away. websockets it is.

what about web-rtc?

kaoet:
Bump this thread. There is no XHR in Chrome manifest v3 service workers so `fetch` must be used there instead of XHR. https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#background-service-workers

momrak:
Is there any chance of this happening any time soon, or what is the status? Are there any other reasons to keep using XHR now that IE is dead?

Zekfad:
I've come across the issue that XHR missing information about redirects.
I've made fetch binding and custom Client on userland, so you can use my [fetch_client](https://pub.dev/packages/fetch_client) as a workaround.

Deishelon:
Any updates on this?

AlejandroArciniegas:
How can I help moving this forward?

Dampfwalze:
[...]

PR #978, opened 2023-07-05
Support Request Cancellation, Timeouts via RequestController

### LINKED ISSUE #204: Cancellable requests

### LINKED ISSUE #424: Design an API to support aborting requests
On the web `HttpRequest` supports `abort`: https://api.dart.dev/stable/2.8.1/dart-html/HttpRequest/abort.html

We have a proposal for adding abort on the `dart:io` `HttpClientRequest`: https://dart-review.googlesource.com/c/sdk/+/147339

We need to come up with a design for supporting `abort` through the interfaces in this package and validate that we can build it on both `dart:html` and `dart:io` implementations.

The easiest option might be to add a method on `StreamedResponse`. There is some precedent for that with `IOStreamedResponse.detachSocket`. https://github.com/dart-lang/http/blob/9063ba379b7e7386cced830feb27cc7a5fd3ec7e/lib/src/io_streamed_response.dart#L36 (we need to make sure the differences between these are clearly documented.

I do wonder if we need to consider some sort of support for the more convenient methods than `send` though...

cc @kevmoo @jakemac53 @grouma @zichangg @lrhn for thoughts

### LINKED ISSUE #567: Abort or cancel multipart request

SamJakob:
Closes #204, closes #424, closes #567
Supersedes and closes #521, supersedes and closes #602, supersedes and closes #819

This PR implements an opt-in `RequestController` API to manage requests. The API is opt-in, future-proof and backwards-compatible both for implementations of `Client` and for end users. If a controller is not specified by the user, the existing functionality of the package is maintained. Likewise, clients can opt-in to the new API and show they are doing so by overriding `supportsController` to be true and implementing the functionality. Later functionality of a similar nature can also be added in an opt-in manner by simply adding it to the `RequestController` class.

See the full examples below, but a brief tour of the RequestController's functionality is as follows:
```dart
// Create a request controller, can be passed into multiple requests to
// cancel them all at once or to set timeouts on all of them at once.
// Or, one can be created per request.
final controller = RequestController(
  // Optional timeout for the overall request (entire round trip).
  timeout: Duration(seconds: 5),
  // Optional timeouts for each state of the request.
  // If a state is not specified, it will not timeout.
  partialTimeouts: PartialTimeouts(
    // Timeout for connecting to the server.
    connectTimeout: Duration(seconds: 5),
    // Timeout for the request body to be sent and a response to become
    // available from the server.
    sendTimeout: Duration(seconds: 4),
    // Timeout for processing the response body client-side.
    receiveTimeout: Duration(seconds: 6),
  ),
);
```

It can then be used as follows:
```dart
final response = await client.get(
  // URL that hangs for 5 seconds before responding.
  Uri.parse('http://localhost:3456/longrequest'),
  controller: controller,
);
// Unhandled exception:
// TimeoutException after 0:00:04.000000: Future not completed

// or,
controller.cancel(); // Causes request future to throw CancelledException
```

I note, from above, that there was existing work in both request cancellation and timeouts however I felt that this different approach could cleanly solve both issues with a single, more ergonomic, user-friendly and intuitive API. This work started originally as a private fork of the package to implement some functionality required for another package, however I felt that the results might be viable for inclusion in the upstream package.

---

Additionally, with respect to #819, I happened to implement this similarly to make it possible to abort a response of an `IOClient` (useful for applications that need to interrupt a response with a large payload). For this, I created a `ClosableStreamedResponse` that extends `StreamedResponse` with an additional `close` future.

This should only be necessary for the `IOStreamedResponse` (for which it is implemented with detachSocket and destroy which has been tested and works). The browser client appears to obtain the entire buffered response from the browser (via the `onLoad` event) and return that, so there's no sense or utility in closing that response - it can be aborted with `xhr.abort`.

For `ClosableStreamedResponse`s to be cancellable with the controller, `Response.fromStream` grabs the `ActiveRequestTracker` for the response's request (if there is one) and, if that controller has `cancel` called on it, closes the response stream if and only if the response's stream is a ClosableStreamedResponse. (See lines 70-78 of `response.dart`).

---

## Examples

<details>
<summary><b>Partial Timeouts (for each step of the request-response cycle)</b></summary>

<br>

```dart

Future<void> main() async {
  Client client = Client();

  // Create a request controller, can be passed into multiple requests to
  // cancel them all at once or to set timeouts on all of them at once.
  // Or, one can be created per request.
  final controller = RequestController(
    // Optional timeout for the overall request (entire round trip).
    timeout: Duration(seconds: 5),
    // Optional timeouts for each state of the request.
    // If a state is not specified, it will not timeout.
    partialTimeouts: PartialTimeouts(
      // Timeout for connecting to the server.
      connectTimeout: Duration(seconds: 5),
      // Timeout for the request body to be sent and a response to become
      // available from the server.
      sendTimeout: Duration(seconds: 4),
      // Timeout for processing the response body client-side.
      receiveTimeout: Duration(seconds: 6),
    ),
  );

  final response = await client.get(
    // URL that hangs for 5 seconds before responding.
    Uri.parse('http://localhost:3456/longrequest'),
    controller: controller,
  );

  // Unhandled exception:
  // TimeoutException after 0:00:04.000000: Future not completed
  print(response.body);
}
```

_In this case, the send timeout triggers a `TimeoutException` because no response is received from the server._

Server-side logs:
```
04:29:58.608 PM [Request (/longrequest)]        resume
04:29:58.610 PM [Request (/longrequest)]        readable
04:29:58.610 PM [Request (/longrequest)]        end
04:29:58.610 PM [Request (/longrequest)]        close
04:30:02.618 PM [Response (/longrequest)]       close
```

</details>

<details>
<summary><b>Partial Timeouts (another example)</b></summary>

<br>

```dart

Future<void> main() async {
  Client client = Client();

  // Create a request controller, can be passed into multiple requests to
  // cancel them all at once or to set timeouts on all of them at once.
  // Or, one can be created per request.
  final controller = RequestController(
    // Optional timeout for the overall request (entire round trip).
    timeout: Duration(seconds: 10),
    // Optional timeouts for each state of the request.
    // If a state is not specified, it will not timeout.
    partialTimeouts: PartialTimeouts(
      // Timeout for connecting to the server.
      connectTimeout: Duration(seconds: 5),
      // Timeout for the request body to be sent and a response to become
      // available from the server.
      sendTimeout: Duration(seconds: 4),
      // Timeout for processing the response body client-side.
      receiveTimeout: Duration(seconds: 6),
    ),
  );

  final response = await client.get(
    // URL that immediately writes headers and a partial body, then hangs for
    // 5 seconds before responding with the remainder.
    Uri.parse('http://localhost:3456/longpoll'),
    controller: controller,
  );

  // Hello, world! (the payload sent immediately)
  // Hello, world! (the payload sent after 5 seconds)
  print(response.body);
}
```

_In this case, none of the timeouts trigger an exception because the response body becomes available immediately, even if it is not complete and the server hangs for only 5 seconds before completing the response, meaning the receive timeout of 6 seconds is not triggered either._

Server-side logs:
```
04:37:36.225 PM [Request (/longpoll)]   resume
04:37:36.226 PM [Request (/longpoll)]   readable
04:37:36.226 PM [Request (/longpoll)]   end
04:37:36.226 PM [Request (/longpoll)]   close
04:37:41.230 PM [Response (/longpoll)]  finish
04:37:41.230 PM [Request (/longpoll)]   resume
04:37:41.231 PM [Response (/longpoll)]  close
```

</details>

<details>
<summary><b>Cancelation Example (cancel during response)</b></summary>

<br>

```dart

Future<void> main() async {
  Client client = Client();

  // Create a request controller, can be passed into multiple requests to
  // cancel them all at once or to set timeouts on all of them at once.
  // Or, one can be created per request.
  final controller = RequestController();

  final responseFuture = client.get(
    // URL that hangs for 5 seconds before responding.
    Uri.parse('http://localhost:3456/longrequest'),
    controller: controller,
  );

  await Future.delayed(Duration(milliseconds: 500));
  controller.cancel();

  final response = await responseFuture;
  print(response.body);

  // Unhandled exception:
  // CancelledException: Request cancelled
  print(response.body);
}
```

_Cancels after 500ms which is whilst the response is being written. The connection is destroyed so both the client and server clean up immediately after the request is cancelled._

Server-side logs:
```
04:41:11.651 PM [Request (/longrequest)]        resume
04:41:11.651 PM [Request (/longrequest)]        readable
04:41:11.652 PM [Request (/longrequest)]        end
04:41:11.652 PM [Request (/longrequest)]        close
04:41:12.137 PM [Response (/longrequest)]       close
```

</details>

<details>
<summary><b>Cancellation Example (cancelled immediately)</b></summary>

<br>

```dart

Future<void> main() async {
  Client client = Client();

  // Create a request controller, can be passed into multiple requests to
  // cancel them all at once or to set timeouts on all of them at once.
  // Or, one can be created per request.
  final controller = RequestController();

  final responseFuture = client.get(
    // URL that hangs for 5 seconds before responding.
    Uri.parse('http://localhost:3456/longrequest'),
    controller: controller,
  );

  controller.cancel();

  final response = await responseFuture;
  print(response.body);

  // Unhandled exception:
  // CancelledException: Request cancelled
  print(response.body);
}
```

_The request is cancelled immediately and isn't even logged on the server._

Server-side logs:
```
(none)
```

</details>

---

Server implementation and Dart client example used to test:
https://github.com/SamJakob/cancellable_http_dart_testing

---

<details>

</details>


pkgs/http/lib/src/client.dart
```diff
@@ -34,6 +35,13 @@ import 'streamed_response.dart';
 /// another instance of [Client] and add functionality on top of that. This
 /// allows all classes implementing [Client] to be mutually composable.
 abstract interface class Client {
+  /// Whether this client supports the [RequestController] API.
+  ///
+  /// If this is `true`, [send] will use the supplied [RequestController] and
+  /// will allow cancelling the request and specifying a timeout. Otherwise,
+  /// a specified [RequestController] will be ignored.
+  bool get supportsController;
```
natebosch, line 43: This is a breaking change for `implements Client` uses. For example https://github.com/google/googleapis.dart/blob/ef216bc000480b20d8a08776343bd2f8eacd423d/googleapis_auth/lib/src/auth_client.dart#L10

natebosch (commented):
This is an interesting API.

What are your use cases for being able to independently control parts of the timeout?

I don't think we'll be able to move forward with this PR in the short term. It will take a deeper review to see if this API is in the direction we'd want to go with cancellation.

cc @brianquinlan

pkgs/http/lib/src/client.dart
SamJakob: Oops - as was already the case with `BaseClient` that was supposed to return `false` by default to allow clients to opt-in.

natebosch: Adding the API at all is breaking. Any class which `implement Client` without `extends BaseClient` will be missing the override.

SamJakob:
I think generally the ability to control timeouts over individual parts of the lifecycle would be beneficial for specific use cases - particularly those dealing with large payloads - off the top of my head, some examples could be:
- to abort a connection if connecting to the server takes too long, but not once the initial connection has been made.
- to trigger a timeout only once the response is being delivered (for example, uploading a large payload that takes - say 30 seconds - to upload and subsequently alotting a time frame in which to start receiving a response).

These could be used similarly to the `headerTimeout` and `contentTimeout` discussed in #521.

Also, `sendTimeout` refers to the point in the lifecycle where the response headers are delivered because that I considered to be the end of the `sending` phase, but maybe it's more intuitive to call that `headersTimeout` and have `responseTimeout` be `contentTimeout`.

I would imagine it is most useful for things like `StreamedResponse` where you could set a timeout for the initial connection and response headers (`connectTimeout` and `sendTimeout`) but not for the response or the request as a whole.

Also, the lifecycle is tracked implicitly for the purposes of properly supporting cancellation in any client implementing the `RequestController` API, so I felt it useful to expose the timeouts for each of those stages too.

pkgs/http/lib/src/client.dart
SamJakob: Oh yes, of course. I'll do some testing against something like the linked client (`AuthClient`) for that then - I didn't initially find any clients that implemented `Client` so I presumed they would use `BaseClient` like `ReplyClient` does.

That might need further review then, but off the top of my head a backwards-compatible solution could be to introduce some stub that could be mixed in or included and tested for.

For example: https://dartpad.dev/bc9a693d707f95a0356291817f38db1e

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,8 +56,107 @@ class BrowserClient extends BaseClient {
...
+            onCancel: handleCancel,
+          ));
+        }
+
+        xhr.onReadyStateChange.listen((_) {
+          // If the connection is at the OPENED stage and the
+          // connectCompleter has not yet been marked as completed, complete it.
+          if (xhr.readyState == HttpRequest.OPENED) {
+            if (connectCompleter != null) {
+              connectCompleter.complete();
```
brianquinlan, line 115: Does HttpRequest.OPENED really imply that a connection has been established? I thought that it just meant that `.open()` has been called.

brianquinlan:
I think that `partialTimeouts` is a pretty complex API to implement and I'm not sure that it can be implemented completely on any `Client` other than `IOClient`.

Would it make sense to ask people who need these more advanced features to use `package:dio`? Though I'm not sure if their timeout, cancellation and progress support works [if adapting from a `package:http` `Client`](https://github.com/cfug/dio/blob/main/plugins/native_dio_adapter/lib/src/conversion_layer_adapter.dart).

pkgs/http/lib/src/browser_client.dart
SamJakob: No, it doesn't ([as suggested here on StackOverflow](https://stackoverflow.com/questions/58324522/how-to-check-that-xmlhttprequest-connection-was-established-before-data-arrive)) so that probably shouldn't be implemented in the BrowserClient.

That said, whilst looking into the browser side of things, I saw that the new fetch API ([caniuse: fetch API](https://caniuse.com/?search=fetch)) supports streams ([caniuse: ReadableStream](https://caniuse.com/?search=ReadableStream)) and could probably be used to implement this functionality as it is on the Dart side.

What kind of support is being aimed for on the browser side? Is there a reason XMLHTTPRequest is still being used or might it be worth looking at moving to the fetch API? - See #595

SamJakob:
I agree that it is complex and maybe there is a point to be made that its complexity falls outside the scope of the package. I had some reservations about `partialTimeouts` with respect to testing too. That said, I replied to your review comment to address implementing it in `BrowserClient` with a streaming implementation of the `fetch` API that could allow for implementation of both platforms in a fairly similar way.

Also, as mentioned before, tracking the lifecycle in a similar manner would still be essential (at least on `dart:io`) to ensure the request is cancelled properly on certain platforms. It would just be a matter of determining the costs/benefits of exposing timeouts over it as an API.

I think the major benefit of this package is that it encapsulates all of the platform-specific implementations and exposes a single consistent API so as long as it is done consistently I would advocate including the feature in this package because it would save anyone needing a cross-platform HTTP solution that also requires control over `partialTimeouts` from having to re-implement each platform implementation.

JaffaKetchup:
Just seen that this has been re-opened! 🎉

I've been wondering whether an approach without a dedicated controller could work.

I'm not sure why the timeouts would be best in the controller, apart from the advantage that the controller could be reused, but what if the HTTP method took a future, that, if completed, cancelled the request. Users could then control this through a standard `Completer`. (This pattern is similar to what we offer to users in a package I maintain). This wouldn't be reusable, but maybe is a more fluent way to express this? I guess it's down to opinion though.

Regardless, hoping this can be merged as relying on Dio for this is a bit annoying!

(Separately, I was wondering whether aborting HTTP is possible in WASM?)

PR #993, opened 2023-07-21
Harmonize response header behavior

brianquinlan:
- Uniform whitespace handing
- Uniform content-length handling
---

<details>

</details>

### DIFF
--- a/pkgs/cronet_http/lib/src/cronet_client.dart
+++ b/pkgs/cronet_http/lib/src/cronet_client.dart
@@ -269,14 +269,15 @@ class CronetClient extends BaseClient {
     final contentLengthHeader = responseHeaders['content-length'];
     int? contentLength;
     if (contentLengthHeader != null) {
-      contentLength = int.tryParse(contentLengthHeader);
-      if (contentLength == null || contentLength < 0) {
+      if (!RegExp(r'^\d+$').hasMatch(contentLengthHeader)) {
         throw ClientException(
           'Invalid content-length header [$contentLengthHeader].',
           request.url,
         );
       }
+      contentLength = int.parse(contentLengthHeader);
     }
+
     return StreamedResponse(responseDataController.stream, result.statusCode,
         contentLength: contentLength,
         reasonPhrase: result.statusText,
--- a/pkgs/cronet_http/lib/src/cronet_client.dart
+++ b/pkgs/cronet_http/lib/src/cronet_client.dart
@@ -267,10 +267,18 @@ class CronetClient extends BaseClient {
         .map((key, value) => MapEntry(key.toLowerCase(), value.join(',')));
 
     final contentLengthHeader = responseHeaders['content-length'];
+    int? contentLength;
+    if (contentLengthHeader != null) {
+      contentLength = int.tryParse(contentLengthHeader);
+      if (contentLength == null || contentLength < 0) {
+        throw ClientException(
+          'Invalid content-length header [$contentLengthHeader].',
+          request.url,
+        );
+      }
+    }
     return StreamedResponse(responseDataController.stream, result.statusCode,
-        contentLength: contentLengthHeader == null
-            ? null
-            : int.tryParse(contentLengthHeader),
+        contentLength: contentLength,
         reasonPhrase: result.statusText,
         request: request,
         isRedirect: result.isRedirect,
--- a/pkgs/cronet_http/lib/src/cronet_client.dart
+++ b/pkgs/cronet_http/lib/src/cronet_client.dart
@@ -21,6 +21,7 @@ import 'package:http/http.dart';
 import 'messages.dart' as messages;
 
 final _api = messages.HttpApi();
+final _digitRegex = RegExp(r'^\d+$');
 
 final Finalizer<String> _cronetEngineFinalizer = Finalizer(_api.freeEngine);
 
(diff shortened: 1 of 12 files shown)


pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -267,10 +267,19 @@ class CronetClient extends BaseClient {
         .map((key, value) => MapEntry(key.toLowerCase(), value.join(',')));
 
     final contentLengthHeader = responseHeaders['content-length'];
+    int? contentLength;
+    if (contentLengthHeader != null) {
```
natebosch, line 271: [optional] 
```suggestion
    int? contentLength;
    if (responseHeaders['content-length'] case final contentLengthHeader?) {
```

pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -267,10 +267,19 @@ class CronetClient extends BaseClient {
...
+      if (!RegExp(r'^\d+$').hasMatch(contentLengthHeader)) {
+        throw ClientException(
+          'Invalid content-length header [$contentLengthHeader].',
+          request.url,
+        );
+      }
+      contentLength = int.parse(contentLengthHeader);
```
natebosch, line 278: I was first going to suggest pulling out the RegExp instead of reconstructing it for each request - but I think we can skip the regex entirely.

```suggestion
      contentLength = int.tryParse(contentLengthHeader);
      if (contentLength == null) {
        throw ClientException(
          'Invalid content-length header [$contentLengthHeader].',
          request.url,
        );
      }
```

pkgs/cupertino_http/lib/src/cupertino_client.dart
```diff
@@ -279,6 +279,15 @@ class CupertinoClient extends BaseClient {
       throw ClientException('Redirect limit exceeded', request.url);
     }
 
+    final headers = response.allHeaderFields
+        .map((key, value) => MapEntry(key.toLowerCase(), value));
```
natebosch, line 283: [optional] Another place we _could_ use destructuring and collection literals, but it's not a clear win.

```suggestion
    final headers = {
      for (final MapEntry(:key, :value) in response.allHeaderFields.entries)
        key.toLowerCase(): value
    };
```

pkgs/http/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.1.1
+* `BrowserClient` throws `ClientException` when the `'Content-Length'` header
```
natebosch, line 2:
```suggestion
## 1.1.1

* `BrowserClient` throws `ClientException` when the `'Content-Length'` header
```

pkgs/http/lib/src/browser_client.dart
```diff
@@ -64,6 +64,14 @@ class BrowserClient extends BaseClient {
...
 
     unawaited(xhr.onLoad.first.then((_) {
+      final contentLengthHeader = xhr.responseHeaders['content-length'];
+      if (!RegExp(r'^\d+$').hasMatch(contentLengthHeader ?? '0')) {
+        completer.completeError(ClientException(
+          'Invalid content-length header [$contentLengthHeader].',
+          request.url,
+        ));
+        return;
+      }
```
natebosch, line 74: I would avoid the RegExp check against `'0'` in case of `null`. I'd also consider pulling out the `RegExp` construction so it only happens once.

[optional] Also, you could optionally move the variable to be defined with `case` and make it only in scope within the conditional.

```suggestion
      if (xhr.responseHeaders['content-length'] case final contentLengthHeader?
          when !digitRegex.hasMatch(contentLengthHeader)) {
        completer.completeError(ClientException(
          'Invalid content-length header [$contentLengthHeader].',
          request.url,
        ));
        return;
      }
```

pkgs/cupertino_http/lib/src/cupertino_client.dart
natebosch, line 283: The Dart SDK has traditionally changed headers to lower case and we ended up needing to add an option to _not_ do that, because some servers _do_ care about header case.

See https://github.com/dart-lang/http/issues/609

pkgs/cronet_http/lib/src/cronet_client.dart
brianquinlan: We also need to check that it is non-negative. I used the regex because it matches the production in RFC-2616:
https://datatracker.ietf.org/doc/html/rfc2616#section-14.13

I don't have strong feelings about this though.

pkgs/cupertino_http/lib/src/cupertino_client.dart
brianquinlan: These are response headers. This is just moving some code so that the `'content-length'` header can be examined

pkgs/http/lib/src/browser_client.dart
brianquinlan: Done

pkgs/cupertino_http/lib/src/cupertino_client.dart
brianquinlan: I don't actually think that this is better.

brianquinlan:
Ready for review!

natebosch (approved):

Issue #957, opened 2023-06-05
Create a new package:http Client based on Java APIs


brianquinlan:
This is a tracking issue for Alexander James' gSoC project:
[GSoC Idea](https://github.com/dart-lang/sdk/wiki/Dart-GSoC-2023-Project-Ideas#idea-build-a-dart-http-client-using-java-apis)
[Google Group](https://groups.google.com/g/dart-http-gsoc-2023)
[Calendar](https://calendar.google.com/calendar/embed?src=913904eb9fc214910147b0693d6e561b276d9a07cdc640b6e32392c75745ac65%40group.calendar.google.com&ctz=America%2FLos_Angeles)

PR #995, opened 2023-07-27
[java_http] send request body

alex-james-dev:
This is a PR for the `java_http` [GSoC '23 project](https://summerofcode.withgoogle.com/programs/2023/projects/NKUQqmSA) 🌞.
Relevant tracking issue: https://github.com/dart-lang/http/issues/957.

## Purpose
The purpose of this PR is to make the `package:java_http` client send the request body to the server.
We are now passing another two [package:http_client_conformance_tests](https://github.com/dart-lang/http/tree/b2067710f88980fc0fee43ec3380bce089f001db/pkgs/http_client_conformance_tests) :tada::

1. testIsolate()
2. testRequestBody()

The testRequestBody() tests were initially failing because we can't send a StreamedRequest object between isolates.

## Reviewers
@HosseinYousefi
@natebosch

---

<details>

</details>


pkgs/java_http/lib/src/java_client.dart
```diff
@@ -39,23 +39,29 @@ class JavaClient extends BaseClient {
...
+    final (statusCode, reasonPhrase, responseHeaders, responseBody) =
+        await Isolate.run(() async {
       final httpUrlConnection = URL
-          .ctor3(request.url.toString().toJString())
+          .ctor3(requestUrl.toString().toJString())
           .openConnection()
           .castTo(HttpURLConnection.type, deleteOriginal: true);
 
-      request.headers.forEach((headerName, headerValue) {
+      requestHeaders.forEach((headerName, headerValue) {
```
HosseinYousefi, line 56: Prefer using a normal `for` loop. https://dart.dev/tools/linter-rules/avoid_function_literals_in_foreach_calls

HosseinYousefi (commented):
LGTM

pkgs/java_http/lib/src/java_client.dart
```diff
@@ -78,12 +84,27 @@ class JavaClient extends BaseClient {
...
+    HttpURLConnection httpUrlConnection,
+    Uint8List requestBody,
+  ) {
+    if (requestBody.isEmpty) return;
+
+    httpUrlConnection.setDoOutput(true);
+    final outputStream = httpUrlConnection.getOutputStream();
+    requestBody.forEach(outputStream.write);
+
+    outputStream
```
natebosch, line 97: [optional] Is there any value in minimizing the uses of `callMethodWithArgs` and avoiding interop calls in a loop?

(untested suggestion)

```suggestion
    httpUrlConnection.setDoOutput(true);
    final outputStream = httpUrlConnection.getOutputStream();
    final bodyArray = JArray(jbyte.type, requestBody.length)
      ..setRange(0, requestBody.length, requestBody);
    outputStream
      ..write1(bodyArray)
```

natebosch (approved):

pkgs/java_http/lib/src/java_client.dart
alex-james-dev: Sounds like a good idea to me!

I've pushed a commit with the suggestions and the tests were all passing locally.

Thank you :smile:

alex-james-dev: I've been learning about extension methods recently, could we add an extension method to `Uint8List`?

```dart
extension on Uint8List {
  JArray<jbyte> toJArray() =>
      JArray(jbyte.type, length)..setRange(0, length, this);
}
```

Then, the code would go from:
```dart
    final bodyArray = JArray(jbyte.type, requestBody.length)
      ..setRange(0, requestBody.length, requestBody);

    httpUrlConnection.getOutputStream()
      ..write1(bodyArray)
      ..flush()
      ..close();
```

to:

```dart
    httpUrlConnection.getOutputStream()
      ..write1(requestBody.toJArray())
      ..flush()
      ..close();
```

My intuition is that adding an extension method to `Uint8List` is unnecessary at the moment. But if it becomes a common occurence to convert `Uint8List` to `JArray<jbyte>` in `java_http` then it would be a nice idea to add an extension method.

natebosch: > could we add an extension method to `Uint8List`?

This extension method does look like a good idea.

natebosch (approved):
LGTM as is, or with the discussed extension method `toJArray()`.

pkgs/java_http/lib/src/java_client.dart
alex-james-dev: Awesome! Let's go for the extension method `toJArray()` on `Uint8List`.

alex-james-dev:
Hey @natebosch,

I've pushed a commit which adds the extension method `toJArray()` on `Uint8List`.

I'm new to extension methods and I'm not sure where the best place to put the extension method is.
I ended up taking inspiration from `package:checks` and created an `extensions` directory. 

But I'm open to ideas about where to put the extension method.

Thank you!

natebosch (approved):

natebosch:
> I'm open to ideas about where to put the extension method.

For something very short like this which is used in 1 place, I usually start with a private extension locally within the library. A "utils" or "extensions" directory or library like you chose here is also great.

It's a pretty low-impact decision in this case which is why I approved without commenting. Since you're looking for input on best practices my recommendation would be a private local extension for this one.

```dart
extension on Uint8List {
```

For example:

https://github.com/dart-lang/async/blob/b65622afa33c5bfc574ae6b34d5a61f18a98f83c/lib/src/cancelable_operation.dart#L560

alex-james-dev:
Hey @natebosch, 

Thanks for explaining when to use an unnamed extension and where to store named extensions, I'll definitely remember that for the future :brain:.

I've pushed a commit to make the change to a private extension, I think it looks better. The example you linked was very helpful.

I appreciate your help!

PR #996, opened 2023-07-27
Ignore generated flutter plugin files

natebosch:
These files both start with comments stating that they should not be
edited by hand nor checked into source control. Add them to the
`.gitignore`.

### DIFF
--- a/pkgs/java_http/.gitignore
+++ b/pkgs/java_http/.gitignore
@@ -1,1 +1,3 @@
 build/
+.flutter-plugins
+.flutter-plugins-dependencies


kevmoo:
Should these just be in a global ignore for the user then?

natebosch:
> Should these just be in a global ignore for the user then?

I suppose it could - but couldn't we just as easily say that `.dart_tool/` should be in a global ignore?

These files are ignored because of a property of the project, not because of a property of my personal machine. I lean towards putting them in the project ignore.

These are also included in the flutter project templates:
https://github.com/flutter/flutter/blob/eab2087a0ffe3f2b5949c219bcf3184a0dbd9423/packages/flutter_tools/templates/app_shared/.gitignore.tmpl#L28

But I don't know if that's saying much, because the flutter templates also over-include machine specific things like a bunch of different editors/IDEs and OS specific stuff...

kevmoo (approved):

PR #999, opened 2023-07-31
Add java_http .gitattributes file

alex-james-dev:
This is a PR for the `java_http` [GSoC '23 project](https://summerofcode.withgoogle.com/programs/2023/projects/NKUQqmSA) 🌞.
Relevant tracking issue: https://github.com/dart-lang/http/issues/957.

## Purpose
The purpose of this PR is to make `jnigen` generated code in `java_http` hidden by default in GitHub diffs.
`cronet_http` and `cupertino_http` already hide their generated files in GitHub diffs:
- [`cronet_http`'s .gitattributes](https://github.com/dart-lang/http/blob/1a42b4a16636e86b5fed60666b0de91219c5110c/pkgs/cronet_http/.gitattributes)
- [`cupertino_http`'s .gitattributes](https://github.com/dart-lang/http/blob/1a42b4a16636e86b5fed60666b0de91219c5110c/pkgs/cupertino_http/.gitattributes)

This PR makes `java_http` have the same behaviour as the other `package:http` clients by hiding generated code in GitHub diffs.

## Reviewers
@natebosch

---

<details>

</details>

### DIFF
--- a/pkgs/java_http/.gitattributes
+++ b/pkgs/java_http/.gitattributes
@@ -1,0 +1,2 @@
+# Hides jnigen generated code in GitHub diffs.
+lib/src/third_party/** linguist-generated
--- a/pkgs/java_http/.gitattributes
+++ b/pkgs/java_http/.gitattributes
@@ -1,2 +1,2 @@
-# Hides jnigen generated code in GitHub diffs.
-lib/src/third_party/** linguist-generated
+# Vendored code is excluded from language stats in the GitHub repo.
+lib/src/third_party/** linguist-vendored
--- a/pkgs/java_http/.gitattributes
+++ b/pkgs/java_http/.gitattributes
@@ -1,2 +1,5 @@
 # Vendored code is excluded from language stats in the GitHub repo.
 lib/src/third_party/** linguist-vendored
+
+# jnigen generated code is hidden in GitHub diffs.
+lib/src/third_party/java/** linguist-generated


pkgs/java_http/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Hides jnigen generated code in GitHub diffs.
+lib/src/third_party/** linguist-generated
```
natebosch, line 2: Hmm, Is it possible to more narrowly target this?

I agree that marking these existing generated files as generated is a good idea - I understanding the reasoning for using this target. When I saw the path I was expecting to see `linguist-vendored`, not `linguist-generated`.

I think the sole next directory is probably specific enough to assume it's generated. How about

```suggestion
lib/src/third_party/** linguist-vendored
lib/src/third_party/java/** linguist-generated
```

The first line will catch if we happen to add other third party source for some other reason.

alex-james-dev: Ah I see what you mean, just because it's in the `third-party` directory it doesn't necessarily mean it's been generated by `jnigen`. In the future, we could add third-party code to `third_party` that hasn't been generated by `jnigen`.  Therefore, we shouldn't mark the entire `third_party` directory as `linguist-generated` but instead `linguist-vendored`.

The suggestions look good and have been committed :heavy_check_mark:

Thanks @natebosch!

alex-james-dev: > I think the sole next directory is probably specific enough to assume it's generated.

Would this catch the case where we are generating bindings for classes whose package name doesn't begin with `java`. For example, I think `okhttp3.Request` bindings would currently go in `third_party/okhttp3/Request.dart`.

To solve this we could put all the `jnigen` generated bindings in their own directory such as `third_party/generated`. Then `.gitattributes` would be something like:

```gitattributes
lib/src/third_party/** linguist-vendored
lib/src/third_party/generated/** linguist-generated
```

Although, at the moment we are only generating bindings for classes whose package name does begin with `java`. So we might not need to handle this case for now.

> Hmm, Is it possible to more narrowly target this?

Alternatively, we could also consider generating the `jnigen` bindings as a single file such as `native_java_bindings.dart`. Then `.gitattributes` would be something like:

```gitattributes
lib/src/third_party/** linguist-vendored
lib/src/third_party/native_java_bindings.dart linguist-generated
```

@natebosch What are your thoughts? :brain:

natebosch: > To solve this we could put all the `jnigen` generated bindings in their own directory such as `third_party/generated`.

This sounds like the optimal layout, but I'm also on board with _not_ making this change in favor of reduced churn.

I'd be fine with either (in slight preference order):

1. Keeping this PR as is - we can deal with handling other `linguist-generated` content if and when it starts showing up in PRs, which may never happen.
2. Refactor into a `lib/src/third_party/generated/**` so that we never have to worry about adjusting the config for new generated content, which may be unnecessary churn if we never generate for other classes.

alex-james-dev: > Keeping this PR as is - we can deal with handling other linguist-generated content if and when it starts showing up in PRs, which may never happen.

I think we should go ahead with keeping the PR as is, it feels like the best option to me!

natebosch (approved):

PR #1010, opened 2023-08-21
Don't check the formatting of `ffigen`ed code

brianquinlan:
Don't check the formatting of `native_cupertino_bindings.dart`.
---

<details>

</details>

### DIFF
--- a/.github/workflows/cupertino.yml
+++ b/.github/workflows/cupertino.yml
@@ -36,7 +36,14 @@ jobs:
         name: Install dependencies
         run: flutter pub get
       - name: Check formatting
-        run: dart format --output=none --set-exit-if-changed .
+        # Don't lint the generated file native_cupertino_bindings.dart
+        # This approach is simpler than using `find` and excluding that file
+        # because `dart format` also excludes other file e.g. ones in
+        # directories start with '.'.
+        run: >
+          mv lib/src/native_cupertino_bindings.dart lib/src/native_cupertino_bindings.tmp &&
+          dart format --output=none --set-exit-if-changed . &&
+          mv lib/src/native_cupertino_bindings.tmp lib/src/native_cupertino_bindings.dart
         if: always() && steps.install.outcome == 'success'
       - name: Analyze code
         run: flutter analyze --fatal-infos
--- a/pkgs/cupertino_http/lib/src/native_cupertino_bindings.dart
+++ b/pkgs/cupertino_http/lib/src/native_cupertino_bindings.dart
@@ -305,8 +305,9 @@ class NativeCupertinoHttp {
   }
 
   late final _callocPtr = _lookup<
-      ffi.NativeFunction<
-          ffi.Pointer<ffi.Void> Function(ffi.Size, ffi.Size)>>('calloc');
+          ffi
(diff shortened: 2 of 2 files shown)


natebosch:
We haven't needed to opt generated files out of formatting checks in other repos - is there some other change we can make to our processes that would let us validate the formatting of this file on CI? Is there a mismatch in the SDK version we are using on CI vs locally or something like that?

natebosch:
Will we roll this back once we have an SDK with the bug fix?

brianquinlan:
> We haven't needed to opt generated files out of formatting checks in other repos - is there some other change we can make to our processes that would let us validate the formatting of this file on CI? Is there a mismatch in the SDK version we are using on CI vs locally or something like that?

`ffigen` does not generate code consistently with `dart format`.

Normally I run `dart format` on the generated code myself but I don't see any value in doing that - the code is not meant to be read by humans.

The regression in `dart format` is what motivated me to change the linting logic but is there any value in checking the formatting in a generated file?

.github/workflows/cupertino.yml
```diff
@@ -36,7 +36,14 @@ jobs:
...
         run: flutter pub get
       - name: Check formatting
-        run: dart format --output=none --set-exit-if-changed .
+        # Don't lint the generated file native_cupertino_bindings.dart
+        # This approach is simpler than using `find` and excluding that file
+        # because `dart format` also excludes other file e.g. ones in
+        # directories start with '.'.
+        run: >
+          mv lib/src/native_cupertino_bindings.dart lib/src/native_cupertino_bindings.tmp &&
+          dart format --output=none --set-exit-if-changed . &&
```
natebosch, line 45: The `&&` means the file will not move back if any other files are not formatted correctly. That might not matter on CI though since we typically wouldn't run the other steps if the formatting check failed? I think leaving the mutated state on CI shouldn't hurt, but it makes me a little nervous.

natebosch (approved):

.github/workflows/cupertino.yml
brianquinlan: I'm not a bash expert but I think that this becomes significantly more complex because you have to save the exist code of `dart format` and then use it as the return value of the line.

brianquinlan:
> > is there any value in checking the formatting in a generated file?
> 
> Mostly avoiding the infrastructure to ignore those files. I also do find it helpful to have formatted code if I jump-to-definition and end up in that file in my IDE.

The code is formatted 99% correctly so reading it in an IDE isn't a problem.

.github/workflows/cupertino.yml
natebosch: Yeah, I'd probably drop it in a subshell and use `trap` if we wanted to avoid the inconsistent state on failure.

```yaml
        run: >
          (set -e;
          mv lib/src/native_cupertino_bindings.dart lib/src/native_cupertino_bindings.tmp;
          cleanup() {
            mv lib/src/native_cupertino_bindings.tmp lib/src/native_cupertino_bindings.dart;
          };
          trap cleanup EXIT
          dart format --output=none --set-exit-if-changed .;)
```

I'd be fine either way  though - I don't think we currently care about repo consistency after a formatting failure so it's fine to keep the code simpler and focused on the happy case.

PR #1011, opened 2023-08-23
Cleanup `package:http` utils

parlough:
1. Simplifies `mapToQuery` to avoid intermediate list building, and the unnecessary/repetitive null checking on the `encoding` parameter. This does mean a null `encoding` is no longer accepted, but the API isn't exposed as part of the package and all usages of it pass in non-null values.
2. Remove now unnecessary `[charset] may not be null` doc comment now that `charset` is non-nullable.
3. Uses a [`case` clause](https://dart.dev/language/branches#if-case) to promote `input` to `TypedData`.


pkgs/http/lib/src/utils.dart
```diff
@@ -12,14 +12,11 @@ import 'byte_stream.dart';
...
 ///     //=> "foo=bar&baz=bang"
-String mapToQuery(Map<String, String> map, {Encoding? encoding}) {
-  var pairs = <List<String>>[];
-  map.forEach((key, value) => pairs.add([
-        Uri.encodeQueryComponent(key, encoding: encoding ?? utf8),
-        Uri.encodeQueryComponent(value, encoding: encoding ?? utf8)
-      ]));
-  return pairs.map((pair) => '${pair[0]}=${pair[1]}').join('&');
-}
+String mapToQuery(Map<String, String> map, {Encoding encoding = utf8}) =>
```
brianquinlan, line 15: Since this is only used in one place and the encoding is provided, you con probably change this to:

```dart
... {required Encoding encoding})
```

brianquinlan (commented):
Nice!

brianquinlan (approved):

Issue #876, opened 2023-02-22
[cronet_http] Support running in a different isolate


knaeckeKami:
Hi!

I'd like to give some feedback on cronet_http:

Currently, it is not possible to initiate http requests from an different isolate than the main isolate.

This due to the usage of `setMessageHandler()` in the call to `receiveBroadcastStream()`: https://github.com/dart-lang/http/blob/e2c4e6f7aea17e89a6c56d0ff3d3a200983fb9fe/pkgs/cronet_http/lib/src/cronet_client.dart#L239

Which is not supported on non-root isolates, see https://github.com/flutter/flutter/issues/119207 and

https://api.flutter.dev/flutter/services/BackgroundIsolateBinaryMessenger/setMessageHandler.html

```dart
void setMessageHandler(String channel, MessageHandler? handler) {
  throw UnsupportedError(
      'Background isolates do not support setMessageHandler(). Messages from the host platform always go to the root isolate.');
}
```

It's a common use case to have the network logic in a separate isolate in order to not block the UI while e.h. parsing JSON, but this makes it more difficult to do so.

cupertino_http uses ffi, so it does not have this issue.

Reproducible code:

run this on an android device:

```dart

import 'dart:isolate';
import 'package:flutter/services.dart';
import 'package:cronet_http/cronet_http.dart';

void main() {

  RootIsolateToken? token = RootIsolateToken.instance;
  Isolate.run(() async {
    print('Isolate running ${Isolate.current.debugName}');
    try {
      BackgroundIsolateBinaryMessenger.ensureInitialized(token!);
      final resp = await CronetClient.defaultCronetEngine().get(Uri.parse('https://www.google.com'));
      print(resp.body);
    } catch (e) {
      print(e);
    }
  });


}
```

brianquinlan:
Thanks for the report. I think that we need to wait for the issue with `setMessageHandler` to be fixed.

atrope:
Can we provide a workaround  to this issue so we can use http in isolates? Maybe a flag or something?

brianquinlan:
Could you explain what you mean about adding a flag?

atrope:
@brianquinlan as @knaeckeKami wrote, on iOS it works because it is using ffi maybe we can have a flag in android http to use another way of making the requests without the use of `setMessageHandler` because by the looks of it it will not be supported in the near future by flutter and i think using http in an isolate can make a huge difference in a lot of use cases.

knaeckeKami:
[...]

PR #1016, opened 2023-09-07
Switch `cronet_http` to use jnigen

brianquinlan:
- Fixes https://github.com/dart-lang/http/issues/876
- Uses `package:jnigen` to provide bindings to Cronet

---

<details>

</details>


pkgs/cronet_http/android/src/main/kotlin/io/flutter/plugins/cronet_http/UrlRequestCallbackProxy.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    override fun onFailed(
+        request: UrlRequest,
+        info: UrlResponseInfo?,
+        error: CronetException
+    ) {
+        callback.onFailed(request, info, error);
+    }
+}
```
HosseinYousefi, line 77: Missing newline at end.

pkgs/cronet_http/jnigen.yaml
```diff
@@ -0,0 +1,34 @@
...
+suspend_fun_to_async: true
+
+output:
+  bindings_type: dart_only
+  dart:
+    path: 'lib/src/jni/jni_bindings.dart'
+    structure: single_file
+
+class_path:
+  - '/Users/bquinlan/Downloads/cronet-api-113.5672.61/classes.jar'
```
HosseinYousefi, line 19: Is this still needed?

HosseinYousefi (approved):
LGTM! Thanks @brianquinlan!

Although maybe someone well-versed in `http` can take a closer look at the `http` specific parts.

pkgs/cronet_http/jnigen.yaml
brianquinlan: Removed.

pkgs/cronet_http/android/src/main/kotlin/io/flutter/plugins/cronet_http/UrlRequestCallbackProxy.kt
brianquinlan: Added

pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -186,26 +212,114 @@ class CronetClient extends BaseClient {
...
-        throw ClientException(
-            'Exception building CronetEngine: ${e.toString()}', request.url);
-      }
-    }
+    _engine ??= CronetEngine.build();
 
     final stream = request.finalize();
-
     final body = await stream.toBytes();
+    final responseCompleter = Completer<StreamedResponse>();
```
natebosch, line 219: [nit] This method is going from ~100 lines to ~140 lines.

Are there any blocks that can be pulled out into other methods that we compose here?

pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -186,26 +212,114 @@ class CronetClient extends BaseClient {
...
+            jb.$UrlRequestCallbackProxy_UrlRequestCallbackInterfaceImpl(
+      onResponseStarted: (urlRequest, responseInfo) {
+        responseStream = StreamController();
+        final responseHeaders =
+            _cronetToClientHeaders(responseInfo.getAllHeaders());
+        int? contentLength;
+
+        switch (responseHeaders['content-length']) {
+          case final contentLengthHeader?
+              when !_digitRegex.hasMatch(contentLengthHeader):
```
natebosch, line 241: [out of scope for this PR] Would it make sense to use `int.tryParse` and report an invalid header if we get back `null`? Can we skip the regex check in this code?

pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -186,26 +212,114 @@ class CronetClient extends BaseClient {
...
+          contentLength: contentLength,
+          reasonPhrase: responseInfo
+              .getHttpStatusText()
+              .toDartString(releaseOriginal: true),
+          request: request,
+          isRedirect: false,
+          headers: responseHeaders,
+        ));
+
+        byteBuffer = jb.ByteBuffer.allocateDirect(1024 * 1024);
```
natebosch, line 263: [out of scope for this PR] Consider adding a constant for this number - it looks like it is repeated a few times

pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -186,26 +212,114 @@ class CronetClient extends BaseClient {
...
+              const Stream.empty(), // Cronet provides no body for redirects.
+              responseInfo.getHttpStatusCode(),
+              contentLength: 0,
+              reasonPhrase: responseInfo
+                  .getHttpStatusText()
+                  .toDartString(releaseOriginal: true),
+              request: request,
+              isRedirect: true,
+              headers: _cronetToClientHeaders(responseInfo.getAllHeaders())));
+        }
```
natebosch, line 279: Should we have a return here?

natebosch (approved):

pkgs/cronet_http/lib/src/cronet_client.dart
brianquinlan: Yes! I added a test that would have caught this bug.

pkgs/cronet_http/lib/src/cronet_client.dart
brianquinlan: ACK (I think that this disappears in my next PR)

pkgs/cronet_http/lib/src/cronet_client.dart
brianquinlan: Negative numbers are also not allowed so it ends up being two checks. I'm not sure if that is better.

pkgs/cronet_http/lib/src/cronet_client.dart
brianquinlan: I'll try moving `cronetCallbacks` to a separate class in a follow-up PR and see how it looks

pkgs/cronet_http/lib/src/cronet_client.dart
natebosch: Given that I missed that behavior, I have a slight preference for making the negative check explicit with `if (contentLengthHeader.isNegative)` over implicitly checked by the regex, but I don't feel strongly about it.

brianquinlan: I used the regex because it is a direct transliteration from the [production in the RFC](https://datatracker.ietf.org/doc/html/rfc2616#section-14.13)

PR #1035, opened 2023-10-25
fix: missing HTTP method in cronet client

kientux:
In `send` function of `CronetClient`, HTTP method in request is never set to the builder. With default behavior, Cronet will use `GET` when there's no body and `POST` otherwise, so every `PUT` request will become `POST`.

This PR just set the missing HTTP method to the builder.

### DIFF
--- a/pkgs/cronet_http/lib/src/cronet_client.dart
+++ b/pkgs/cronet_http/lib/src/cronet_client.dart
@@ -332,6 +332,7 @@ class CronetClient extends BaseClient {
       builder.setUploadDataProvider(
           jb.UploadDataProviders.create2(body.toJByteBuffer()), _executor);
     }
+    builder.setHttpMethod(request.method.toJString());
     builder.build().start();
     return responseCompleter.future;
   }


pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -332,6 +332,7 @@ class CronetClient extends BaseClient {
       builder.setUploadDataProvider(
           jb.UploadDataProviders.create2(body.toJByteBuffer()), _executor);
     }
+    builder.setHttpMethod(request.method.toJString());
```
kevmoo, line 335: Weird this is not caught in an existing test.

@brianquinlan ?

brianquinlan: Because I accidentally deleted the test while also breaking the code :-(

brianquinlan:
I've incorporated your change into https://github.com/dart-lang/http/pull/1058 - thanks for the PR!

PR #1044, opened 2023-11-09
Require a package:jni version without known buffer overflows

brianquinlan:
- `package:jni` 0.7.1 has a potential buffer overflow when converting Java to Dart strings.

---

<details>

</details>

### DIFF
--- a/pkgs/cronet_http/CHANGELOG.md
+++ b/pkgs/cronet_http/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.2
+
+* Require `package:jni >= 0.7.2` to remove a potential buffer overflow.
+
 ## 0.4.1
  
 * Require `package:jni >= 0.7.1` so that depending on `package:cronet_http` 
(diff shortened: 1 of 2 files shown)


natebosch (approved):
We could publish this, I don't think it will hurt anything, but I also don't think it helps.

This doesn't prevent anyone from resolving the old package (they could get old `cronet_http` with old `jni`) and nothing stops projects from `dart pub upgrade` to get the working versions with the current `cronet_http`.

What effect is intended with this change?

brianquinlan:
> We could publish this, I don't think it will hurt anything, but I also don't think it helps.
> 
> This doesn't prevent anyone from resolving the old package (they could get old `cronet_http` with old `jni`) and nothing stops projects from `dart pub upgrade` to get the working versions with the current `cronet_http`.
> 
> What effect is intended with this change?

I was actually writing you in chat about that... My intent was to provide some signal that people should upgrade and this is the only way that I could think to do it.

I would expect that few people have any version constraints applied to `package:jni` so they will get the new version automatically.

For people who pin their dependencies, hopefully the CHANGELOG for `package:cronet_http` is attractive enough to make it seem worthwhile.

But I'm happy to not do a release for this if it won't have any positive effect (checking this code in still seems sensible).

natebosch:
> My intent was to provide some signal that people should upgrade and this is the only way that I could think to do it.

I don't think I would expect this publish to be noticed as sending that signal. Either way a `dart pub upgrade` is the way to get out of the situation and we need to inform some users about that either way.

PR #1046, opened 2023-11-14
Populate package:http_profile

derekxu16:


pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
+  Map<String, dynamic>? connectionInfo;
+
+  /// The content length of the request, in bytes.
+  int? contentLength;
+
+  /// The cookies presented to the server (in the 'cookie' header).
+  List<String>? cookies;
+
+  /// Events that occurred while issuing the request.
+  List<HttpProfileRequestEvent>? events;
```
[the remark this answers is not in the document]
derekxu16: I should have made these fields private based on the way I envisioned this class to be used. I envisioned that a user of this package would have to populate an events array they defined themself and then use that to instantiate an `HttpProfileRequestData` object. Is it ok if I make these fields private?

`events` is currently under `HttpProfileRequestData` in the service extension response (https://github.com/dart-lang/sdk/blob/385c0063af8fbe77aff8ace14afb2a1c633dbdef/runtime/vm/service/service_extension.md?plain=1#L313-L316). It seems important to reflect that events can happen at any time, so I think we need to move the property up a level in the service response type. What do you think @bkonyi?

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
+    this.cookies,
+    this.events,
+    this.followRedirects,
+    this.headers,
+    this.maxRedirects,
+    this.persistentConnection,
+    this.proxyDetails,
+  });
+
+  HttpProfileRequestData.buildErroneousRequestData({
```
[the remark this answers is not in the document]
derekxu16: As mentioned above, I should have made all of the fields of this class private. Then this constructor would be the one that lets a user write to the `error` field, and the other constructor would let a user write to every field except `error`.

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
brianquinlan: > I should have made these fields private based on the way I envisioned this class to be used. I envisioned that a user of this package would have to populate an events array they defined themself and then use that to instantiate an `HttpProfileRequestData` object. Is it ok if I make these fields private?

It may be more convenient for some clients to provide information at different times than `HttpClient`. For example, neither `cupertino_http` nor `cronet_http` keep track of redirects so it would be easiest for them to add the redirect info to the profiling information as they go (i.e. somewhere in [this function](https://github.com/dart-lang/http/blob/f0a02f98f7c921e86ecc81c70f38bb6fbccc81b9/pkgs/cupertino_http/lib/src/cupertino_client.dart#L177) rather than store that information until they can construct an `HttpProfileRequestData`.

It also might be better UX in DevTools if you can track redirects even if the request fails.

> `events` is currently under `HttpProfileRequestData` in the service extension response (https://github.com/dart-lang/sdk/blob/385c0063af8fbe77aff8ace14afb2a1c633dbdef/runtime/vm/service/service_extension.md?plain=1#L313-L316). It seems important to reflect that events can happen at any time, so I think we need to move the property up a level in the service response type. What do you think @bkonyi?

You can change the location is this package without changing the service extension.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
+  /// Events that occurred while issuing the request.
+  List<HttpProfileRequestEvent>? events;
+
+  /// The error associated with the failed request.
+  String? error;
+
+  /// Whether redirects were followed automatically.
+  bool? followRedirects;
+
+  Map<String, dynamic>? headers;
```
brianquinlan, line 62: At some point you'll have to document the semantics of fields like this - what can the the `dynamic` part be? Is it String|List<String>?

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
 
+import 'dart:async' show StreamController, StreamSink;
+import 'dart:developer'
+    show addHttpClientProfilingData, getHttpClientProfilingData, Timeline;
 import 'dart:io';
 
+/// Describes an event that occurred while issuing an HTTP request.
+class HttpProfileRequestEvent {
+  /// The time at which the event occurred, in microseconds.
+  final int timestamp;
```
bkonyi, line 13: Is it maybe worth making this a `DateTime`? That way we can handle the conversion without developers possibly needing to pass around an integer.

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
bkonyi: > It seems important to reflect that events can happen at any time, so I think we need to move the property up a level in the service response type. What do you think @bkonyi?

I don't see any issue moving `events` up into `HttpProfileRequest` and would be okay with updating the protocol to match. This would be something relatively easy to work around in DevTools (check for `events` under `HttpProfileRequest` first and then check under `HttpProfileRequestData` if it's not there).

pkgs/http_profile/lib/http_profile.dart
bkonyi: I _think_ header values can only be `String`, but I could be wrong.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -14,20 +192,95 @@ final class HttpClientRequestProfile {
...
+    _updated();
+  }
+
+  /// Details about the request.
+  set requestData(HttpProfileRequestData value) {
+    _data['requestData'] = value.toJson();
+    _updated();
+  }
+
+  StreamController<List<int>> _requestBody = StreamController<List<int>>();
```
bkonyi, line 238: Ubernit: this can be `final _requestBody = ...`, so can `_responseBody` below.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -14,20 +192,95 @@ final class HttpClientRequestProfile {
...
+    // tree shaken away.
     if (const bool.fromEnvironment('dart.vm.product') || !profilingEnabled) {
       return null;
     }
     final requestProfile = HttpClientRequestProfile._();
+    addHttpClientProfilingData(requestProfile._data);
+    // This entry is needed to support the id parameter of
+    // ext.dart.io.getHttpProfileRequest.
+    requestProfile._data['id'] =
+        'from_package/${getHttpClientProfilingData().length}';
```
bkonyi, line 283: Can items ever be removed from `getHttpClientProfilingData()`? We might want to leave the ID assignment to the SDK code if we don't want to expose a `getNextHttpClientProfilingDataId()` like method.

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
derekxu16: > rather than store that information until they can construct an `HttpProfileRequestData`

By making the `HttpClientRequestProfile.requestData` setter only an entire object, I was trying to enforce the service response not having a populated `HttpProfileRequest.request` field until the initial request is completed (https://github.com/dart-lang/sdk/blob/385c0063af8fbe77aff8ace14afb2a1c633dbdef/runtime/vm/service/service_extension.md?plain=1#L271). We can allow users to populate fields of `HttpClientRequestProfile.requestData` as they go instead, but then we would need some other way of determining when the service response should have a populated `HttpProfileRequest.request` field. Some ideas are populating the field whenever `HttpProfileResponseData.endTime` is non-null, or populating the field whenever at least one field of `HttpClientRequestProfile.requestData` is non-null.

> You can change the location is this package without changing the service extension.

Yeah I just thought it might be confusing if we started putting events in there that weren't related to the initial request, since the service extension documentation says that the data in `HttpProfileRequest.request` is related to the initial request (https://github.com/dart-lang/sdk/blob/385c0063af8fbe77aff8ace14afb2a1c633dbdef/runtime/vm/service/service_extension.md?plain=1#L269). But, it looks like we're already doing that (https://github.com/dart-lang/sdk/blob/39e60fa3fc0820628ce63abc4adcaa2f9e048cc2/sdk/lib/_http/http_impl.dart#L203). So yeah, we don't have to move the property location in the service extension response type, but since @bkonyi's ok with it, we might as well.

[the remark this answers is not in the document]
brianquinlan: I think that giving maximum flexibility to package developers in how their populate fields is desirable.

```
  // Will not be provided if the initial request has not yet completed.
  HttpProfileRequestData request [optional];
```

Is the comment supposed to be descriptive or prescriptive, i.e., the receiver may not receive `request` if the request has not completed **or** the sender must not set `request` until the request has been completed? I suspect that the first interpretation is correct but I'm not sure.

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
derekxu16: Resolving this because this thread is obsolete.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
 import 'dart:io';
 
+/// Describes an event that occurred while issuing an HTTP request.
+class HttpProfileRequestEvent {
+  /// The time at which the event occurred, in microseconds.
+  final int timestamp;
+  final String name;
+  final Map<String, dynamic>? arguments;
+
+  HttpProfileRequestEvent(this.timestamp, this.name, this.arguments);
```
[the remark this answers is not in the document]
derekxu16: DevTools doesn't seem to use them, so I've removed them here.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async' show StreamController, StreamSink;
+import 'dart:developer'
+    show addHttpClientProfilingData, getHttpClientProfilingData, Timeline;
 import 'dart:io';
 
+/// Describes an event that occurred while issuing an HTTP request.
```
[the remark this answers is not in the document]
derekxu16: I've added this example to the documentation of `HttpClientRequestProfile.addEvent`.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
...
+class HttpProfileRequestEvent {
+  /// The time at which the event occurred, in microseconds.
+  final int timestamp;
+  final String name;
+  final Map<String, dynamic>? arguments;
+
+  HttpProfileRequestEvent(this.timestamp, this.name, this.arguments);
+
+  Map<String, dynamic> toJson() => <String, dynamic>{
```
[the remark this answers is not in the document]
derekxu16: I've made these private.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,186 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async' show StreamController, StreamSink;
+import 'dart:developer'
+    show addHttpClientProfilingData, getHttpClientProfilingData, Timeline;
 import 'dart:io';
 
+/// Describes an event that occurred while issuing an HTTP request.
+class HttpProfileRequestEvent {
```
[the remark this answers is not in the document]
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
derekxu16: We have discussed this on a call and decided to allow users of this package to populate fields as they go, and to have the service extension RPCs always return all information that is available (i.e. we will not prevent information from being sent when a certain specific field has not yet been populated).

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Good idea. I've left the ID assignment to the SDK code.

pkgs/http_profile/lib/http_profile.dart
derekxu16: The timestamps returned by the service extensions currently use the timeline's monotonic clock. We could keep returning those kinds of timestamps from the service extensions and use `DateTime`s in this package if we add some conversion logic to the service extension code. Do you think we should do that?

bkonyi: Ah that's a good point, I forgot we used the timeline's clock. In that case, if we stick with the current implementation, we should be very careful to mention this in documentation and guide developers to use [`Timeline.now`](https://api.flutter.dev/flutter/dart-developer/Timeline/now.html) to get this timestamp as close to the request as possible.

We _could_ switch this to using `DateTime` in the package and figure out how to convert to the monotonic clock in the service extension, but at that point we might just be better serializing the `DateTime` itself and just using that instead. I probably used the monotonic clock in the first place because we were scraping all this data from the timeline, but since we're not doing that anymore and we're doing this conversion in DevTools, it might be a good time to simplify this.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+
+/// Describes proxy authentication details associated with an HTTP request.
+final class HttpProfileProxyData {
+  final String? _host;
+  final String? _username;
+  final bool? _isDirect;
+  final int? _port;
+
+  HttpProfileProxyData(
+      {String? host, String? username, bool? isDirect, int? port})
```
bkonyi, line 34: Ubernit: trailing comma

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+
+/// Describes details about an HTTP request.
+final class HttpProfileRequestData {
+  final Map<String, dynamic> _data;
+
+  final void Function() _updated;
+
+  /// The elements of [connectionInfo] can either be [String]s or [int]s.
+  set connectionInfo(Map<String, dynamic /*String|int*/ > value) {
+    _data['connectionInfo'] = value;
```
bkonyi, line 56: Should we check that `value is int || value is String` and throw an exception if it's not?

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+
+  /// The content length of the request, in bytes.
+  set contentLength(int value) {
+    _data['contentLength'] = value;
+    _updated();
+  }
+
+  /// The cookies presented to the server (in the 'cookie' header).
+  set cookies(List<String> value) {
+    _data['cookies'] = value;
```
bkonyi, line 68: Are we okay with developers being able to modify `value` after setting this property? The way this is currently written, developers can insert new cookies directly into the list and bypass the call to `_updated()`.

If we're not okay with that, we should create a copy of `value` that we insert into `_data`.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+  }
+
+  /// Whether redirects were followed automatically.
+  set followRedirects(bool value) {
+    _data['followRedirects'] = value;
+    _updated();
+  }
+
+  set headers(Map<String, List<String>> value) {
+    _data['headers'] = value;
```
bkonyi, line 85: Ditto

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+  }
+
+  /// Proxy authentication details for the request.
+  set proxyDetails(HttpProfileProxyData value) {
+    _data['proxyDetails'] = value._toJson();
+    _updated();
+  }
+
+  HttpProfileRequestData._(
+      Map<String, dynamic> this._data, void Function() this._updated);
```
bkonyi, line 109: Nit: trailing comma.

Also, can this be a const constructor?

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+/// Describes details about a response to an HTTP request.
+final class HttpProfileResponseData {
+  final Map<String, dynamic> _data;
+
+  final void Function() _updated;
+
+  /// Records a redirect that the connection went through. The elements of
+  /// [redirect] can either be [String]s or [int]s.
+  void addRedirect(Map<String, dynamic /*String|int*/ > redirect) {
+    _data['redirects'].add(redirect);
```
bkonyi, line 121: Ditto comment about being able to avoid the call to `_updated()` here and anywhere else we can insert a collection into `_data`.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+  }
+
+  /// The error associated with the failed request.
+  set error(String value) {
+    _data['error'] = value;
+    _updated();
+  }
+
+  HttpProfileResponseData._(
+      Map<String, dynamic> this._data, void Function() this._updated) {
```
bkonyi, line 202: Nit: trailing comma.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Switching everything to use `DateTime` sounds good to me.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+    _updated();
+  }
+
+  /// The error associated with the failed request.
+  set error(String value) {
+    _data['error'] = value;
+    _updated();
+  }
+
+  /// Whether redirects were followed automatically.
```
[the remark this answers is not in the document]
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+    _data['followRedirects'] = value;
+    _updated();
+  }
+
+  set headers(Map<String, List<String>> value) {
+    _data['headers'] = value;
+    _updated();
+  }
+
+  /// If [followRedirects] is true, this is the maximum number of redirects that
```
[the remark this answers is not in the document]
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+
+/// Describes details about a response to an HTTP request.
+final class HttpProfileResponseData {
+  final Map<String, dynamic> _data;
+
+  final void Function() _updated;
+
+  /// Records a redirect that the connection went through. The elements of
+  /// [redirect] can either be [String]s or [int]s.
+  void addRedirect(Map<String, dynamic /*String|int*/ > redirect) {
```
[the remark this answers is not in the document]
derekxu16: You could theoretically put arbitrary data in here like with `connectionInfo`, because DevTools just lists out the entries of the map, but I've switched the type of the parameter of `addRedirect` to be a class that only has `statusCode`, `method` and `location` fields. Let me know if you think there's any benefit to switching the parameter type back to `Map<String, dynamic>`.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+    _updated();
+  }
+
+  /// The cookies set by the server (from the 'set-cookie' header).
+  set cookies(List<String> value) {
+    _data['cookies'] = value;
+    _updated();
+  }
+
+  /// The elements of [connectionInfo] can either be [String]s or [int]s.
```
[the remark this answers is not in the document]
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+    _data['headers'] = value;
+    _updated();
+  }
+
+  // The compression state of the response.
+  //
+  // This specifies whether the response bytes were compressed when they were
+  // received across the wire and whether callers will receive compressed or
+  // uncompressed bytes when they listen to the response body byte stream.
+  set compressionState(String value) {
```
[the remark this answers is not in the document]
derekxu16: The value should be one of the values in `HttpClientResponseCompressionState`, let me know if you think that's ok or if I should make a copy of that enum in this file.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+
+  /// The time at which the response was completed, as a microsecond value on
+  /// the monotonic clock used by the [Timeline]. Note that DevTools will not
+  /// consider the request to be complete until [endTime] is non-null.
+  set endTime(int value) {
+    _data['endTime'] = value;
+    _updated();
+  }
+
+  /// The error associated with the failed request.
```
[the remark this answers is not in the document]
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Ah, true. Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
derekxu16: Done.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,208 @@
...
+  final void Function() _updated;
+
+  /// Records a redirect that the connection went through. The elements of
+  /// [redirect] can either be [String]s or [int]s.
+  void addRedirect(Map<String, dynamic /*String|int*/ > redirect) {
+    _data['redirects'].add(redirect);
+    _updated();
+  }
+
+  /// The cookies set by the server (from the 'set-cookie' header).
```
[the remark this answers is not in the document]
derekxu16: They're stringified cookies, so basically anything that can be returned from the `toString` method of `dart:io`'s `Cookie` class. My latest commit uses the `Cookie` type here, but I've realized that that isn't going to work. I will update this.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,254 @@
...
+final class HttpProfileRequestData {
+  final Map<String, dynamic> _data;
+
+  final void Function() _updated;
+
+  /// Information about the networking connection used in the HTTP request.
+  ///
+  /// This information is meant to be used for debugging.
+  ///
+  /// It can contain any arbitrary data as long as the values are of type
```
brianquinlan, line 80: Does that mean that the values in the map must be [String] or [int]?

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -2,8 +2,254 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async' show StreamController, StreamSink;
+import 'dart:developer' show addHttpClientProfilingData, Timeline;
 import 'dart:io';
```
brianquinlan, line 7: Where does dart:io get used? Can we remove it?

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
brianquinlan: Maybe put an example?

pkgs/http_profile/lib/http_profile.dart
derekxu16: It's being used to access the getter and setter of `HttpClient.enableTimelineLogging`. It will also be used for the `HttpClientResponseCompressionState` enum, if we decide not to make a copy of the enum (https://github.com/dart-lang/http/pull/1046#discussion_r1414586231).

pkgs/http_profile/lib/http_profile.dart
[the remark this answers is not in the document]
derekxu16: Done.

brianquinlan (approved):
I'm OK with submitting this. Before any sort of release you will need tests and some way to demonstrate that it actually works.

Do you have an SDK patch that works with this?

I'd be happy to try this out.

derekxu16:
I still need to update the types in the service extension spec to make them compatible with the types here. I've made those changes locally and checked the ways in which DevTools needs to be updated in response to them. I've noticed that it would be best if DevTools didn't need to consider when `requestStartTimestamp`, `requestMethod` or `requestUri` are null. Can I make these required parameters of `HttpClientRequestProfile.profile`?

derekxu16:
Cool, sounds good. And then regarding the comment above about needing tests, did you mean unit tests? I can add unit tests now. I also want to add integration tests ensuring that the service extensions correctly return profiling information recorded using this package, but I won't be able to add those until this package is published.

brianquinlan:
> Cool, sounds good. And then regarding the comment above about needing tests, did you mean unit tests?

Yep.

> I can add unit tests now. I also want to add integration tests ensuring that the service
> extensions correctly return profiling information recorded using this package, but
> I won't be able to add those until this package is published.

Why not?

derekxu16:
> Why not?

We would need to have a VM Service to communicate with in the integration tests. I guess we could copy [this helper](https://github.com/dart-lang/sdk/blob/5a76c005c65f3038c6acfe501ebc391e72cb0dcc/pkg/dds/test/common/test_helper.dart#L13) into this package, and then use that to write the tests. I was thinking of just adding them to`package:vm_service`'s tests though.

pkgs/http_profile/lib/http_profile.dart
```diff
@@ -14,20 +278,110 @@ final class HttpClientRequestProfile {
...
+    // Always return `null` in product mode so that the profiling code can be
+    // tree shaken away.
     if (const bool.fromEnvironment('dart.vm.product') || !profilingEnabled) {
       return null;
     }
     final requestProfile = HttpClientRequestProfile._();
+    // This entry is needed to support the id parameter of
+    // ext.dart.io.getHttpProfileRequest.
+    requestProfile._data['id'] =
+        'from_package/${addHttpClientProfilingData(requestProfile._data)}';
```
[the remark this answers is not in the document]
derekxu16: Done.

brianquinlan:
Let me know if I am blocking you from landing this.

derekxu16:
It can't be landed yet because I'm still in the process of rolling the new `package:vm_service` version into DevTools.

brianquinlan:
Do the tests/lint get run automatically? I would guess not because you don't have a github workflow setup to do that. If, when you sync this PR to main, there are no presubmit failures, you can just submit and I'll setup the automation.

PR #1048, opened 2023-11-21
Update lints to latest, etc

kevmoo:
Updated min SDK on http_client_conformance_tests
Bumped pkg:http version on some examples


natebosch (dismissed):
Packages with changes should get a version bump to `-wip`

kevmoo:
> Packages with changes should get a version bump to `-wip`

done!

natebosch (approved):

pkgs/cupertino_http/example/integration_test/client_test.dart
```diff
@@ -45,7 +46,7 @@ void testClient(Client client) {
       }
       final request = StreamedRequest('POST', uri);
       request.sink.add(data);
-      request.sink.close();
+      unawaited(request.sink.close());
       await client.send(request);
```
lrhn, line 50: You could also try
```dart
await (request.sink.close(), client.send(request)).wait;
```
(But the current text is equivalent to what was before, so that's the safer bet.)

PR #1049, opened 2023-11-21
[http] use pkg:web, require Dart 3.2

kevmoo:
Now supports wasm
Prepare to release

### DIFF
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -3,8 +3,10 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:html';
+import 'dart:js_interop';
 import 'dart:typed_data';
+
+import 'package:web/helpers.dart';
 
 import 'base_client.dart';
 import 'base_request.dart';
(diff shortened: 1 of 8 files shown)


pkgs/http/lib/src/browser_client.dart
```diff
@@ -55,19 +57,22 @@ class BrowserClient extends BaseClient {
           'HTTP request failed. Client is already closed.', request.url);
     }
     var bytes = await request.finalize().toBytes();
-    var xhr = HttpRequest();
+    var xhr = XMLHttpRequest();
     _xhrs.add(xhr);
     xhr
-      ..open(request.method, '${request.url}', async: true)
```
srujzs, line 61: Why not keep this? \*shrug\*

srujzs (approved):

pkgs/http/lib/src/browser_client.dart
kevmoo: it *is* there. `async` is positional in the source API.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -112,3 +117,30 @@ class BrowserClient extends BaseClient {
     _xhrs.clear();
   }
 }
+
+extension on XMLHttpRequest {
```
kevmoo, line 121: Might be nice to put this in `pkg:web/helpers`

PR #1051, opened 2023-11-21
Update platform-specific imports for wasm

kevmoo:
Fix some docs

### DIFF
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -26,8 +26,8 @@ BaseClient createClient() {
   return BrowserClient();
 }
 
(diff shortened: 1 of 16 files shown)


pkgs/http/lib/src/browser_client.dart
```diff
@@ -26,8 +26,8 @@ BaseClient createClient() {
   return BrowserClient();
 }
 
-/// A `dart:html`-based HTTP client that runs in the browser and is backed by
-/// XMLHttpRequests.
+/// A `package:web`-based HTTP client that runs in the browser and is backed by
+/// `XMLHttpRequest`.
```
srujzs, line 30: nit: `[XMLHttpRequest]` instead? That should work in dartdoc if it's in scope.

srujzs (approved):

PR #1064, opened 2023-12-01
[http_client_conformance_tests] Updates to support wasm compilation

kevmoo:
With wasm, the hybrid logic decodes JSON numbers as double
This fix adds in a helper to make sure we get `int` when desired


kevmoo:
FYI @jakemac53 @srujzs 

Partying off https://github.com/dart-lang/test/pull/2144

With this PR (and the below delta to pkg:http) all tests pass with `dart test -p chrome --compiler dart2wasm`

```yaml

dependency_overrides:
  test:
    git:
      url: https://github.com/dart-lang/test
      ref: release-wasm
      path: pkgs/test
  test_api:
    git:
      url: https://github.com/dart-lang/test
      ref: release-wasm
      path: pkgs/test_api
  test_core:
    git:
      url: https://github.com/dart-lang/test
      ref: release-wasm
      path: pkgs/test_core
```

pkgs/http_client_conformance_tests/lib/src/request_methods_tests.dart
```diff
@@ -8,7 +8,7 @@ import 'package:stream_channel/stream_channel.dart';
 import 'package:test/test.dart';
 
 import 'request_methods_server_vm.dart'
-    if (dart.library.html) 'request_methods_server_web.dart';
```
kevmoo, line 11: Oops! This also messed things up.

CC @srujzs – about the lint we discussed!

kevmoo: gotta use `dart.library.js_interop`

pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart
```diff
@@ -10,12 +10,17 @@ import 'dart:io';
 
 import 'package:dart_style/dart_style.dart';
 
-const vm = '''// Generated by generate_server_wrappers.dart. Do not edit.
+const _export = '''export 'server_queue_helpers.dart'
+    show StreamQueueOfNullableObjectExtension;''';
```
brianquinlan, line 14: Cute ;-)

brianquinlan:
> dart test -p chrome --compiler dart2wasm

Do you want to add that to the test matrix?

pkgs/http/pubspec.yaml
```diff
@@ -20,3 +20,20 @@ dev_dependencies:
   shelf: ^1.1.0
   stream_channel: ^2.1.0
   test: ^1.16.0
+
+dependency_overrides:
```
brianquinlan, line 24: This this temporary? Should we release with these overrides?

jakemac53: They don't affect downstream users, and are required right now for the tests to pass until https://github.com/dart-lang/test/pull/2144 lands (sometime next week hopefully).

You could also put these in a `pubspec_overrides.yaml` file if you prefer, which might be a bit cleaner?

brianquinlan: If the intention is to not affect downstream users, shouldn't we use `dev_dependency_overrides`?

kevmoo: NO! Oops

pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart
kevmoo: That's how I roll!

pkgs/http/pubspec.yaml
brianquinlan: > NO! Oops

Does that mean that you are changing something here?

jakemac53: There is only `dependency_overrides`, and they never have any effect on any package other than the one in which they are written. 

In this case, they are only overrides of dev dependencies, so it is safe to release that way. You almost never would want to release with any dependency overrides of regular deps though (although in the case of cycles it _can_ be necessary).

Or, wait until the new version of package:test is released before landing this.

kevmoo: dropped

pkgs/http_client_conformance_tests/lib/src/server_queue_helpers.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:async/async.dart';
+
+extension StreamQueueOfNullableObjectExtension on StreamQueue<Object?> {
+  Future<int> get nextAsInt async => ((await next) as num).toInt();
```
brianquinlan, line 8: Could you add something like:
// When run under dart2wasm, JSON numbers are always returned floats.

And please file an issue to run the tests under dart2wasm.

brianquinlan (approved):
Nice!

pkgs/http_client_conformance_tests/lib/src/server_queue_helpers.dart
kevmoo: Done!

PR #1069, opened 2023-12-04
Document that runWithClient must be called for every isolate

brianquinlan:
- Document that runWithClient must be called for every isolate

---

<details>

</details>


natebosch (approved):

natebosch:
Might be worth expanding a bit, as it it sounds a bit like the call effects global state and only needs to be run once, not that it needs to wrap the Zone where the client is used.. Something like "When an Isolate is spawned it does not inherit any variables from the calling Zone so `runWithClient` needs to be used in each Isolate to run code that will use an Http client."

PR #1090, opened 2023-12-22
feat: support multiple headers

medz:
- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/http.dart
+++ b/pkgs/http/lib/http.dart
@@ -20,6 +20,7 @@ export 'src/base_response.dart';
 export 'src/byte_stream.dart';
 export 'src/client.dart' hide zoneClient;
 export 'src/exception.dart';
+export 'src/headers.dart';
 export 'src/multipart_file.dart';
 export 'src/multipart_request.dart';
 export 'src/request.dart';
@@ -34,7 +35,7 @@ export 'src/streamed_response.dart';
 /// the same server, you should use a single [Client] for all of those requests.
 ///
 /// For more fine-grained control over the request, use [Request] instead.
-Future<Response> head(Uri url, {Map<String, String>? headers}) =>
+Future<Response> head(Uri url, {Map<String, Iterable<String>>? headers}) =>
     _withClient((client) => client.head(url, headers: headers));
 
 /// Sends an HTTP GET request with the given headers to the given URL.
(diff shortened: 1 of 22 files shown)


brianquinlan:
Hi @medz, is this ready for review?

medz:
@brianquinlan I'm ready, but there are a few problems I can't solve. I observed that the mock client uses isolate, and my headers implementation cannot send messages. I don't have a good solution for the time being.

pkgs/http/lib/http.dart
```diff
@@ -157,7 +166,8 @@ Future<String> read(Uri url, {Map<String, String>? headers}) =>
 ///
 /// For more fine-grained control over the request and response, use [Request]
 /// instead.
-Future<Uint8List> readBytes(Uri url, {Map<String, String>? headers}) =>
+Future<Uint8List> readBytes(Uri url,
```
medz, line 169: In addition, similar to here, I think it is completely possible to use Object? to set headers, because headers are made according to MDN standards, so in theory it can allow many structures to be converted into headers

brianquinlan:
You are changing the public interface to several widely-used interfaces, which would be massively breaking.

Multiple header values are supported by separating them my commas and assigning them to the name field e.g.

```dart
headers = {"fruit": "apple,banana"};
```

Maybe it would be more practical to add a function to split and combine comma-separated field values e.g.

```dart
List<String> splitHeaderValues(String values);
joinHeaderValues(Iterable<String> values);
```

...but now that I think about it, those functions would be trivial.

Maybe the best solution would be to document? Like at https://pub.dev/documentation/http/latest/http/BaseRequest/headers.html

medz:
> You are changing the public interface to several widely-used interfaces, which would be massively breaking.
> 
> Multiple header values are supported by separating them my commas and assigning them to the name field e.g.
> 
> ```dart
> headers = {"fruit": "apple,banana"};
> ```
> 
> Maybe it would be more practical to add a function to split and combine comma-separated field values e.g.
> 
> ```dart
> List<String> splitHeaderValues(String values);
> joinHeaderValues(Iterable<String> values);
> ```
> 
> ...but now that I think about it, those functions would be trivial.
> 
> Maybe the best solution would be to document? Like at https://pub.dev/documentation/http/latest/http/BaseRequest/headers.html

This PR will indeed cause breaking changes, but it will make subsequent releases permanent. Splitting using `,` is indeed possible, but many http header values contain `,` which can easily cause wrong splitting.

Maybe changing the type of e.g. http.post(headers: xxx) to Object? would be a good choice, and content built with previous versions can be migrated to the new headers implementation without changing the value type. I will make a commit later to implement it

medz:
The PR close, new #1106 replace this

PR #1095, opened 2023-12-29
Reduce duplicate response header parsing in browser client

parlough:


brianquinlan:
Do we have tests that cover this?

parlough:
Sorry I haven't had a chance to come back to this yet, but it does look like this flow is under tested. I'll come back to this soon and work on increasing coverage here.

PR #1107, opened 2024-01-04
Update URL parsing in README.md

talperets:
To enhance the reliability of URL construction, especially during external API testing, I recommend replacing manual URL construction with the `Uri.parse` method. Manually constructing URLs can often result in broken links, whereas `URI.parse` provides a more robust and error-resistant approach.

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/pkgs/http/README.md
+++ b/pkgs/http/README.md
@@ -15,7 +15,7 @@ you to make individual HTTP requests with minimal hassle:
 ```dart
 import 'package:http/http.dart' as http;
 
-var url = Uri.https('example.com', 'whatsit/create');
+var url = Uri.parse('example.com/whatsit/create');
 var response = await http.post(url, body: {'name': 'doodle', 'color': 'blue'});
 print('Response status: ${response.statusCode}');
 print('Response body: ${response.body}');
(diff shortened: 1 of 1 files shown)


brianquinlan:
I think that @natebosch feels that we should use explicit scheme factories rather than `parse` when possible.

natebosch:
> Manually constructing URLs can often result in broken links, whereas `URI.parse` provides a more robust and error-resistant approach.

I strongly disagree with this sentence. Writing a URL as a String (especially with `$variable` interpolation) is what I call a "manual" Uri construction - the author must handle all details of constructing that string correctly, including escaping characters etc. That is `Uri.parse('https://foo.com/$someUserInput')` is brittle and can result in an invalid or misinterpreted Url where `Uri.https('foo.com', someUserInput)` handles the escaping and will not result in an invalid or misinterpreted URL. Handling query parameters correctly is even more fraught.

Manual deconstructing a Url is similarly a bad idea. I'd also reject a change which rewrites a `Uri.parse(someVariable)` into something like `Uri.https(findHost(someVariable), findPath(someVariable))` - parsing a String that isn't in the source code is the intended use for `Uri.parse`.

PR #1109, opened 2024-01-09
Add `BaseResponseWithUrl.url`

brianquinlan:
Closes https://github.com/dart-lang/http/pull/699
Closes https://github.com/dart-lang/http/pull/623
Closes https://github.com/dart-lang/http/pull/692
Fixes https://github.com/dart-lang/http/issues/293

Adds a new class `BaseResponseWithUrl`, which can hold  a new `BaseResponse.url` fields until `package:http` v2 is released.

_This is an experiment in package evolution - lets see how it works out!_

Users can test whether the new field is present in a `BaseResponse`. For example:

```dart
final client = Client();
final response = client.get(Uri.https('example.com', '/'));
Uri? finalUri;
if (response case BaseResponseWithUrl(:final url)) {
  finalUri = url;
}

// Do something with `finalUri`.
client.close();
```

Adds `BaseResponseWithUrl.url`, which is the final URL of the response after following any redirects.

Conformance tests pass in a separate PR:
https://github.com/dart-lang/http/pull/1110

As soon as this PR lands, I'll create a follow-up with the "next-breaking-release" label that moves `uri` to `BaseRequest` and deprecates `BaseResponseWithUrl`.

---

<details>

</details>


brianquinlan:
@AlexV525 Any thoughts about this approach?

pkgs/http/lib/http.dart
```diff
@@ -25,7 +25,7 @@ export 'src/multipart_request.dart';
 export 'src/request.dart';
 export 'src/response.dart';
 export 'src/streamed_request.dart';
-export 'src/streamed_response.dart';
+export 'src/streamed_response.dart' hide StreamedResponseV2;
```
natebosch, line 28: [optional] Consider using a `show` instead. I don't think we've ever formalized it, but I think we've been moving in that direction for public libraries.

I'll probably at some point add explicit `show` clauses for all the exports here.

pkgs/http/lib/src/base_request.dart
```diff
@@ -132,13 +132,25 @@ abstract class BaseRequest {
...
-          persistentConnection: response.persistentConnection,
-          reasonPhrase: response.reasonPhrase);
+
+      if (response is BaseResponseV2) {
+        return StreamedResponseV2(ByteStream(stream), response.statusCode,
+            contentLength: response.contentLength,
+            request: response.request,
+            headers: response.headers,
+            isRedirect: response.isRedirect,
+            url: (response as BaseResponseV2).url,
```
natebosch, line 142: I think we can avoid the cast with if-case a pattern to pull out the `url` field

```suggestion
      if (response case BaseResponseV2(:final url)) {
        return StreamedResponseV2(ByteStream(stream), response.statusCode,
            contentLength: response.contentLength,
            request: response.request,
            headers: response.headers,
            isRedirect: response.isRedirect,
            url: url,
```

pkgs/http/lib/src/base_response.dart
```diff
@@ -68,3 +71,34 @@ abstract class BaseResponse {
...
+mixin BaseResponseV2 on BaseResponse {
+  /// The [Uri] of the response returned by the server.
+  ///
+  /// If no redirects were followed, [url] will be the same as the requested
+  /// [Uri].
+  ///
+  /// If redirects were followed, [url] will be the [Uri] of the last redirect
+  /// that was followed.
+  ///
+  /// May be `null` when the response is for a special URL such as "about:".
```
natebosch, line 102: Double checking - is this a limitation of the Web platform?

Would we be able to make this non-nullable if we override a `null` one from the browser back to the request Uri?

pkgs/http/test/io/request_test.dart
```diff
@@ -46,15 +46,17 @@ void main() {
     final response = await request.send();
 
     expect(response.statusCode, equals(302));
+    expect(
+        (response as http.BaseResponseV2).url, serverUrl.resolve('/redirect'));
```
natebosch, line 50: [nit] Prefer to encode the type expectation with a matcher instead of casting and relying on the cast failure to fail the test.

```suggestion
      expect(
          response,
          isA<http.BaseResponseV2>()
              .having((r) => r.url, 'url', serverUrl.resolve('/redirect')));
```

pkgs/http/test/io/request_test.dart
```diff
@@ -46,15 +46,17 @@ void main() {
...
   });
 
   test('with redirects', () async {
     final request = http.Request('GET', serverUrl.resolve('/redirect'));
     final response = await request.send();
-
     expect(response.statusCode, equals(200));
     final bytesString = await response.stream.bytesToString();
     expect(bytesString, parse(containsPair('path', '/')));
+    expect((response as http.BaseResponseV2).url, serverUrl.resolve('/'));
```
natebosch, line 59:
```suggestion
    expect(
        response,
        isA<http.BaseResponseV2>()
            .having((r) => r.url, 'url', serverUrl.resolve('/redirect')));
```

natebosch (approved):

pkgs/http/lib/src/base_response.dart
AlexV525: Agreed. What is the purpose of handling URLs such as "about:"? IMO they cannot exist in responses.

AlexV525:
> @AlexV525 Any thoughts about this approach?

The implementation looks simple and good to me. Left some comments above.

pkgs/http/lib/http.dart
brianquinlan: Done

pkgs/http/lib/src/base_request.dart
brianquinlan: Done

pkgs/http/test/io/request_test.dart
brianquinlan: Done

pkgs/http/lib/src/base_response.dart
brianquinlan: The standards docs for `XMLHttpRequest` aren't clear on when this can happen, but [`XMLHttpRequest.responseUrl`](https://xhr.spec.whatwg.org/#the-responseurl-attribute) can be the empty string. 

I'll make it non-nullable we can hope for the best.

brianquinlan:
FWIW, I renamed `BaseResponseV2` to `BaseResponseWithUrl` to allow for the possibility that we will add more than one new field/method in a minor release and don't want to break users because they add `with BaseResponseV2` to **their** classes!

brianquinlan:
@lrhn Do you want to provide feedback here? I'd like to land this relatively soon.

pkgs/http/lib/src/base_response.dart
```diff
@@ -68,3 +71,34 @@ abstract class BaseResponse {
...
+/// if (response case BaseResponseWithUrl(:final url)) {
+///   finalUri = url;
+/// }
+/// // Do something with `finalUri`.
+/// client.close();
+/// ```
+///
+/// [url] will be added to [BaseResponse] when `package:http` version 2 is
+/// released and this mixin will be deprecated.
+mixin BaseResponseWithUrl on BaseResponse {
```
lrhn, line 95: I'd make this `abstract interface BaseResponseWithUrl implements BaseResponse`.


Being a mixin with just abstract members means that it is just an interface, you just have to apply it with `with` instead of `implements`.
That's just more complexity for no actual effect. The `abstract interface` is the minimal and idiomatic way to add an abstract member to an interface.

pkgs/http/lib/src/io_client.dart
```diff
@@ -46,6 +47,22 @@ class _ClientSocketException extends ClientException
...
+
+  _IOStreamedResponseV2(super.stream, super.statusCode,
+      {super.contentLength,
+      super.request,
+      super.headers,
+      super.isRedirect,
+      required this.url,
+      super.persistentConnection,
+      super.reasonPhrase,
+      super.inner});
```
lrhn, line 63: Consider ordering these parameters in some consistent way.
If nothing else, because they are ordered in a way I can't see, consider putting the required parameter first.
(Maybe it's just because GitHub shows `required` as red that it stands out.)

pkgs/http/lib/src/streamed_response.dart
```diff
@@ -26,3 +26,19 @@ class StreamedResponse extends BaseResponse {
       super.reasonPhrase})
       : stream = toByteStream(stream);
 }
+
+/// This class is private to `package:http` and will be removed when
+/// `package:http` v2 is released.
+class StreamedResponseV2 extends StreamedResponse with BaseResponseWithUrl {
```
lrhn, line 32: Don't love the `V2` name, but if it's private and goes away in version 2 of the package, it is probably as direct as it can be. And being ugly is a feature, it makes one want to get rid of it. So LGTM.

lrhn (approved):
LGTM, but consider using an interface instead of a mixin.

pkgs/http/lib/src/streamed_response.dart
brianquinlan: LOL

pkgs/http/lib/src/io_client.dart
brianquinlan: I put the required parameter first.

pkgs/http/lib/src/base_response.dart
brianquinlan: That makes sense, thanks! Done.

JHubi1:
Am I missing something, or is this not working?

When running the following code:
```dart
request = await http
    .get(
  Uri.parse(tmpHost),
  headers: (jsonDecode(prefs!.getString("hostHeaders") ?? "{}") as Map)
      .cast<String, String>(),
)
    .timeout(const Duration(seconds: 5), onTimeout: () {
  return http.Response("Error", 408);
});
Uri finalUri = Uri.parse(tmpHost);
if (request case http.BaseResponseWithUrl(:final url)) {
  finalUri = url;
}
print(finalUri);
```
It always prints the value of tmpHost, and not the actual final URL. I tried it as well by using a client, same result. Neither protocol change nor full short URL services update that URL

PR #1114, opened 2024-01-12
Add the ability to get response headers as a `Map<String, List<String>>`

brianquinlan:
Adds a extension to `BaseResponse` like:

```dart
extension HeadersWithFieldLists on BaseResponse {
  /// The HTTP headers returned by the server.
  ///
  /// The header names are converted to lowercase and stored with their
  /// associated header values.
  ///
  /// Cookies can be parsed using the dart:io `Cookie` class:
  ///
  /// ```dart
  /// import "dart:io";
  /// import "package:http/http.dart";
  ///
  /// void main() async {
  /// final response = await Client().get(Uri.https('example.com', '/'));
  /// final cookies = [
  ///   for (var value i
  ///       in response.headersFieldValueList['set-cookie'] ?? <String>[])
  ///     Cookie.fromSetCookieValue(value)
  /// ];
  Map<String, List<String>> get headersFieldValueList ...
}
```

Fixes https://github.com/dart-lang/http/issues/24

---

<details>

</details>


pkgs/http/lib/src/base_response.dart
```diff
@@ -68,3 +70,68 @@ abstract class BaseResponse {
     }
   }
 }
+
+// "token" as defined in RFC 2616, 2.2
```
natebosch, line 74: [nit] Use `///` and make the comments on private members docs.

pkgs/http/lib/src/base_response.dart
```diff
@@ -68,3 +70,68 @@ abstract class BaseResponse {
...
+  /// import "package:http/http.dart";
+  ///
+  /// void main() async {
+  /// final response = await Client().get(Uri.https('example.com', '/'));
+  /// final cookies = [
+  ///   for (var value i
+  ///       in response.headersFieldValueList['set-cookie'] ?? <String>[])
+  ///     Cookie.fromSetCookieValue(value)
+  /// ];
+  Map<String, List<String>> get headersFieldValueList {
```
natebosch, line 122: This name is pretty long - can we consider any shorter ones? Maybe `headerSplitValues`?

natebosch (approved):

pkgs/http/lib/src/base_response.dart
brianquinlan: Done

pkgs/http/lib/src/base_response.dart
brianquinlan: Done

PR #1128, opened 2024-02-07
Create a simple WebSocket interface

brianquinlan:
- This just defines the interface that will be implemented using `dart:io`, `package:web` and `package:cupertino_http`.

---

<details>

</details>


pkgs/web_socket/lib/src/web_socket.dart
```diff
@@ -0,0 +1,135 @@
...
+  ///
+  /// - A close frame was received from the peer. [CloseReceived.code] and
+  ///   [CloseReceived.reason] will be set by the peer.
+  /// - A failure occured (e.g. the peer disconnected). [CloseReceived.code] and
+  ///   [CloseReceived.reason] will be a failure code defined by
+  ///   (RFC-6455)[https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1]
+  ///   (e.g. 1006).
+  ///
+  /// Errors will never appear in this [Stream].
+  Stream<WebSocketEvent> get events;
```
brianquinlan, line 134: This Stream and its relationship with the `sendText`/`sendBytes`/`close` methods is my biggest concern with this API.

As documented now, `sendText`, etc. throw if the WebSocket is closed. But the `ClosedReceived` event is delivered asynchronously so you can't rely on it being delivered in time to prevent those calls. For example:

```dart
timer = Timer.periodic(const Duration(seconds: 2), (timer) {
  websocket.sendText('<stock>GOOG</stock>');
});

await for (final event in websocket.events) {
  switch (event) {
  // Should probably do something with the stock quote.
  case CloseReceived:
    // This is not sufficient - the implementation may add `CloseReceived` to the `StreamController` and then
    // close the `StreamController` before this event is received. This probably means that, in practice, every
    // call to `sendText`/`sendBytes` will need to be guarded in a `try`/`catch`
    timer.cancel();
  }
}
```

Some alternatives:
1. just throw away `sendText`/`sendBytes` after the WebSocket is closed - this is what JavaScript does
2. use `SyncStreamController` (will that work?)
3. make a synchronous callback for events instead of using a `Stream`.

pkgs/web_socket/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
```
natebosch, line 1: [nit] Versions `0.0.x` don't always work well - I can't recall exactly why...

```suggestion
## 0.1.0-wip
```

pkgs/web_socket/lib/src/web_socket.dart
natebosch: I think even if we synchronously surface a `ClosedReceived` there is still the possibility of a `sendText` failing to write because of a closed socket, though it may be significantly less less likely.

I would probably start with throwing, and we can consider an option to silently discard if that has a clear use case.

pkgs/web_socket/pubspec.yaml
```diff
@@ -0,0 +1,11 @@
+name: web_socket
+description: "TODO: enter a descirption here"
+version: 1.0.0
```
natebosch, line 3: I would start without a version, and make it explicit when it gets into a publishable state.

```suggestion
publish_to: none
```

natebosch (approved):

pkgs/web_socket/lib/src/web_socket.dart
brianquinlan: > I think even if we synchronously surface a ClosedReceived there is still the possibility of a sendText failing to write because of a closed socket, though it may be significantly less less likely.

My plan was for `sendText`/`sendBytes` to not throw in that case. Any other behavior would be difficult on the web, [which only allows `send` to throw if the connection isn't established](https://websockets.spec.whatwg.org/#dom-websocket-send).

I have a [conformance test to verify that behavior](https://github.com/dart-lang/http/blob/7b30f614cfda972960f0a48017587a64d61c1781/pkgs/web_socket_conformance_tests/lib/src/disconnect_after_upgrade_tests.dart) (the tests are still very rough at this point).

I've updated the docs to make a note of that.

> I would probably start with throwing, and we can consider an option to silently discard if that has a clear use case.

SGTM!

PR #1151, opened 2024-03-05
[package:http_profile ] Add getters to classes

derekxu16:


pkgs/http_profile/lib/src/http_profile_request_data.dart
```diff
@@ -46,20 +62,43 @@ final class HttpProfileRequestData {
...
+  /// profile?.requestData.connectionInfo = {
+  ///   'localPort': 1285,
+  ///   'remotePort': 443,
+  ///   'connectionPoolId': '21x23',
+  /// };
+  ///
+  /// // Invalid
+  /// profile?.requestData.connectionInfo?['localPort'] = 1285;
+  /// ```
+  set connectionInfo(Map<String, dynamic /*String|int*/ >? value) {
```
derekxu16, line 79: I had to give the new getter a nullable return type, because `_requestData` won't have a `'connectionInfo'` entry until the setter is called. The setter's parameter type needs to be a supertype of the getter's return type, so my options were to make it nullable as well, or give the getter and setter different names. I chose to make the setter's parameter type nullable. Does that seem ok?

pkgs/http_profile/lib/src/http_profile_response_data.dart
```diff
@@ -38,6 +51,12 @@ final class HttpProfileResponseData {
...
   }
 
+  /// An unmodifiable list containing the redirects that the connection went
+  /// through.
+  List<HttpProfileRedirectData> get redirects =>
+      UnmodifiableListView((_data['redirects'] as List<Map<String, dynamic>>)
+          .map(HttpProfileRedirectData._fromJson));
+
   /// The body of the response.
   StreamSink<List<int>> get bodySink => _body.sink;
```
derekxu16, line 61: Should we add a getter that returns `_body.stream`?

brianquinlan: Hmmm...the issue with that is that listening to a `Stream` is destructive.

Maybe it would make sense to return a [split of the stream](https://pub.dev/documentation/async/latest/async/StreamSplitter-class.html) and document that e.g.

```dart
/// A **copy** of the bodyStream.
///
/// This getter is meant for use in tests and can result in significant additional memory usage.
Stream<List<int>> bodyStream;
```

brianquinlan (approved):

pkgs/http_profile/lib/src/http_profile_response_data.dart
derekxu16: Implementing the `requestData.bodyStream` getter as

```
Stream<List<int>> get bodyStream => StreamSplitter(_body.stream).split();
```

doesn't work, because the `.split()` call makes the `StreamSplitter` listen to `_body.stream`, which means that after the getter is called, it's impossible to listen to the `_data` backing map's `'_requestBodyStream'` entry.

https://github.com/dart-lang/http/blob/c7d90069eaef1c38683672e41897032f8cbcca9f/pkgs/http_profile/lib/src/http_client_request_profile.dart#L104

So to implement the getter, we have to unconditionally split the stream and return one copy from the getter and store the other copy in `_data['_requestBodyStream']`, but I don't think we want to do this, because it will increase memory consumption.

brianquinlan:
Merge away!

PR #1154, opened 2024-03-08
[package:http_profile] Store request and response bodies in the backing map as lists instead of as streams

derekxu16:
This PR also adds getters that return the request and response bodies as lists of ints

### DIFF
--- a/pkgs/http_profile/lib/src/http_profile_response_data.dart
+++ b/pkgs/http_profile/lib/src/http_profile_response_data.dart
@@ -32,14 +32,22 @@ final class HttpProfileResponseData {
   final void Function() _updated;
   final StreamController<List<int>> _body = StreamController<List<int>>();
 
+  Map<String, dynamic> get _responseData =>
+      _data['responseData'] as Map<String, dynamic>;
+
   /// Records a redirect that the connection went through.
   void addRedirect(HttpProfileRedirectData redirect) {
     _checkAndUpdate();
-    (_data['redirects'] as List<Map<String, dynamic>>).add(redirect._toJson());
+    (_responseData['redirects'] as List<Map<String, dynamic>>)
+        .add(redirect._toJson());
   }
 
-  /// The body of the response.
+  /// A sink that can be used to record the body of the response.
   StreamSink<List<int>> get bodySink => _body.sink;
+
+  /// The body of the response represented as an unmodifiable list of bytes.
+  List<int> get bodyBytes =>
+      UnmodifiableListView(_data['responseBodyBytes'] as List<int>);
 
   /// Information about the networking connection used in the HTTP response.
   ///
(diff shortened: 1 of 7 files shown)


pkgs/http_profile/test/http_profile_response_data_test.dart
```diff
@@ -199,28 +199,20 @@ void main() {
...
 
-  test('using HttpClientRequestProfile.requestBodySink', () async {
-    final requestBodyStream =
-        backingMap['_requestBodyStream'] as Stream<List<int>>;
-
-    profile.requestData.bodySink.add([1, 2, 3]);
-    profile.requestData.close();
-
-    expect(await requestBodyStream.expand((i) => i).toList(), [1, 2, 3]);
-  });
```
derekxu16, line 215: I moved this test case to `test/http_profile_request_data_test.dart`.

pkgs/http_profile/lib/src/http_profile_response_data.dart
```diff
@@ -176,11 +184,12 @@ final class HttpProfileResponseData {
   ///
   /// [endTime] is the time when the response was fully received. It defaults
   /// to the current time.
-  void close([DateTime? endTime]) {
+  Future<void> close([DateTime? endTime]) async {
```
brianquinlan, line 187: I wonder if we should but a note here - and in the http_profile_request_data equivalent - that it is safe to unawait this Future? Probably we don't need to say that.

pkgs/http_profile/lib/src/http_profile_response_data.dart
```diff
@@ -176,11 +184,12 @@ final class HttpProfileResponseData {
...
     _checkAndUpdate();
     _isClosed = true;
-    unawaited(bodySink.close());
-    _data['endTime'] = (endTime ?? DateTime.now()).microsecondsSinceEpoch;
+    await bodySink.close();
+    _responseData['endTime'] =
```
brianquinlan, line 191: So this was incorrect before?

brianquinlan (approved):

pkgs/http_profile/lib/src/http_profile_response_data.dart
derekxu16: It worked before because we were awaiting `_body.stream.expand(...).toList(...)` on the SDK side, so it was impossible for this unawaited `bodySink.close()` to cause the `'requestBody'` or `'responseBody'` entries to get populated incorrectly here (https://github.com/dart-lang/sdk/blob/9efa95674966c1d8b498c1846372f36db689f8c7/sdk/lib/io/network_profiling.dart#L34-L44).

With the changes in this PR, we need to await `bodySink.close()` to guarantee that all the bytes that have been added to the sink have also been added to `_data['requestBodyBytes']` before the SDK tries to read `_data['requestBodyBytes']`.

https://github.com/dart-lang/http/blob/bb1f7a83e4770f6ebc8673a54507833f40eeb433/pkgs/http_profile/lib/src/http_client_request_profile.dart#L81-L83

PR #1206, opened 2024-05-22
[package:http_profile] Modify experimental status notice in README.md

derekxu16:


derekxu16:
Now that we're planning to publish the package under tools.dart.dev, do we want to change the notice to something like this? Or, do we want to remove the notice?

pkgs/http_profile/README.md
```diff
@@ -56,14 +56,8 @@ into an HTTP client.
...
 
-**NOTE**: This package is currently experimental and published under the
-[labs.dart.dev](https://dart.dev/dart-team-packages) pub publisher in order to
-solicit feedback.
-
-For packages in the labs.dart.dev publisher we generally plan to either graduate
-the package into a supported publisher (dart.dev, tools.dart.dev) after a period
-of feedback and iteration, or discontinue the package. These packages have a
-much higher expected rate of API and breaking changes.
+**NOTE**: This package is currently experimental, so new versions that include
```
brianquinlan, line 59: I don't know if "status: experimental" goes with the tools publisher - I don't see any other packages published under that publisher with such a notice? Could you ask in the dart-ecosystem team chat?

derekxu16: I've just updated the notice.

brianquinlan (approved):

PR #1214, opened 2024-05-28
Mark flutter .metadata files as generated

natebosch:
Removes one file with a comment indicating it should not be hand edited
from the default view for github PR diffs.


natebosch:
I took the list I found in https://github.com/google-gemini/generative-ai-dart/pull/154 and removed the entires which didn't correspond to any file in this PR which left us with one file concretely marked as generated.

I'd also be OK with preemptively adding the other patterns in case those files get checked in later.

There are others which I remain skeptical will ever get hand edits and need review - but so far I'm only adding the ones which are documented that way.

brianquinlan (approved):

Issue #1219, opened 2024-05-31
[http_client_conformance_tests] Add flag to skip tests for folded headers


Anikate-De:
According to the current internet standard ([RFC 9110: HTTP Semantics](https://datatracker.ietf.org/doc/html/rfc9110)), [`"\r\n" is invalid`](https://datatracker.ietf.org/doc/html/rfc9110#section-5.5)

> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message.

`OkHttp` [(please see this)](https://square.github.io/okhttp/changelogs/changelog_2x/#version-250) and many modern web frameworks (like Flask) reject Header values which contain such illegal characters.

However, the `server headers folded headers` test group adds rigidity by forcing clients to replace invalid characters with space.

I (thanks to @brianquinlan) suggest adding a boolean flag `supportsFoldedHeaders` to [`testResponseHeaders()`](https://github.com/dart-lang/http/blob/master/pkgs/http_client_conformance_tests/lib/src/response_headers_tests.dart#L14) which ultimately lets us skip the test: [folded headers](https://github.com/dart-lang/http/blob/master/pkgs/http_client_conformance_tests/lib/src/response_headers_tests.dart#L157)

Clients rejecting headers with illegal values can also be considered to be conformant to the HTTP Standard.

Anikate-De:
I'm willing to take up this issue as it will be beneficial for making the upcoming package `ok_http` more conformant.

PR #1223, opened 2024-06-03
pkgs/ok_http: Close and remove all idle connections from the resource pool on response

Anikate-De:
Items in this PR:

- [x] Generated JNI Bindings for `okhttp3.ConnectionPool` (allows us to use the [`evictAll()`](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-connection-pool/evict-all.html?query=fun%20evictAll()) method)
- [x] Close all idle connections in the resource pool `onResponse()`
    - **WHY?:** During `testRequestCookies()`, the server determines the requests only by [binding on the socket](https://github.com/dart-lang/http/blob/9f022d26364fb09748cbca25834b23fdc26baf05/pkgs/http_client_conformance_tests/lib/src/request_cookies_server.dart#L20). And, OkHttp never frees or closes connections for all `enqueue()` calls. This led to the server assuming multiple requests as a single entity (as OkHttp never stopped communicating with the socket,) and parsing both `cookie` headers as one. This gave a cookies list with a length of 2, which failed the 2nd test case `multiple cookies semicolon separated`. By explicitly closing idle connections, we protect consecutive requests from being overlapped.
- [x] Add `testRequestCookies()`
- [x] Add `testResponseHeaders()` since all TCs are passing now that #1219 is closed (see #1222)

---

<details>

</details>

### DIFF
--- a/pkgs/ok_http/lib/src/ok_http_client.dart
+++ b/pkgs/ok_http/lib/src/ok_http_client.dart
@@ -126,6 +126,9 @@ class OkHttpClient extends BaseClient {
               request: request,
               contentLength: contentLength,
             ));
+
+            // Close and remove all idle connections from the resource pool.
+            _client.connectionPool().evictAll();
           },
           onFailure: (bindings.Call call, JObject ioException) {
             responseCompleter.completeError(
--- a/pkgs/ok_http/example/integration_test/client_test.dart
+++ b/pkgs/ok_http/example/integration_test/client_test.dart
@@ -19,10 +19,12 @@ Future<void> testConformance() async {
     testResponseBody(OkHttpClient(), canStreamResponseBody: false);
     testRequestHeaders(OkHttpClient());
     testRequestMethods(OkHttpClient(), preservesMethodCase: true);
+    testResponseHeaders(OkHttpClient(), supportsFoldedHeaders: false);
     testResponseStatusLine(OkHttpClient());
     testCompressedResponseBody(OkHttpClient());
     testServerErrors(OkHttpClient());
     testIsolate(OkHttpClient.new);
+    testRequestCookies(OkHttpClient(), canSendCookieHeaders: true);
     testResponseCookies(OkHttpClient(), canReceiveSetCookieHeaders: true);
   });
 }
--- a/pkgs/ok_http/jnigen.yaml
+++ b/pkgs/ok_http/jnigen.yaml
@@ -23,6 +23,7 @@ classes:
   - "okhttp3.Call"
   - "okhttp3.Headers"
   - "okhttp3.Callback"
+  - "okhttp3.ConnectionPool"
 
 # Exclude the deprecated methods listed below
 # They cause syntax errors during the `dart format` step of JNIGen.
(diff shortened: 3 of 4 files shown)


pkgs/ok_http/lib/src/ok_http_client.dart
```diff
@@ -126,6 +126,9 @@ class OkHttpClient extends BaseClient {
               request: request,
               contentLength: contentLength,
             ));
+
+            // Close and remove all idle connections from the resource pool.
+            _client.connectionPool().evictAll();
```
brianquinlan, line 131: I read your explanation but I still don't understand why this is necessary. Since the socket is closed by the server after receiving the headers, how can it be reused?

https://github.com/dart-lang/http/blob/9f022d26364fb09748cbca25834b23fdc26baf05/pkgs/http_client_conformance_tests/lib/src/request_cookies_server.dart#L48

Anikate-De: The problem here is that even if the socket is closed, the server keeps listening to the request
```dart
request.listen((line)) {
```
and even when you encounter an empty line, it only adds the `cookies` list to the channel sink.

This seemed very peculiar to me, but I confirmed that it was indeed the problem by doing:
If you create a StreamSubscription and cancel it when an empty line is encountered.
```dart
late StreamSubscription reqSubscription;
reqSubscription = request.listen((line) {
  if (line.toLowerCase().startsWith('cookie:')) {
    cookies.add(line);
  }

  if (line.isEmpty) {
    // A blank line indicates the end of the headers.
    channel.sink.add(cookies);
    reqSubscription.cancel(); // <- here
  }
});
```
then all works well, without the need of explicitly removing idle connections.

Anikate-De: This is what I used for reference: https://stackoverflow.com/questions/31183730/okhttp-asynchronous-request-doesnt-stop-immediately-after-onresponse

brianquinlan: Hmmm... the test seems racy. Does this formulation work:

```dart
await for (final line in request) {
        if (line.toLowerCase().startsWith('cookie:')) {
          cookies.add(line);
        }

        if (line.isEmpty) {
          // A blank line indicates the end of the headers.
          channel.sink.add(cookies);
          break;
        }
}

socket.writeAll(...)
await socket.close()
```

Anikate-De: Yes, it does!

Anikate-De: Do you suggest changing the test case instead? I didn't go that route because `CronetClient` seems to pass both cases without a problem.

brianquinlan: > Do you suggest changing the test case instead? I didn't go that route because `CronetClient` seems to pass both cases without a problem.

Yeah, I think that the test is incorrect and it is a coincidence that it works ;-)

Anikate-De: Alright, then it's good to close this PR. We would still need `okhttp3.ConnectionPool` and `evictAll()` for properly closing the client if I'm not wrong.

brianquinlan: > Alright, then it's good to close this PR. We would still need `okhttp3.ConnectionPool` and `evictAll()` for properly closing the client if I'm not wrong.

You probably want to empty the connection pool when in response to `Client.close()`. From the [docs](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html#:~:text=Shutdown%20Isn't%20Necessary,executor%20service%20with%20shutdown())

Anikate-De: Yeah, that's exactly what I was going to follow.
1. call shutdown on `ExecutorService`
2. empty the connection pool
3. release client object (JNI)

Issue #422, opened 2020-05-23
Best practices on if and when you should close HTTP connections in flutter apps?


anik786:
According to the docs it says:

> If you're making multiple requests to the same server, you can keep open a persistent connection by using a Client rather than making one-off requests. If you do this, make sure to close the client when you're done:

However, in most mobile apps, you tend to send most requests to a single server.

Does that mean, for an app that does a lot of API calls to a server, that we should keep the connection up throughout the app lifetime?

Also should we close and re-open the connection when the app suspends and resume states?

b1acKr0se:
No response? I'm interested in this as well.

muskan911:
i wanna know this too as i have an application with a lot of api requests and which takes time in getting the response and whenever abruptly the screen is changed or another api is called before the earlier request was completed it hangs and shows an error Co nection Closed Before Full Header was received. I hvae been stuck on this for very long any help is appreciated!

cheapestorderconcept:
I have been facing the same issue when making api requests. I couldn't close the connection when my app navigate to other pages which makes my app to be hanging. Any form of help or explanation will be appreciated

mikeesouth:
Hmm. I'm also following this. In my app I have a singleton http client that I initialize once. I then call this client synchronously with get/post/puts. I never close it. This seems to work just fine and I've made several hundreds calls without any issues. However. My end users have reported that the app seemed to stop talking to the server after a while but the animations and buttons were still working. Just that all requests times out. This problem occured after a while and what seems to be at random. I've debugged this for quite some time and at one point I got a state where I called myHttpClient.get() but I did not receive the request server side. After 30 seconds (my configured timeout) the request timed out. I could keep on sending requests to the server but none of these requests reached the server. It seemed like that http client was disposed but I did not get any exception other than requests timing out. So I figured, this is probably what my end users are experiencing. That all http requests suddenly just time out. Could this be because of some pool of requests being "filled up" with broken requests? If the wifi is shaky and a requests break, maybe that pool slot is "broken" and after a while all pool slots are broken and the http client is completely broken?

Would a better practice be to initialize a new HttpClient for every request and closing that after each request? Any downsides?

b1acKr0se:
@mikeesouth one obvious downside is that it's much slower than using an existing client (about 2x slower in my rough testing). I also have a production app that I never close the connection and it seems that it doesn't have your problem.

yinkyAde:
[...]

PR #1234, opened 2024-06-13
Add a note saying that we only create a single `Client`.

brianquinlan:
There was some confusion about the semantics raised in https://github.com/dart-lang/http/issues/422

---

<details>

</details>

### DIFF
--- a/pkgs/flutter_http_example/lib/main.dart
+++ b/pkgs/flutter_http_example/lib/main.dart
@@ -15,6 +15,7 @@ import 'http_client_factory.dart'
 
 void main() {
   runApp(Provider<Client>(
+      // Share a single `Client` throughout the application.
       create: (_) => http_factory.httpClient(),
       child: const BookSearchApp(),
       dispose: (_, client) => client.close()));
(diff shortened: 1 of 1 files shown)


natebosch:
Can you expand this? Maybe add a bit more detail in the comment in source, and elaborate further in the CL description if necessary.

With the current comment, my first assumption was that it means `http_factory.httpClient()` returns a singleton instance and I was confused by the naming.

I _think_ what it actually means is `Provider<Client>` is a singleton provider, and it only calls it's `create` argument once. Is that correct?
I don't know anything about `Provider`, and I don't think we should assume the reader does either. If we're describing some property of that library I think we should make it explicit.

brianquinlan:
> Can you expand this? Maybe add a bit more detail in the comment in source, and elaborate further in the CL description if necessary.
> 
> With the current comment, my first assumption was that it means `http_factory.httpClient()` returns a singleton instance and I was confused by the naming.
> 
> I _think_ what it actually means is `Provider<Client>` is a singleton provider, and it only calls it's `create` argument once. Is that correct? I don't know anything about `Provider`, and I don't think we should assume the reader does either. If we're describing some property of that library I think we should make it explicit.

You're right. I've updated the comment. Is it more clear now?

natebosch (approved):
Very clear now. Thanks!

PR #1242, opened 2024-06-24
pkgs/ok_http: DevTools Networking Support.

Anikate-De:
Items in this PR:

- [x] Add profiling enabled/disabled test groups
- [x] Create Kotlin callback `RedirectReceivedCallback` to allow Dart code to operate on redirect responses (crucial for HTTP Profiling)
- [x] Create `client_profile_test.dart` (re-used from [cronet_http](https://github.com/dart-lang/http/blob/8c325b9ca33d878a86d69c5048a8e6e18379663c/pkgs/cronet_http/example/integration_test/client_profile_test.dart))
- [x] Added necessary implementation in `ok_http_client` to support DevTools
- [x] Updated `CHANGELOG.md`

---

<details>

</details>

### DIFF
--- a/pkgs/ok_http/android/src/main/kotlin/com/example/ok_http/RedirectInterceptor.kt
+++ b/pkgs/ok_http/android/src/main/kotlin/com/example/ok_http/RedirectInterceptor.kt
@@ -11,7 +11,18 @@ package com.example.ok_http
 
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
+import okhttp3.Response
 import java.io.IOException
+
+/**
+ * Callback interface utilized by the [RedirectInterceptor].
+ *
+ * Allows Dart code to operate upon the intermediate redirect responses.
+ */
+interface RedirectReceivedCallback {
+    fun onRedirectReceived(response: Response, location: String)
+}
+
 
 class RedirectInterceptor {
     companion object {
@@ -26,7 +37,10 @@ class RedirectInterceptor {
          * @return OkHttpClient.Builder
          */
         fun addRedirectInterceptor(
-            clientBuilder: OkHttpClient.Builder, maxRedirects: Int, followRedirects: Boolean
+            clientBuilder: OkHttpClient.Builder,
+            maxRedirects: Int,
+            followRedirects: Boolean,
+            redirectCallback: RedirectReceivedCallback,
         ): OkHttpClient.Builder {
             return clientBuilder.addInterceptor(Interceptor { chain ->
                 var req = chain.request()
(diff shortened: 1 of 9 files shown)


pkgs/ok_http/example/integration_test/client_profile_test.dart
Anikate-De, line 1: @brianquinlan 
Would it not be better to have a new package called `http_profile_tests`, (like `http_client_conformance_tests`) that can be used by both `cronet_http` and `ok_http`?

This file has been almost fully re-used from `cronet_http`

pkgs/ok_http/android/src/main/kotlin/com/example/ok_http/RedirectInterceptor.kt
```diff
@@ -11,8 +11,19 @@ package com.example.ok_http
...
+import okhttp3.Response
 import java.io.IOException
 
+/**
+ * Callback interface utilized by the [RedirectInterceptor].
+ *
+ * Allows Dart code to operate upon the intermediate redirect responses.
+ */
+interface RedirectReceivedCallback {
+    fun onRedirectReceived(response: Response, location: String)
```
brianquinlan, line 23: In a follow-up PR, would it make sense to migrate the redirect logic to Dart? Actually, would it make sense to make interceptors available in Dart?

Taking a step back, maybe you'd want to expose `OkHttpClient` or a wrapper around that to Dart?

With that users can control caching, connection pools, DNS, etc.

I'm not sure what the best way to do that is though.

brianquinlan (approved):

pkgs/ok_http/android/src/main/kotlin/com/example/ok_http/RedirectInterceptor.kt
Anikate-De: About migrating the redirection logic to Dart, there are issues with throwing exceptions when the interface (`Interceptor`) that we're implementing, is being consumed on another thread.
Hossein and I have been trying to work our way around this issue, starting with https://github.com/dart-lang/native/pull/1211. But so far, `RedirectInterceptor` stands out as an example of this drawback.

---

About exposing configurations to users, I had planned to provide high-level functions (and constructor parameters) within `OkHttpClient` allowing users to tweak the client according to their requirements.

For example, if a user wants to alter the timeout for a request, they can either pass the timeout duration in the constructor or in the function:
```dart
class OkHttpClient extends BaseClient {
  late bindings.OkHttpClient _client;
  OkHttpClient([Duration? timeoutDuration]) {
    // creates a client based on the timeout and assigns it to _client
  }
  void setTimeout(Duration timeoutDuration) {
    // uses a new builder from `_client` and edits its timeout
  }
```

But yes, a wrapper sounds good! We can use the architecture that `CronetEngine` has.

Issue #828, opened 2022-11-18
runWithClient cannot be used to configure the client in the root Zone


brianquinlan:
[`runWithClient`](https://pub.dev/documentation/http/latest/http/runWithClient.html) is not sufficient for the Flutter use case.

Flutter sometimes invokes callbacks on the Root Zone and `runWithClient` cannot be used to set `Client` in that `Zone`.

natebosch:
It makes me sad if we can't use Zones for the use case they fit best.

brianquinlan:
Fixed in d434d42b4c0a75fb3f4170899e9a30265487fbba

escamoteur:
@kevmoo while doing some intense testing with a network monitor I just had to find out that Flutter ignores the HttpClient that I provide with `runWithClient` in its `NetworkImage` with setting breakpoints and debugging I found out it uses this weird Httpclient implementation of the Flutter Engine
![image](https://github.com/dart-lang/http/assets/16635729/a16d6216-3548-4640-beb3-1ce3d259a7b3)

brianquinlan:
@escamoteur You can look at [flutter_http_example](https://github.com/dart-lang/http/tree/master/pkgs/flutter_http_example) to see how you can plumb a `package:http` `Client` through a Flutter application - including using it to load images.

escamoteur:
I just checked your example, there you too use a different ImageProvider than the built in from Flutter.
Am 25. Juni 2024, 21:09 +0100 schrieb ***@***.***:

brianquinlan:
> I just checked your example, there you too use a different ImageProvider than the built in from Flutter.

That's correct. `package:http_image_provider` allows you to use a `package:http`-based `Client` with `Image`.

escamoteur:
And that was my point, according to the documentation `runWithClient`  should be enough to make Flutter use the HttpClient factory but it does not. The docs even press that it would work with Flutter (https://pub.dev/documentation/http/latest/http/runWithClient.html)  So it's a bug and should be fixed or at least the docs should be updated.

Unfortunately, the package you used doesn't have a linked github repository which isn't really good for an official example. I would propose you just add an ImageProvider like the following (this here uses get_it instead of provider so you don't even have to pass the client to every Image widget) directly inside your demo project. 

```dart
class HttpNetworkImage extends ImageProvider<HttpNetworkImage> {
  /// Creates an object that fetches the image at the given URL.
  const HttpNetworkImage(this.url, {this.scale = 1.0, this.headers});

[...]

PR #1248, opened 2024-06-27
Call out more limitations of runWithClient

natebosch:
Towards #828

Add a section in the `runWithClient` doc that mentions HTTP requests
which _aren't_ affected by usage of this API - specifically requests
made through the SDK core libraries, or through instantiating specific
client implementations.

### DIFF
--- a/pkgs/http/lib/src/client.dart
+++ b/pkgs/http/lib/src/client.dart
@@ -198,6 +198,10 @@ Client? get zoneClient {
 /// and the convenience HTTP functions (e.g. [http.get]). If [clientFactory]
 /// returns `Client()` then the default [Client] is used.
 ///
+/// Only fresh `Client` instances using the default constructor are impacted.
+/// HTTP requests made using `dart:io` or `dart:html` APIs,
+/// or using specifically instantiated client implementations, are not affected.
+///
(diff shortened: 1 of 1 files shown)


pkgs/http/lib/src/client.dart
```diff
@@ -198,6 +198,10 @@ Client? get zoneClient {
 /// and the convenience HTTP functions (e.g. [http.get]). If [clientFactory]
 /// returns `Client()` then the default [Client] is used.
 ///
+/// Only fresh `Client` instances using the default constructor are impacted.
+/// HTTP requests made using `dart:io` or `dart:html` APIs,
+/// or using specifically instantiated client implementations, are not affected.
+///
 /// When used in the context of Flutter, [runWithClient] should be called before
```
brianquinlan, line 205: I was going to replace the Flutter section. Are you working on that or should I keep going?

brianquinlan (approved):

pkgs/http/lib/src/client.dart
natebosch: > Are you working on that or should I keep going?

I wasn't planning on looking more deeply at the Flutter stuff. I noticed this was a gap in the docs, even outside of the specifics of Flutter usage, so I think getting the general message added is a useful and easy first step.

PR #1249, opened 2024-06-27
Document that runWithClient should not be used with Flutter

brianquinlan:
I opted not to examine the possibility of using `Zone.registerCallback` because I think that it would be awkward to use in practice.

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/client.dart
+++ b/pkgs/http/lib/src/client.dart
@@ -202,34 +202,26 @@ Client? get zoneClient {
 /// HTTP requests made using `dart:io` or `dart:html` APIs,
 /// or using specifically instantiated client implementations, are not affected.
 ///
-/// When used in the context of Flutter, [runWithClient] should be called before
-/// [`WidgetsFlutterBinding.ensureInitialized`](https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html)
-/// because Flutter runs in whatever [Zone] was current at the time that the
-/// bindings were initialized.
-///
-/// [`runApp`](https://api.flutter.dev/flutter/widgets/runApp.html) calls
-/// [`WidgetsFlutterBinding.ensureInitialized`](https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html)
-/// so the easiest approach is to call that in [body]:
-/// ```
-/// void main() {
-///  var clientFactory = Client.new; // Constructs the default client.
-///  if (Platform.isAndroid) {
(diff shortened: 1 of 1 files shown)


pkgs/http/lib/src/client.dart
```diff
@@ -202,34 +202,26 @@ Client? get zoneClient {
...
 /// If `no_default_http_client=true` is set then any call to the [Client]
 /// factory (i.e. `Client()`) outside of the [Zone] created by [runWithClient]
 /// will throw [StateError].
+///
+/// > [!IMPORTANT]
+/// > Flutter does not guarantee that callbacks are executed in a particular
+/// > [Zone].
+/// >
+/// > Instead of using [runWithClient], Flutter developers can use a framework,
+/// > such as [`package:provider`](https://pub.dev/packages/provider), to make
```
natebosch, line 220: Does `package:http_client_image_provider` work with `package:provider`?

natebosch (approved):

pkgs/http/lib/src/client.dart
brianquinlan: Yep, we even have an usage example in the demo app:
https://github.com/dart-lang/http/blob/bf96551406600a2ec520a4248ff635bfd268edeb/pkgs/flutter_http_example/lib/main.dart#L141

The name `HttpImage` used by `package:http_client_image_provider` seems a bit unfortunate to me. I wonder if proposing to rename it to `HttpClientImageProvider` upstream would be worth it?

natebosch: I don't have an opinion on renaming - but this doc change LGTM given that usage in the example.

ueman: Maintainer of `package:http_image_provider` here: I don't mind the proposed name change. It should be pretty straightforward to change anyway. Just rename it, create an alias for the old name and mark the alias as deprecated.

brianquinlan: I started working on a `package:http_image_provider` PR yesterday but then I thought that the naming confusion is probably mostly caused by the fact that the `Image` widget calls its `ImageProvider` argument `image`.

So you have:
```dart
Image(image: HttpImage(...))  // Looks like we are nesting an images.
```

I guess that:
```dart
Image(image: HttpImageProvider(...)) // Slightly better?
```

is slightly better but really, the `image` parameter should be called something like `provider` so it reads like:

```dart
Image(provider: ...) // Now I actually understand that we are provider the data here.
```

it is probably still worth doing though and maybe I'll file a flutter issue as well

PR #1255, opened 2024-07-03
Clarify when Client.close must be called

brianquinlan:
I'm not happy with this phrasing so any advice would be appreciated.

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/client.dart
+++ b/pkgs/http/lib/src/client.dart
@@ -148,8 +148,10 @@ abstract interface class Client {
 
   /// Closes the client and cleans up any resources associated with it.
   ///
-  /// It's important to close each client when it's done being used; failing to
-  /// do so can cause the Dart process to hang.
+  /// Some clients may maintain a pool of network connections that will not
+  /// be disconnected until the client is closed. This may cause programs run
+  /// using the Dart SDK (but **not** the Flutter SDK) to not terminate until
+  /// the client is closed.
   ///
   /// Once [close] is called, no other methods should be called. If [close] is
   /// called while other asynchronous methods are running, the behavior is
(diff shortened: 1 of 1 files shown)


pkgs/http/lib/src/client.dart
```diff
@@ -148,8 +148,10 @@ abstract interface class Client {
 
   /// Closes the client and cleans up any resources associated with it.
   ///
-  /// It's important to close each client when it's done being used; failing to
-  /// do so can cause the Dart process to hang.
+  /// Some clients may maintain a pool of network connections that will not
+  /// be disconnected until the client is closed. This may cause programs run
```
natebosch, line 152: [nit] We don't need to hedge twice with "Some" and "may". Also I think it reads better adjusting the tense in the middle of the first sentence.

```suggestion
  /// Some clients maintain a pool of network connections that are not
  /// disconnected until the client is closed. This may cause programs run
```

pkgs/http/lib/src/client.dart
```diff
@@ -148,8 +148,10 @@ abstract interface class Client {
...
+  /// using the Dart SDK (but **not** the Flutter SDK) to not terminate until
+  /// the client is closed.
```
natebosch, line 154: WDYT about mentioning more specifically the commands?

```suggestion
  /// be disconnected until the client is closed. This may cause programs
  /// using `dart run` or `dart compile` (but **not** `flutter run` or `flutter build`) to not terminate until
  /// the client is closed.
```

natebosch (approved):

pkgs/http/lib/src/client.dart
brianquinlan: I rephrased slightly.

natebosch (approved):

PR #1286, opened 2024-08-06
Add a workflow for publishing flutter packages

### LINKED ISSUE #1259: Enable firehouse for `package:http`

natebosch:
Closes #1259

Add a second copy of the publish workflow with `use-flutter: true`.
Add `ignore-packages` to both workflows to ignore published packages for
the non-relevant SDK.

### DIFF
--- a/.github/workflows/publish_flutter.yaml
+++ b/.github/workflows/publish_flutter.yaml
@@ -1,6 +1,6 @@
 # A CI configuration to auto-publish pub packages.
 
-name: Publish Dart Packages
+name: Publish Flutter Packages
 
 on:
   pull_request:
@@ -15,4 +15,5 @@ jobs:
     permissions:
       id-token: write # Required for authentication using OIDC
       pull-requests: write # Required for writing the pull request note
-    ignore-packages: pkgs/cupertino_http/,pkgs/cronet_http/,pkgs/ok_http/
+    use-flutter: true
+    ignore-packages: pkgs/http/,pkgs/http_profile/,pkgs/web_socket/,


.github/workflows/publish.yaml
```diff
@@ -12,6 +12,8 @@ jobs:
   publish:
     if: ${{ github.repository_owner == 'dart-lang' }}
     uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
+    with:
+      ignore-packages: pkgs/cupertino_http,pkgs/cronet_http,pkgs/ok_http
     permissions:
       id-token: write # Required for authentication using OIDC
```
natebosch, line 18: @devoncarew - Are these config lines stale? I don't see the same in the other repos I checked.

kevmoo: Latest from here: https://github.com/dart-lang/ecosystem/blob/main/pkgs/blast_repo/lib/src/tweaks/auto_publish_tweak.dart

.github/workflows/publish_flutter.yaml
```diff
@@ -0,0 +1,20 @@
...
+    branches: [ master ]
+  push:
+    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+' ]
+
+jobs:
+  publish:
+    if: ${{ github.repository_owner == 'dart-lang' }}
+    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
+    with:
+      use-flutter: true
```
kevmoo, line 16: Seems like something we should auto-detect, right? Maybe?

natebosch: It did initially, and then that functionality was removed.
https://github.com/dart-lang/ecosystem/pull/158

I'm not sure I understand the "We need the flag anyhow to decide whether to setup Flutter in the first place" statement. @mosuem do you recall what this meant?

mosuem: It means that instead of setting up dart via `setup-dart`, we would use the `setup-flutter` action.

brianquinlan:
Is this ready to land?

mosuem (approved):

brianquinlan:
@natebosch Is this something that you intend to land?

PR #1291, opened 2024-08-09
Fix "unintended_html_in_doc_comment" analysis errors

brianquinlan:
---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/base_response.dart
+++ b/pkgs/http/lib/src/base_response.dart
@@ -117,9 +117,9 @@ var _headerSplitter = RegExp(r'[ \t]*,[ \t]*');
 ///
 /// Set-Cookie strings can contain commas. In particular, the following
 /// productions defined in RFC-6265, section 4.1.1:
-/// - <sane-cookie-date> e.g. "Expires=Sun, 06 Nov 1994 08:49:37 GMT"
-/// - <path-value> e.g. "Path=somepath,"
-/// - <extension-av> e.g. "AnyString,Really,"
+/// - `<sane-cookie-date>` e.g. "Expires=Sun, 06 Nov 1994 08:49:37 GMT"
+/// - `<path-value>` e.g. "Path=somepath,"
+/// - `<extension-av>` e.g. "AnyString,Really,"
 ///
 /// Some values are ambiguous e.g.
 /// "Set-Cookie: lang=en; Path=/foo/"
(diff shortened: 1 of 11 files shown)


pkgs/http/lib/src/base_response.dart
```diff
@@ -128,8 +128,9 @@ var _headerSplitter = RegExp(r'[ \t]*,[ \t]*');
 /// "Set-Cookie: lang=en; Path=/foo/,SID=x23"
 /// would both be result in `response.headers` => "lang=en; Path=/foo/,SID=x23"
 ///
-/// The idea behind this regex is that ",<valid token>=" is more likely to
-/// start a new <cookie-pair> then be part of <path-value> or <extension-av>.
+/// The idea behind this regex is that `,<valid token>=` is more likely to
+/// start a new `<cookie-pair>` then be part of `<path-value>` or
```
Anikate-De, line 132:
```suggestion
/// start a new `<cookie-pair>` than be part of `<path-value>` or
```

pkgs/http_profile/lib/src/utils.dart
```diff
@@ -10,13 +10,13 @@ const _tokenChars = r"!#$%&'*+\-.0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ^_`"
 /// Splits comma-separated header values.
 var _headerSplitter = RegExp(r'[ \t]*,[ \t]*');
 
-/// Splits comma-separated "Set-Cookie" header values.
+/// Splits comma-seperated "Set-Cookie" header values.
```
Anikate-De, line 13: nit:
```suggestion
/// Splits comma-separated "Set-Cookie" header values.
```

pkgs/http_profile/lib/src/utils.dart
```diff
@@ -25,8 +25,9 @@ var _headerSplitter = RegExp(r'[ \t]*,[ \t]*');
 /// "Set-Cookie: lang=en; Path=/foo/,SID=x23"
 /// would both be result in `response.headers` => "lang=en; Path=/foo/,SID=x23"
 ///
-/// The idea behind this regex is that ",<valid token>=" is more likely to
-/// start a new <cookie-pair> then be part of <path-value> or <extension-av>.
+/// The idea behind this regex is that `,<valid token>=` is more likely to
+/// start a new `<cookie-pair>` then be part of `<path-value>` or
```
Anikate-De, line 29: nit:
```suggestion
/// start a new `<cookie-pair>` than be part of `<path-value>` or
```

Anikate-De (changes requested):

Anikate-De (approved):

natebosch:
@devoncarew This PR highlights a false positive for the lint. It doesn't understand square braces well enough and it looks like it flags `[List<int>]`. I think the change to ``` `List<int>` ``` is an improvement, but only slight compared to the obvious improvement from broken HTML. I didn't click on any of these cases when I skimmed the internal results, and the rollout is more churn than I expected.

I also forgot about the unpublished packages where they're unlikely to be noticed since we don't have any workflow where they are visible to users. I don't know whether any IDEs try to render the dartdoc to html for display, but that's the only use case I can imagine is affected.

natebosch (approved):
http and http_profile can both be bumped to the next version with `-wip`.

devoncarew:
> @devoncarew This PR highlights a false positive for the lint. It doesn't understand square braces well enough and it looks like it flags `[List<int>]`. I think the change to `` `List<int>` `` is an improvement, but only slight compared to the obvious improvement from broken HTML. I didn't click on any of these cases when I skimmed the internal results, and the rollout is more churn than I expected.

Gotcha; can you file an issue in the sdk for this? We have had some other recent false positives (https://github.com/dart-lang/sdk/issues/56450) addressed - now fixed in head.

natebosch:
https://github.com/dart-lang/sdk/issues/56465

brianquinlan:
> http and http_profile can both be bumped to the next version with `-wip`.

Done

natebosch (approved):

PR #1307, opened 2024-09-24
Migrate to the dart_style DartFormatter API.

### LINKED ISSUE #1304: Please start passing in a language version to dart_style
Greetings http team friends!

The formatter is [moving to being language version aware](https://github.com/dart-lang/dart_style/issues/1402). This means that when it's parsing some code, it needs to be told what language version to parse it as. The `DartFormatter` constructor now takes [an optional parameter where you can pass in the language version](https://github.com/dart-lang/dart_style/blob/003cb15b8d27e4a72eb25dba6f695cd2c7a6f4fb/lib/src/dart_formatter.dart#L83). In a future version of dart_style, that parameter will become mandatory.

If you own code that uses dart_style as a library, you should update your code to pass in a language version. To migrate:

1. Update your constraint on dart_style to `^2.3.7`. That's the version where the new parameter was added.

2. If you know the precise language version the code should be parsed as, then pass in that version (as an instance of the [pub_semver package's `Version` class](https://pub.dev/documentation/pub_semver/latest/pub_semver/Version-class.html)). This is what "real" tools should do.

    For simple one-off scripts and other utilities where precise behavior doesn't matter much, you can pass in [`DartFormatter.latestLanguageVersion`](https://github.com/dart-lang/dart_style/blob/003cb15b8d27e4a72eb25dba6f695cd2c7a6f4fb/lib/src/dart_formatter.dart#L30) to unconditionally parse the code as the latest language version that the formatter itself supports.

When the `--tall-style` experiment flag ships, the passed in language version will also be used to determine whether you get the current formatting style or the [new "tall" style](https://github.com/dart-lang/dart_style/issues/1253).

The constructor calls that I think are relevant are:

```
http/pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart:
   47:   final formatter = DartFormatter();

http/pkgs/web_socket_conformance_tests/bin/generate_server_wrappers.dart:
   40:   final formatter = DartFormatter();
```

Thank you!

munificent:
Going forward, the formatter needs to know what language version it should use to parse input code. This updates the generator to pass in a version. (In this case, since it's just used for code generation, it unconditionally passes in the latest language version, which should be fine.)

Fix #1304.

### DIFF
--- a/pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart
+++ b/pkgs/http_client_conformance_tests/bin/generate_server_wrappers.dart
@@ -44,7 +44,8 @@ Future<StreamChannel<Object?>> startServer() async => spawnHybridUri(Uri(
 
 void main() async {
   final files = await Directory('lib/src').list().toList();
-  final formatter = DartFormatter();
+  final formatter =
+      DartFormatter(languageVersion: DartFormatter.latestLanguageVersion);
 
   files.where((file) => file.path.endsWith('_server.dart')).forEach((file) {
     final vmPath = file.path.replaceAll('_server.dart', '_server_vm.dart');
--- a/pkgs/http_client_conformance_tests/pubspec.yaml
+++ b/pkgs/http_client_conformance_tests/pubspec.yaml
@@ -11,7 +11,7 @@ environment:
 
 dependencies:
   async: ^2.8.2
-  dart_style: ^2.2.3
+  dart_style: ^2.3.7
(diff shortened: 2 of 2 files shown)


brianquinlan (approved):

munificent:
Do you know what's up with those analysis failures? Can I ignore those?

brianquinlan:
> Do you know what's up with those analysis failures? Can I ignore those?

I think that the failures are unrelated to your change.

PR #1401, opened 2024-11-08
Switch browser_client.dart to use fetch

mraleph:
Rewrite `browser_client.dart` to use `fetch` instead of XHR.

### DIFF
--- a/pkgs/http/CHANGELOG.md
+++ b/pkgs/http/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 1.2.3-wip
 
-* Fixed unintended HTML tags in doc comments. 
+* Fixed unintended HTML tags in doc comments.
+* Switched `BrowserClient` to use Fetch API instead of `XMLHttpRequest`.
 
 ## 1.2.2
 
--- a/pkgs/http/CHANGELOG.md
+++ b/pkgs/http/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.2.3-wip
+## 1.3.0-wip
 
 * Fixed unintended HTML tags in doc comments.
 * Switched `BrowserClient` to use Fetch API instead of `XMLHttpRequest`.
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -97,7 +97,7 @@ class BrowserClient extends BaseClient {
 
       final headers = <String, String>{};
       (response.headers as _IterableHeaders)
-          .forEach((String value, String header) {
+          .forEach((String value, String header, [JSAny? _]) {
         headers[header.toLowerCase()] = value;
       }.toJS);
 
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -10,6 +10,7 @@ import 'package:web/web.dart'
         AbortController,
         HeadersInit,
         ReadableStreamDefaultReader,
+        Response,
         RequestInit,
         window;
 
@@ -30,9 +31,9 @@ BaseClient createClient() {
 }
 
 /// A `package:web`-based HTTP client that runs in the browser and is backed by
-/// [fetch].
+/// `window.fetch`.
 ///
-/// This client inherits some limitations of [fetch].
+/// This client inherits some limitations of `window.fetch`.
 /// - [BaseRequest.persistentConnection];
 /// - Setting [BaseRequest.followRedirects] to `false` will cause
 /// `ClientException` when redirect is encountered;
@@ -95,48 +96,6 @@ class BrowserClient extends BaseClient {
         );
       }
 
-      final bodyStreamReader =
-          response.body?.getReader() as ReadableStreamDefaultReader?;
-      Stream<List<int>> _readLoop() async* {
-        if (bodyStreamReader == null) {
-          return;
-        }
-
-        var isDone = false, isError = false;
-        try {
-          while (true) {
-            final chunk = await bodyStreamReader.read().toDart;
-            if (chunk.done) {
-              isDone = true;
-              break;
-            }
-            yield (chunk.value! as JSUint8Array).toDart;
-          }
-        } catch (e, st) {
-          isError = true;
-          _rethrowAsClientException(e, st, request);
-        } finally {
-          if (!isDone) {
-            try {
-              // catchError here is a temporary workaround for
-              // http://dartbug.com/57046: an exception from cancel() will
-              // clobber an exception which is currently in flight.
-              await bodyStreamReader
-                  .cancel()
-                  .toDart
-                  .catchError((_) => null, test: (_) => isError);
-            } catch (e, st) {
-              // If we have already encountered an error swallow the
-              // error from cancel and simply let the original error to be
-              // rethrown.
-              if (!isError) {
-                _rethrowAsClientException(e, st, request);
-              }
-            }
-          }
-        }
-      }
-
       final headers = <String, String>{};
       (response.headers as _IterableHeaders)
           .forEach((String value, String header) {
@@ -144,7 +103,7 @@ class BrowserClient extends BaseClient {
       }.toJS);
 
       return StreamedResponseV2(
-        _readLoop(),
+        _readBody(request, response),
         response.status,
         headers: headers,
         request: request,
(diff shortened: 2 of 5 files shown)


brianquinlan:
Nice! 

There are some lints to fix here.
Do you want to increase the browser test coverage?
I also noticed that you aren't supporting streaming uploads - presumably that is because it is not well supported across browser versions?

mraleph:
Fixing the lints now. I am a bit surprised that I don't get them in the IDE. 

> I also noticed that you aren't supporting streaming uploads - presumably that is because it is not well supported across browser versions?

Yeah, I am typing the comment on the issue now. tldr is that even when browser supports streaming request it can still refuse to do it if the server is not HTTP/2 or QUIC... Which means we can't really use it by default because it will potentially cause existing code to fail (even if we sniff browser compatibility - which by itself is a non-trivial thing to do :-/). I am honestly surprised how nasty browser implementation is - it does not even provide any useful information in the error message when this happens and you need to go look into the console to see `ERR_H2_OR_QUIC_REQUIRED` there.

pkgs/http/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 1.2.3-wip
```
brianquinlan, line 1: Since this is a feature release, you should bump this to 1.3.0-wip

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
 }
 
 /// A `package:web`-based HTTP client that runs in the browser and is backed by
-/// [XMLHttpRequest].
+/// `window.fetch`.
```
brianquinlan, line 34:
```suggestion
/// [`window.fetch`](https://fetch.spec.whatwg.org/).
```

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
...
+///
+/// This client inherits some limitations of `window.fetch`.
+/// - [BaseRequest.persistentConnection];
+/// - Setting [BaseRequest.followRedirects] to `false` will cause
+/// `ClientException` when redirect is encountered;
```
brianquinlan, line 39:
```suggestion
///   `ClientException` when a redirect is encountered;
```

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
 }
 
 /// A `package:web`-based HTTP client that runs in the browser and is backed by
-/// [XMLHttpRequest].
+/// `window.fetch`.
+///
+/// This client inherits some limitations of `window.fetch`.
+/// - [BaseRequest.persistentConnection];
```
brianquinlan, line 37: Do you mean that this is ignored?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
 }
 
 /// A `package:web`-based HTTP client that runs in the browser and is backed by
-/// [XMLHttpRequest].
+/// `window.fetch`.
+///
+/// This client inherits some limitations of `window.fetch`.
```
brianquinlan, line 36:
```suggestion
/// This client inherits some limitations of `window.fetch`:
```

pkgs/http/lib/src/browser_client.dart
brianquinlan: (or does the markdown get formatted correctly without the indent?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
...
+/// - [BaseRequest.persistentConnection];
+/// - Setting [BaseRequest.followRedirects] to `false` will cause
+/// `ClientException` when redirect is encountered;
+/// - Positive values of [BaseRequest.maxRedirects] are ignored, non-positive
+/// values are treated the same as setting [BaseRequest.followRedirects] to
```
brianquinlan, line 41:
```suggestion
///   values are treated the same as setting [BaseRequest.followRedirects] to
```

pkgs/http/lib/src/browser_client.dart
```diff
@@ -27,18 +31,20 @@ BaseClient createClient() {
...
+/// `false`.
```
brianquinlan, line 42:
```suggestion
///   `false`.
```

brianquinlan (approved):
I want to be you when I grow up.

pkgs/http/lib/src/browser_client.dart
mraleph: It does get formatted correctly but I applied your suggestion because it makes things easier to read.

pkgs/http/lib/src/browser_client.dart
mraleph: Yeah, I started writing a comment then, then went to check the doc if we can implement the necessary behavior, but then forgot what I was doing :) Fixed it now.

mraleph:
I will wait landing this before we can get dev coverage. Don't want to land it with a broken CI. 

> Do you want to increase the browser test coverage?

Do you have something specific in mind? 

I have tested firefox and safari locally and both work (though `package:test` is currently broken in different ways for these two browsers - `-p firefox` tries to use a binary which does not exist anymore and `-p safari` requires user intervention[^1] to run tests).

[^1]: I wasted couple of hours trying to come up with some way to launch and shutdown Safari cleanly after tests - but the only thing that really works requires using `safaridriver`. There is no clean equivalent to a simple "open target URL in new process" we get in other browsers.

brianquinlan:
> I will wait landing this before we can get dev coverage. Don't want to land it with a broken CI.
> 
> > Do you want to increase the browser test coverage?
> 
> Do you have something specific in mind?
> 
> I have tested firefox and safari locally and both work (though `package:test` is currently broken in different ways for these two browsers - `-p firefox` tries to use a binary which does not exist anymore and `-p safari` requires user intervention[1](#user-content-fn-1-edcc263dae46d6f13b074aaf1720aff2) to run tests).

Yeah, that was what I was thinking - but I guess that won't work out right now. It looks like the checks are exceeding except for these tests:
https://github.com/dart-lang/http/actions/runs/11778289469/job/32896956901?pr=1401

> ## Footnotes
> 1. I wasted couple of hours trying to come up with some way to launch and shutdown Safari cleanly after tests - but the only thing that really works requires using `safaridriver`. There is no clean equivalent to a simple "open target URL in new process" we get in other browsers. [↩](#user-content-fnref-1-edcc263dae46d6f13b074aaf1720aff2)

mraleph:
I fixed 3.4 tests

brianquinlan (approved):

kevmoo:
What's the associated bug? Can fix as well?

jezell:
Thanks @mraleph! This is great.

I think reading request into memory still means it's going to blow up even when StreamedRequest could be streamed:

https://github.com/mraleph/http/blob/49f3b3cd83b38944722c2d4a1d60b70c1fb21eb9/pkgs/http/lib/src/browser_client.dart#L64

But, at least it puts the parts in place to enable some solutions to those problems. 

Think what needs to be figured out is how do we get something other than a byte array to here:

https://github.com/mraleph/http/blob/49f3b3cd83b38944722c2d4a1d60b70c1fb21eb9/pkgs/http/lib/src/browser_client.dart#L71

Not sure the proper platform neutral abstraction for the library, but if there was a way to go something like:

```dart
import "package:http/web.dart";
import "package:web/web.dart" as web;

client.send(WebRequest(web.RequestBody));
```

[RequestBody](https://pub.dev/documentation/fetch_api/latest/fetch_api/RequestBody-extension-type.html)

At least that would provide some kind of path to do a file upload or manually pass a stream if you want.

mraleph:
I am planning to look at request streaming as well, because I have already loaded the context into my head so it makes sense to do something before it gets paged out.

csells:
I just wasted a day of my life debugging one of the issues that this PR fixes (surfacing data as it streams in instead of all at once when the connection is closed) only to find that this PR already fixes my issue. Any chance you'd be willing to publish this fix on pub.dev? That would be great.

cc: @mraleph @kevmoo

Reprevise:
> I just wasted a day of my life debugging one of the issues that this PR fixes (surfacing data as it streams in instead of all at once when the connection is closed) only to find that this PR already fixes my issue. Any chance you'd be willing to publish this fix on pub.dev? That would be great.

[http 1.3.0](https://pub.dev/packages/http/changelog) has been released which includes this PR.

csells:
> [http 1.3.0](https://pub.dev/packages/http/changelog) has been released which includes this PR.

I noticed. And it works great! Thank you very much.

gazb23:
Thanks so much - have been banging my head against the wall for a few hours and then found this pull request!

Issue #175, opened 2018-07-13
Decode with utf8 by default for non-text (or all?) content types


cdvv7788:
I am requesting some information from the a server, which returns the following (using postman):

Headers:
Allow →GET, HEAD, OPTIONS
CF-RAY →439e4801cf3db955-MIA
Connection →keep-alive
Content-Encoding →gzip
Content-Type →application/json

Body:
{
    "name": "SARA LUCIA OSSA PEÑA",
}

But, using http, I am getting the following (using a simple request get('url')):
{
    "name": "SARA LUCIA OSSA PEÃA",
}

To fix this, I had to do something like:
`UTF8.decode(response.bodyBytes)`

This works as expected, and the information is retrieved fine. This, however, is a pain to setup (and inconsistent with post, where utf8 is used as default encoding).

Is there a better way to handle this? An argument to the get parameter to force encoding? shouldn't application/json assume utf8 by default?

I came up with the solution after reading https://pub.dartlang.org/documentation/http/latest/http/Response-class.html and the body property. Probably it is encoding the body with a wrong format.

Anyway, thanks for the hard work. Awesome library.

zoechi:
Good comments about this in
https://stackoverflow.com/questions/9254891/what-does-content-type-application-json-charset-utf-8-really-mean

I don't think it's the HTTP clients job to do the decoding. The header is no guarantee that the content will be of that type.
You can always build your own wrapper that does the UTF8-decoding for you so you don't have to repeat yourself.

cdvv7788:
@zoechi The client has to try to decode using a best effort approach. The http client is already decoding, but using the wrong encoding. There are 2 attributes in the response, body and bodyBytes.
In the same link you send, it is mentioned:

> Designating the encoding is somewhat redundant for JSON, since the default (only?) encoding for JSON is UTF-8

Knowing that the content type is json should be enough for the client to interpret it, or at least to avoid intepreting it with the wrong encoding. This is not an edge case, json is probably the most popular format for api communication at the moment. You are right on the wrapper part, but if the headers are coherent, the client should be able to handle the body properly too.

cskau-g:
So, the situation is a bit murky, so bear with me..

The first layer in any of this is HTTP.
HTTP, AFAICT, defines the [default character set to be ISO-8859-1](https://www.w3.org/International/articles/http-charset/index), which is why the Content-Type charset parameter exists as an explicit override.
[...]

PR #1421, opened 2024-12-12
Fix Default Encoding for application/json Content-Type in HTTP Responses

snoopdoggy322:
This merge request addresses a long-standing issue (dart-lang/http#175) where the http package defaults to latin1 encoding when no charset is specified in the Content-Type header. Per [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259), the default encoding for application/json should be utf-8.

The fix ensures that responses with Content-Type: application/json and no explicit charset parameter are decoded using utf8 by default. This change improves compliance with the JSON standard and prevents decoding errors for valid JSON payloads encoded in utf-8.


snoopdoggy322:
closing as broken brach, sorry

PR #1422, opened 2024-12-12
Fix Default Encoding for application/json Content-Type in HTTP Responses

snoopdoggy322:
This merge request addresses a long-standing issue (https://github.com/dart-lang/http/issues/175) where the http package defaults to latin1 encoding when no charset is specified in the Content-Type header. Per [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259), the default encoding for application/json should be utf-8.

The fix ensures that responses with Content-Type: application/json and no explicit charset parameter are decoded using utf8 by default. This change improves compliance with the JSON standard and prevents decoding errors for valid JSON payloads encoded in utf-8.


snoopdoggy322:
can anyone reply to my pull request?

seifibrahim32:
It is fine but, could you try to pass some Chinese or Korean language tests through conformance tests? @snoopdoggy322

kevmoo:
CC @brianquinlan @devoncarew 

Ooo...could be good

pkgs/http/lib/src/response.dart
```diff
@@ -21,10 +21,10 @@ class Response extends BaseResponse {
   ///
   /// This is converted from [bodyBytes] using the `charset` parameter of the
   /// `Content-Type` header field, if available. If it's unavailable or if the
-  /// encoding name is unknown, [latin1] is used by default, as per
-  /// [RFC 2616][].
+  /// encoding name is unknown, [utf8] is used by default, as per
+  /// [RFC3629][].
```
brianquinlan, line 25: Could you site a section? Is this only for JSON data for for all payloads?

snoopdoggy322: Yes, RFC7231 requires interpreting application/json as UTF-8, everything else is described as incorrect interpretation of the specification

pkgs/http/test/response_test.dart
```diff
@@ -45,6 +46,23 @@ void main() {
           headers: {'content-type': 'text/plain; charset=iso-8859-1'});
       expect(response.body, equals('föøbãr'));
     });
+    test('test decoding with empty charset if content type is application/json',
+        () {
+      final utf8Bytes = utf8.encode('{"foo":"Привет, мир!"}');
+      var response = http.Response.bytes(utf8Bytes, 200,
+          headers: {'content-type': 'application/json'});
+      expect(response.body, equals('{"foo":"Привет, мир!"}'));
```
brianquinlan, line 54: Ideally, there should be one `except` per test because you are testing a single scenario.

Could you add tests for these additional scenarios:
1. test `.body` with a non-empty charset with content-type application/json data (e.g. with Russian and iso-8859-1)
2. test `.body` with an empty charset with a content-type of test/plain (e.g. with Russian to verify that the default of iso-8859-1 is used)

pkgs/http/lib/src/response.dart
```diff
@@ -66,10 +66,12 @@ class Response extends BaseResponse {
 
 /// Returns the encoding to use for a response with the given headers.
 ///
-/// Defaults to [latin1] if the headers don't specify a charset or if that
-/// charset is unknown.
+/// If the `Content-Type` header specifies a charset, it will use that charset.
+/// If no charset is provided or the charset is unknown:
+/// - Defaults to [utf8] if the `Content-Type` is `application/json` (since JSON is defined to use UTF-8 by default).
```
brianquinlan, line 71: Wrap to 80 columns

pkgs/http/lib/src/response.dart
```diff
@@ -21,10 +21,10 @@ class Response extends BaseResponse {
   ///
   /// This is converted from [bodyBytes] using the `charset` parameter of the
   /// `Content-Type` header field, if available. If it's unavailable or if the
-  /// encoding name is unknown, [latin1] is used by default, as per
-  /// [RFC 2616][].
+  /// encoding name is unknown, [utf8] is used by default,
```
brianquinlan, line 24: This is only when content-type is UTF-8 right? So could you clarify that in your comments e.g.

If it's unavailable or if the encoding name is unknown:
1. [utf8] is used when the content-type is `'application/json'` (see [RFC 3629]).
2. [latin1] is used in all other cases (see [RFC 2616][]).

brianquinlan (commented):
Excellent work, we are almost there!

Would you also:
1. change the version in pubspec.yaml to version: 1.4.0-wip
2. add a 1.4.0-wip section to CHANGELOG.md with a description of your change

pkgs/http/lib/src/response.dart
snoopdoggy322: @brianquinlan, thanks for the review.
Let me know if anything else needs improvement! 🚀

pkgs/http/lib/src/response.dart
```diff
@@ -21,9 +21,11 @@ class Response extends BaseResponse {
   ///
   /// This is converted from [bodyBytes] using the `charset` parameter of the
   /// `Content-Type` header field, if available. If it's unavailable or if the
-  /// encoding name is unknown, [latin1] is used by default, as per
-  /// [RFC 2616][].
+  /// encoding name is unknown:
+  /// - [utf8] is used when the content-type is 'application/json' (see [RFC 3629]).
```
brianquinlan, line 25:
```suggestion
  /// - [utf8] is used when the content-type is 'application/json' (see [RFC 3629][]).
```

pkgs/http/lib/src/response.dart
```diff
@@ -21,9 +21,11 @@ class Response extends BaseResponse {
...
+  /// - [latin1] is used in all other cases (see [RFC 2616][])
   ///
+  /// [RFC 3629]:https://www.rfc-editor.org/rfc/rfc3629.
```
brianquinlan, line 28:
```suggestion
  /// [RFC 3629]: https://www.rfc-editor.org/rfc/rfc3629.
```

brianquinlan (changes requested):
Could you update from master so I can run the CI?

pkgs/http/lib/src/response.dart
brianquinlan: I think that this is great! Thank you so much for your contribution!

brianquinlan (approved):

PR #1706, opened 2025-02-02
Added cache control for http request within BrowserClient

seifibrahim32:
Added cache control for BrowserClient and passed 85 conformance tests including fixes in tests for html runtime.


README.md
```diff
@@ -6,7 +6,7 @@ A composable, Future-based library for making HTTP requests.
 easy to consume HTTP resources. It's multi-platform, and supports mobile, desktop,
 and the browser.
 
-## Packages
+## Packages
```
brianquinlan, line 9: Remove space

pkgs/http/lib/src/base_request.dart
```diff
@@ -165,4 +165,9 @@ abstract class BaseRequest {
 
   @override
   String toString() => '$method $url';
+
+  /// [cache] property for caching the network request.
+  ///
+  /// Defaults to `default`.
+  String? get cache => HttpCacheOptions.defaultType;
```
brianquinlan, line 172: I don't think that this belongs here since it only applies to `BrowserClient`.  `package:cronet_http` forces the user to set the cache policy for the entire `Client`:
https://pub.dev/documentation/cronet_http/latest/cronet_http/CronetEngine-class.html

Also, why is the type nullable?

brianquinlan (commented):
Please add tests for this. There are cache testing examples here:
https://github.com/dart-lang/http/blob/c7cc4f8e19dd0a050fcf5654b120d789663c48e4/pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart#L12

pkgs/http/lib/src/base_request.dart
seifibrahim32: You are correct, I am thinking of using BrowserClient only that conforms to that option too.

Let me make another commit with cache parameter in the Client.

seifibrahim32:
@brianquinlan

pkgs/http/lib/src/browser_client.dart
```diff
@@ -45,6 +45,10 @@ BaseClient createClient() {
 class BrowserClient extends BaseClient {
   final _abortController = AbortController();
 
+  BrowserClient([CacheOption? cacheOption]) {
```
brianquinlan, line 48: I think that this should probably be a keyword parameter, with a type of `String` and a default of `"default"`. You should also add documentation this this (check out `IOClient` for an example). You can link to https://developer.mozilla.org/en-US/docs/Web/API/Request/cache for the possible values.

You need to include tests in this PR.

seifibrahim32: @brianquinlan Chcek the recent update.

pkgs/http/lib/src/utils.dart
```diff
@@ -68,3 +68,22 @@ Stream<T> onDone<T>(Stream<T> stream, void Function() onDone) =>
       sink.close();
       onDone();
     }));
+
+/// Caching modes for any http request in a browser.
+enum CacheMode {
```
brianquinlan, line 73: I think that this is unnecessary.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -53,6 +64,11 @@ class BrowserClient extends BaseClient {
 
   bool _isClosed = false;
 
+  /// Use different caching mode for a HTTP request.
+  /// Defaults to `default`.
+  // https://developer.mozilla.org/en-US/docs/Web/API/Request/cache
+  String cacheMode;
```
brianquinlan, line 70: This can be private and you can move the documentation to the constructor documentation.

brianquinlan (commented):
Remember to write tests! ;-)

pkgs/http/lib/src/browser_client.dart
```diff
@@ -53,6 +64,11 @@ class BrowserClient extends BaseClient {
 
   bool _isClosed = false;
 
+  /// Use different caching mode for a HTTP request.
+  /// Defaults to `default`.
+  // https://developer.mozilla.org/en-US/docs/Web/API/Request/cache
```
kevmoo, line 69: I'd ask this NOT be buried in a normal comment. A doc comment so folks can go see the supported values would be GREAT!

pkgs/http/lib/src/browser_client.dart
kevmoo: ...and `final`!

pkgs/http/lib/src/browser_client.dart
```diff
@@ -45,6 +44,18 @@ BaseClient createClient() {
...
 
+  /// Create a new browser-based HTTP Client.
+  ///
+  /// If [cacheMode] is provided then it can be used to cache the request
+  /// in the browser.
+  ///
+  /// For example:
+  /// ```dart
+  /// const mode = 'reload';
+  /// final client = BrowserClient(cacheMode: mode);
```
kevmoo, line 55:
```suggestion
  /// final client = BrowserClient(cacheMode: 'reload');
```

kevmoo: it's cleaner

pkgs/http/lib/src/utils.dart
kevmoo, line 73: Where is this used?

pkgs/http/lib/src/utils.dart
```diff
@@ -68,3 +68,22 @@ Stream<T> onDone<T>(Stream<T> stream, void Function() onDone) =>
...
+  noStore('no-store'),
+  noCache('no-cache'),
+  forceCache('force-cache'),
+  onlyIfCached('only-if-cached');
+
+  final String cacheType;
+
+  const CacheMode(this.cacheType);
+
+  static CacheMode fromString(String cacheType) => values.firstWhere(
```
kevmoo, line 85: Can use https://api.dart.dev/beta/latest/dart-core/EnumByName/byName.html

pkgs/http/lib/src/browser_client.dart
seifibrahim32: I see triple slashes /// would go rather than normal comment but, I try to pass the dart analyze so, I would be sure that the quality comes first.

pkgs/http/lib/src/utils.dart
seifibrahim32: We can use it if the user needs to use the caching modes with existing utilities instead of writing everything as String and pass it.

seifibrahim32: @brianquinlan I guess I would move with Kevin for making those cache modes available instead of allowing the feature to make it pass any String.

brianquinlan: OK. Please move this to `browser_client.dart`

pkgs/http/lib/src/utils.dart
seifibrahim32: Since we have this function so we won't need fromString at all.

seifibrahim32:
@brianquinlan I will add tests but, I want to make sure that everything goes good within the design.
@kevmoo I am thinking of changing from String to CacheOption and we get the type within the enum ?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -45,6 +44,22 @@ BaseClient createClient() {
...
+  /// in the browser.
+  ///
+  /// For example:
+  /// ```dart
+  /// final client = BrowserClient(cacheMode: 'reload');
+  /// ```
+  ///
+  /// Defaults to `default`.
+  /// https://developer.mozilla.org/en-US/docs/Web/API/Request/cache
+  BrowserClient([String? cacheMode]) : _cacheMode = cacheMode ?? 'default';
```
brianquinlan, line 61: Why does this need to be nullable?

```suggestion
  BrowserClient({CacheMode cacheMode = CacheMode.defaultType}) : _cacheMode = cacheMode;
```

pkgs/http/lib/src/browser_client.dart
```diff
@@ -184,3 +200,17 @@ Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
 extension type _IterableHeaders._(JSObject _) implements JSObject {
   external void forEach(JSFunction fn);
 }
+
+/// Caching modes for any http request in a browser.
+enum CacheMode {
```
brianquinlan, line 205: Please move this above BrowserClient

pkgs/http/lib/src/browser_client.dart
```diff
@@ -184,3 +200,17 @@ Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
 extension type _IterableHeaders._(JSObject _) implements JSObject {
   external void forEach(JSFunction fn);
 }
+
+/// Caching modes for any http request in a browser.
```
brianquinlan, line 204:
```suggestion
/// Caching mode used by the [BrowserClient].
///
/// Sets the value of the browser Fetch API
/// [`Request.cache`](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache) property.
```

brianquinlan (changes requested):

seifibrahim32:
I need to make a directory for integration tests so I can make some tests @brianquinlan similar to Cronet's client integration tests.

Since we should test those tests on browser runtime!.

I tried to run client_test.dart via dart test -p chrome and it is impossible to run also, adding to conformance tests won't help since We should test passing values to BrowserClient constructor.

brianquinlan:
> @brianquinlan I need to make a directory for integration tests so I can make some tests similar to Cronet's client integration tests.

No, you don't need a separate directory for these tests.

> Since we should test those tests on browser runtime!.
> 
> I tried to run client_test.dart via dart test -p chrome and it is impossible to run also, adding to conformance tests won't help since We should test passing values to BrowserClient constructor.

You don't want to update the conformance tests, just the unit tests, which can be run with `dart test -p chrome  test/html/`

You can add your tests to `test/html/client_test.dart` or make a new file in `test/html` (e.g. `cache_test.dart` if you have lots of tests to add.

seifibrahim32:
I need to enable the server itself and work within the browser.
Are there any steps I should do since I could easily add better tests.?

brianquinlan:
> I need to enable the server itself and work within the browser. Are there any steps I should do since I could easily add better tests.?
> 
> @brianquinlan
> 
> Would you like to use any external api e.g. Google Books?

No will need to write your own server to verify that the expected requests are sent. You can refer to the cronet tests for reference.

seifibrahim32:
> > I need to enable the server itself and work within the browser. Are there any steps I should do since I could easily add better tests.?
> > 
> > @brianquinlan
> > 
> > Would you like to use any external api e.g. Google Books?
> 
> No will need to write your own server to verify that the expected requests are sent. You can refer to the cronet tests for reference.

So, you need me to create any server using Flask/Node.js or whatever to ensure ?

seifibrahim32:
> > I need to enable the server itself and work within the browser. Are there any steps I should do since I could easily add better tests.?
> > 
> > @brianquinlan
> > 
> > Would you like to use any external api e.g. Google Books?
> 
> No will need to write your own server to verify that the expected requests are sent. You can refer to the cronet tests for reference.

I don't understand , you mean I dont need to have a server from scratch to verify mine.!

The idea is Cronet unit tests aren't browser-based also, HttpServer is `dart:io` import means that JS interop will reject it

brianquinlan:
> > > I need to enable the server itself and work within the browser. Are there any steps I should do since I could easily add better tests.?
> > > @brianquinlan
> > > Would you like to use any external api e.g. Google Books?
> > 
> > 
> > No will need to write your own server to verify that the expected requests are sent. You can refer to the cronet tests for reference.
> 
> I don't understand , you mean I dont need to have a server from scratch to verify mine.!
> 
> The idea is Cronet unit tests aren't browser-based also, HttpServer is `dart:io` import means that JS interop will reject it

You can use [`spawnHybridCode`](https://api.flutter.dev/flutter/flutter_test/spawnHybridCode.html) or `spawnHybridUri`. You can find some examples here:
https://github.com/search?q=repo%3Adart-lang%2Fhttp+spawnHybrid&type=code

seifibrahim32:
@brianquinlan I have added some tests.

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+    expectLater(client.send(request), throwsA(isA<ClientException>()));
+    request.sink.add('{"hello": "world"}'.codeUnits);
+    unawaited(request.sink.close());
+
+    client.close();
+  }, skip: 'Need to fix server tests for browser');
+
+  test('#send with an invalid URL', () {
+    var client = BrowserClient(cacheMode: CacheMode.onlyIfCached);
```
brianquinlan, line 74: What is this testing?

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+import 'package:http/browser_client.dart';
+import 'package:http/http.dart' as http;
+import 'package:http/src/exception.dart';
+import 'package:test/test.dart';
+
+import 'utils.dart';
+
+void main() {
+  test('#send a StreamedRequest with default type', () async {
```
brianquinlan, line 18: Do any of these tests require streaming? If you look at the cronet tests:
https://github.com/dart-lang/http/blob/master/pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart

You'll see that they are confirming that the server receives the expected number of requests.

brianquinlan (changes requested):

pkgs/http/test/html/cache_test.dart
seifibrahim32: Test that the client ensures of smooth request with invalid url successfully.
So it ensures that fetch API did the process.

pkgs/http/test/html/cache_test.dart
seifibrahim32: the server is IO-based not web based. so, we can try to manuever those tests but adding retry parameter in the test() function.

brianquinlan (commented):
Hey @seifibrahim32 

This PR is requiring a lot of review effort. Could you please only ping me when you think that it is fully ready for review? Meaning that all of the tests are in place and pass, the code has the correct style, the new functionality is documented, and you've updated the CHANGELOG and pubspec.yaml?

seifibrahim32:
Sure, the PR needs alot of efforts since we try to use browser and dart:io unfortunately can't hold on with the browser setup.

seifibrahim32:
@brianquinlan Hello, I am back once again.
The tests are ready..
Waiting to hear back from you.

pkgs/http/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.5.0
```
brianquinlan, line 1: You can move this into the 1.4.0-wip section.

And remove "with fixing tests".

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
brianquinlan, line 1: 2025

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  test('#send a GET with default type', () async {
+    var client = BrowserClient(cacheMode: CacheMode.defaultType);
+    await client.get(url);
+    var response = await client.get(url);
+    client.close();
+
+    expect(response.statusCode, 200);
+    expect(response.reasonPhrase, 'OK');
+    expect(response.body, parse(allOf(containsPair('numOfRequests', 2))));
```
brianquinlan, line 33: Shouldn't this be one? For the fetch api, do you need to manually set etags?

pkgs/http/test/html/cache_test.dart
brianquinlan, line 1: I think that you should have a GET request example for every cache type.

pkgs/http/CHANGELOG.md
seifibrahim32: Ok

pkgs/http/test/html/cache_test.dart
seifibrahim32: Ok

pkgs/http/test/html/cache_test.dart
seifibrahim32: Nope, 2 because it is called twice.

Check cronet cache tests as well for reference.

seifibrahim32: What do you mean by etags ?
You mean the timeout skipping comments ?

If yes, I guess yes but overall it works.

brianquinlan: I thought that `CacheMode.defaultType` enabled caching? So why is the server accessed twice is caching is enabled? 

When caching is enabled in `cronet_http`, only one request is seen by the server:
https://github.com/dart-lang/http/blob/900da9fe7d576caa4882053cf26d3a70bd3106e0/pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart#L48

You can see that the test server for cronet_http (https://github.com/dart-lang/http/blob/900da9fe7d576caa4882053cf26d3a70bd3106e0/pkgs/cronet_http/example/integration_test/cronet_configuration_test.dart#L24)

sets both the `Cache-Control` and `etag` headers.

The whatwg spec has information on how etags are used:

image

seifibrahim32: So, I think I will push another test with different etag values.

You are correct.

seifibrahim32: Lets call it regression test to check a time when server enables caching on a specific route.

seifibrahim32: @brianquinlan I found out it is one not 2 when etag value is inserted.

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,83 @@
...
+void main() {
+  late Uri url;
+  setUp(() async {
+    final channel = spawnHybridUri(Uri(path: '/test/stub_server.dart'));
+    var port = (await channel.stream.first as num).toInt();
+    url = echoUrl.replace(port: port);
+  });
+
+  test('#send a GET with default type', () async {
+    var client = BrowserClient(cacheMode: CacheMode.defaultType);
```
brianquinlan, line 26: You are only doing a single request here so I'm not sure how this is testing the caching behavior.

pkgs/http/test/html/cache_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  test('#send a GET with no-cache type', () async {
+    var client = BrowserClient(cacheMode: CacheMode.noCache);
+
+    await client.get(url);
+    var response = await client.get(url);
+    client.close();
+    expect(
+        response.body,
+        parse(anyOf(containsPair('numOfRequests', 2),
```
brianquinlan, line 52: Shouldn't `numOfRequests` always be 2?

pkgs/http/test/html/cache_test.dart
seifibrahim32: Nope, it is two GET requests, when it is set to default type it is cached twice.

pkgs/http/test/html/cache_test.dart
seifibrahim32: Yes, see most of other tests.

seifibrahim32: Do you have other suggestions regarding the tests? @brianquinlan

pkgs/http/test/html/cache_test.dart
seifibrahim32: I guess, I need to drain the request so, I can see the difference.

brianquinlan:
Hi @seifibrahim32 

This PR is simply taking too much effort to land.

seifibrahim32:
I will work on it soon because I was a little bit busy.

seifibrahim32:
@brianquinlan

brianquinlan:
Hi @seifibrahim32, 

Even if you find time to work on this, the problem is that it is taking too much effort to review. If you find time to work on this and are confident that you have addressed all pending issues, please reopen.

Cheers,
Brian

seifibrahim32:
Ok reopen it please and I will start working on them. 🙏@brianquinlan

Because I have no access to re-open it again at the moment unless I closed it.

PR #1715, opened 2025-02-12
Fix browser clients in workers

simolus3:
The browser client implementations calls `window.fetch()` to send HTTP requests. This makes it impossible to use the package in contexts where `window` is unavailable, such as web workers or Node.JS.

Instead of using `window`, this PR changes the implementation to just call `fetch()` on the global scope.

### DIFF
--- a/pkgs/http/CHANGELOG.md
+++ b/pkgs/http/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 * Fixed default encoding for application/json without a charset
   to use utf8 instead of latin1, ensuring proper JSON decoding.
+* Avoid references to `window` in `BrowserClient`, restoring support for web
+  workers and NodeJS.
 
-## 1.3.0-wip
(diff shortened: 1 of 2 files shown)


brianquinlan (commented):
Thank you!

Could you add a note to CHANGELOG.md in the 1.4.0-wip section, like:

* Fixes a bug where the `window` object was used, making it impossible to use with web workers and Node.js.

pkgs/http/CHANGELOG.md
```diff
@@ -2,8 +2,10 @@
 
 * Fixed default encoding for application/json without a charset
   to use utf8 instead of latin1, ensuring proper JSON decoding.
+* Avoid references to `window` in `BrowserClient`, restoring support for web
+  workers and NodeJS.
 
-## 1.3.0-wip
```
brianquinlan, line 6: Nice catch!

PR #1716, opened 2025-02-12
Http-parser: Make CaseInsensitiveMap be extension type.

lrhn:
Avoids extending a class unnecessarily, which *may* pave the way for `package:collection` making the class final.

(Nothing uses that `CaseInsensitiveMap` is a type at all, the constructors could all just be functions returning `Map`.)

Change is not breaking, the type is not publicly visible.

### DIFF
--- a/pkgs/http_parser/CHANGELOG.md
+++ b/pkgs/http_parser/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.1.3
+
+* Makes `CaseInsensitiveMap` an extension type.
+  Allows `package:collection` to make the class `final`.
+
 ## 4.1.2
 
 * Fixed a bug where parsing quoted header values could require a regex to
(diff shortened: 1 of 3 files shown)


brianquinlan (approved):
That's neat! My first introduction to extension types.

kevmoo:
Hold a sec. Strictly, this is breaking right? Folks might implement/extend this?

Maybe we say that's okay, but we should at least discuss.

brianquinlan:
> Hold a sec. Strictly, this is breaking right? Folks might implement/extend this?
> 
> Maybe we say that's okay, but we should at least discuss.

Good point! 

@lrhn said that this type is not [publicly visible but it seems to be](https://github.com/dart-lang/http/blob/e28f9f5ec65d83d0bfb299a78c9365cc4dc38ba1/pkgs/http_parser/lib/http_parser.dart#L6)

The only instances of `extends CaseInsensitiveMap`/`implements CaseInsensitiveMap` that I found on github are false positives.

brianquinlan:
@lrhn Do you want to land this?

lrhn:
It's safer not to land.

The type *is* exported by `lib/http_parser.dart`, so I was wrong that it's not public.

It probably *shouldn't* be exported &mdash; its interface is the same as `Map`, so it's not needed as a type, and it doesn't appear as a type in any other public API. 
Nobody should ever *want* to extend it, but until we decide to remove that export that, this would be breaking.

I'll close, and see if I can remember to reopen if things have changed.

PR #1725, opened 2025-03-06
Add custom message when terminating a connection

Javier-Novales:
Add support for providing a custom message when terminating a connection

This allows for clients to get customized messages when connections are terminated early.

### DIFF
--- a/pkgs/http2/CHANGELOG.md
+++ b/pkgs/http2/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## 2.3.2-wip
+- Add support for providing custom message when terminating a connection.
 
 ## 2.3.1
 
--- a/pkgs/http2/lib/transport.dart
+++ b/pkgs/http2/lib/transport.dart
@@ -73,7 +73,7 @@ abstract class TransportConnection {
   Future finish();
 
   /// Terminates this connection forcefully.
-  Future terminate([int? errorCode]);
+  Future terminate([int? errorCode, String? message]);
 }
 
 abstract class ClientTransportConnection extends TransportConnection {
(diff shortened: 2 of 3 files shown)


pkgs/http2/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## 2.3.2-wip
+- Add support for providing custom message when terminating a connection.
```
brianquinlan, line 2: Nit: add a blank line before the first bullet.

pkgs/http2/lib/transport.dart
```diff
@@ -73,7 +73,7 @@ abstract class TransportConnection {
   Future finish();
 
   /// Terminates this connection forcefully.
-  Future terminate([int? errorCode]);
+  Future terminate([int? errorCode, String? message]);
```
brianquinlan, line 76: Since this class is not `final`, changing the signature of this method is probably breaking. I don't know how important that is.

brianquinlan (commented):
Is it possible to write tests for this?

pkgs/http2/lib/transport.dart
mosuem: Filed https://github.com/bmw-tech/dart_apitool/issues/207 - I do believe this is a breaking change, so this requires a version rev to `3.0.0`. We should use that opportunity to do some changes like revving the minimum SDK version.

mosuem:
As this is a breaking change, we have to update the major version number according to https://semver.org/spec/v2.0.0-rc.1.html. If we do that, we might do some other major version changes to go with it.

So I filed #1745 which will upgrade the used version of the Dart SDK and set the new version to `3.0.0-wip`. Once that is landed, you should `git rebase master` your PR and replace `3.0.0-wip` with `3.0.0` both in the `pubspec.yaml` and the `CHANGELOG.md`. This will enable us to publish the new version.

mosuem (approved):

PR #1766, opened 2025-05-05
Ignore received data after the response stream has been closed

brianquinlan:

<details>

</details>

### DIFF
--- a/pkgs/cupertino_http/lib/src/cupertino_client.dart
+++ b/pkgs/cupertino_http/lib/src/cupertino_client.dart
@@ -56,6 +56,9 @@ class _TaskTracker {
   final responseCompleter = Completer<URLResponse>();
   final BaseRequest request;
   final StreamController<Uint8List> responseController;
+
+  /// Whether the response stream subscription has been cancelled.
+  bool responseListenerCancelled = false;
   final HttpClientRequestProfile? profile;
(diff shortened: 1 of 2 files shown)


pkgs/cupertino_http/lib/src/cupertino_client.dart
```diff
@@ -366,6 +371,9 @@ class CupertinoClient extends BaseClient {
     request.headers.forEach(urlRequest.setValueForHttpHeaderField);
     final task = urlSession.dataTaskWithRequest(urlRequest);
     final subscription = StreamController<Uint8List>(onCancel: () {
+      final taskTracker = _tasks[task];
+      if (taskTracker == null) return;
```
liamappelbe, line 375: This will skip `task.cancel()` if there's no tracker. Is that intended? If not, I think you can simplify this to `_tasks[task]?.responseListenerCancelled = true;`

liamappelbe (approved):

pkgs/cupertino_http/lib/src/cupertino_client.dart
brianquinlan: The tracker is removed when the request is completed, in which case cancelling the response stream is a no-op (because no more data will arrive).

PR #1772, opened 2025-05-19
Reorder content-disposition in multi file

lwj1994:
resort content-type plz. to fix uplaod file error


brianquinlan:
Could you explain what your issue is and how this PR fixes it?

lwj1994:
@brianquinlan Reordered the multipart headers to place Content-Disposition before Content-Type . While RFC 7578 implies header order should not matter, Aliyun OSS (and many other server-side parsers) strictly expects Content-Disposition to be the first header for efficient parsing. This change aligns the behavior with the de-facto standard used by browsers and cURL.

PR #1773, opened 2025-05-20
[WIP] feat: support abortion of in-flight HTTP requests

### LINKED ISSUE #204: Cancellable requests

### LINKED ISSUE #424: Design an API to support aborting requests

### LINKED ISSUE #567: Abort or cancel multipart request

JaffaKetchup:
Resolves #424, closes #204, closes #424, closes #567
Closes #978, closes #521, closes #602, closes #819

As discussed in #424, this adds an `Abortable` interface for `BaseRequests` that provides an `abortTrigger` which `Client`s may use to abort requests.

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/abortable.dart
+++ b/pkgs/http/lib/src/abortable.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'base_request.dart';
+import 'client.dart';
+
+/// Enables a request to be recognised by a [Client] as abortable
+abstract mixin class Abortable implements BaseRequest {
+  /// This request will be aborted if this completes
+  ///
+  /// A common pattern is aborting a request when another event occurs (such as
+  /// a user action). A [Completer] may be used to implement this.
+  ///
+  /// Another pattern is a timeout. Use [Future.delayed] to achieve this.
+  ///
+  /// This future must not complete to an error.
+  ///
+  /// This future may complete at any time - a [AbortedRequest] will be thrown
+  /// by [send]/[Client.send] if it is completed before the request is complete.
+  ///
+  /// Non-'package:http' [Client]s may unexpectedly not support this trigger.
+  abstract final Future<void>? abortTrigger;
+}
+
+/// Thrown when a request is aborted using [Abortable.abortTrigger]
+///
+/// This is not thrown when `abortTrigger` is completed after the request has
+/// already completed.
+class AbortedRequest implements Exception {
+  /// Indicator that the request has been aborted by [Abortable.abortTrigger]
+  const AbortedRequest();
+}
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -14,6 +14,7 @@ import 'package:web/web.dart'
         RequestInit,
         Response;
 
+import 'abortable.dart';
 import 'base_client.dart';
 import 'base_request.dart';
 import 'exception.dart';
@@ -69,6 +70,11 @@ class BrowserClient extends BaseClient {
 
     final bodyBytes = await request.finalize().toBytes();
     try {
+      Future<void>? canceller;
+      if (request case Abortable(:final Future<void> abortTrigger)) {
+        canceller = abortTrigger.whenComplete(() => _abortController.abort());
+      }
+
       final response = await _fetch(
         '${request.url}'.toJS,
         RequestInit(
@@ -85,6 +91,8 @@ class BrowserClient extends BaseClient {
           redirect: request.followRedirects ? 'follow' : 'error',
         ),
       ).toDart;
+
+      canceller?.ignore();
 
       final contentLengthHeader = response.headers.get('content-length');
 
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -4,6 +4,7 @@
 
 import 'dart:io';
 
+import 'abortable.dart';
 import 'base_client.dart';
 import 'base_request.dart';
 import 'base_response.dart';
@@ -123,7 +124,15 @@ class IOClient extends BaseClient {
         ioRequest.headers.set(name, value);
       });
 
+      Future<void>? canceller;
+      if (request case Abortable(:final Future<void> abortTrigger)) {
+        canceller = abortTrigger
+            .whenComplete(() => ioRequest.abort(const AbortedRequest()));
+      }
+
       var response = await stream.pipe(ioRequest) as HttpClientResponse;
+
+      canceller?.ignore();
 
       var headers = <String, String>{};
       response.headers.forEach((key, values) {
--- a/pkgs/http/example/main.dart
+++ b/pkgs/http/example/main.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert' as convert;
 
 import 'package:http/http.dart' as http;
@@ -9,6 +10,14 @@ void main(List<String> arguments) async {
       Uri.https('www.googleapis.com', '/books/v1/volumes', {'q': '{http}'});
 
   // Await the http get response, then decode the json-formatted response.
+  try {
+    final e =
+        await http.AbortableRequest('GET', url, abortTrigger: Future.error(10))
+            .send();
+    await e.stream.drain<void>();
+  } on http.AbortedRequest {
+    print('Cancelled');
+  }
   var response = await http.get(url);
   if (response.statusCode == 200) {
     var jsonResponse =
--- a/pkgs/http/lib/http.dart
+++ b/pkgs/http/lib/http.dart
@@ -19,6 +19,7 @@ export 'src/base_request.dart';
 export 'src/base_response.dart'
     show BaseResponse, BaseResponseWithUrl, HeadersWithSplitValues;
 export 'src/byte_stream.dart';
+export 'src/abortable.dart';
 export 'src/client.dart' hide zoneClient;
 export 'src/exception.dart';
 export 'src/multipart_file.dart';
--- a/pkgs/http/lib/src/base_request.dart
+++ b/pkgs/http/lib/src/base_request.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'package:meta/meta.dart';
 
 import '../http.dart' show ClientException, get;
+import 'abortable.dart';
 import 'base_client.dart';
 import 'base_response.dart';
 import 'byte_stream.dart';
@@ -20,6 +21,10 @@ import 'utils.dart';
 /// [BaseClient.send], which allows the user to provide fine-grained control
 /// over the request properties. However, usually it's easier to use convenience
 /// methods like [get] or [BaseClient.get].
+///
+/// Subclasses/implementers should mixin/implement [Abortable] to support
+/// abortion of requests. A future breaking version of 'package:http' will
+/// merge [Abortable] into [BaseRequest], making it a requirement.
 abstract class BaseRequest {
   /// The HTTP method of the request.
   ///
--- a/pkgs/http/lib/src/multipart_request.dart
+++ b/pkgs/http/lib/src/multipart_request.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 import 'dart:math';
 
+import 'abortable.dart';
 import 'base_request.dart';
 import 'boundary_characters.dart';
 import 'byte_stream.dart';
@@ -159,4 +160,16 @@ class MultipartRequest extends BaseRequest {
         growable: false);
     return '$prefix${String.fromCharCodes(list)}';
   }
+}
+
+/// A [MultipartRequest] which supports abortion using [abortTrigger]
+///
+/// A future breaking version of 'package:http' will merge this into
+/// [MultipartRequest], making it a requirement.
+final class AbortableMultipartRequest extends MultipartRequest with Abortable {
+  AbortableMultipartRequest(super.method, super.url, {this.abortTrigger})
+      : super();
+
+  @override
+  final Future<void>? abortTrigger;
 }
--- a/pkgs/http/lib/src/request.dart
+++ b/pkgs/http/lib/src/request.dart
@@ -7,6 +7,7 @@ import 'dart:typed_data';
 
 import 'package:http_parser/http_parser.dart';
 
+import 'abortable.dart';
 import 'base_request.dart';
 import 'byte_stream.dart';
 import 'utils.dart';
@@ -181,4 +182,15 @@ class Request extends BaseRequest {
     if (!finalized) return;
     throw StateError("Can't modify a finalized Request.");
   }
+}
+
+/// A [Request] which supports abortion using [abortTrigger]
+///
+/// A future breaking version of 'package:http' will merge this into [Request],
+/// making it a requirement.
+final class AbortableRequest extends Request with Abortable {
+  AbortableRequest(super.method, super.url, {this.abortTrigger}) : super();
+
+  @override
+  final Future<void>? abortTrigger;
 }
--- a/pkgs/http/lib/src/streamed_request.dart
+++ b/pkgs/http/lib/src/streamed_request.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'abortable.dart';
 import 'base_client.dart';
 import 'base_request.dart';
 import 'byte_stream.dart';
@@ -52,4 +53,16 @@ class StreamedRequest extends BaseRequest {
     super.finalize();
     return ByteStream(_controller.stream);
   }
+}
+
+/// A [StreamedRequest] which supports abortion using [abortTrigger]
+///
+/// A future breaking version of 'package:http' will merge this into
+/// [StreamedRequest], making it a requirement.
+final class AbortableStreamedRequest extends StreamedRequest with Abortable {
+  AbortableStreamedRequest(super.method, super.url, {this.abortTrigger})
+      : super();
+
+  @override
+  final Future<void>? abortTrigger;
 }


brianquinlan:
The outline looks good so far! Let me know if you need me to press the button to run the workflows.

pkgs/http_client_conformance_tests/pubspec.yaml
```diff
@@ -3,18 +3,21 @@ description: >-
...
-  dart_style: '>=2.3.7 <4.0.0'
+  dart_style: ">=2.3.7 <4.0.0"
   http: ^1.2.0
   stream_channel: ^2.1.1
   test: ^1.21.2
 
+dependency_overrides:
+  http:
+    path: ../http
+
```
JaffaKetchup, line 21: Need to remove this, but not sure of how to make the tests work without it. Once published ofc, the dependency can be bumped to `^1.5.0`.

pkgs/http_client_conformance_tests/lib/src/abort_server.dart
```diff
@@ -0,0 +1,43 @@
...
+///     - close current request
+///     - exit server
+void hybridMain(StreamChannel<Object?> channel) async {
+  final channelQueue = StreamQueue(channel.stream);
+
+  late HttpServer server;
+  server = (await HttpServer.bind('localhost', 0))
+    ..listen((request) async {
+      // TODO: might have to ignore exceptions in the server because it will
+      // probably be disconnected
```
JaffaKetchup, line 26: This doesn't seem to be necessary, but I'll leave it up to you.

pkgs/http_client_conformance_tests/lib/http_client_conformance_tests.dart
```diff
@@ -90,6 +94,8 @@ void testAll(
   bool canSendCookieHeaders = false,
   bool canReceiveSetCookieHeaders = false,
   bool supportsMultipartRequest = true,
+  // TODO: make this false, for now true to see what breaks.
+  bool supportsAbort = true,
```
JaffaKetchup, line 98: I'll leave this up to you, not sure what's best.

pkgs/http/lib/src/abortable.dart
```diff
@@ -0,0 +1,43 @@
...
+  ///
+  /// Some clients may not support abortion, or may not support this trigger.
+  abstract final Future<void>? abortTrigger;
+}
+
+/// Thrown when an HTTP request is aborted
+///
+/// Usually, this is due to [Abortable.abortTrigger]. See documentation on that
+/// property for more info.
+class AbortedRequest implements Exception {
```
brianquinlan, line 40: I think that this should `extend ClientException` - what would mean that the request URL is accessible and that code can just catch `ClientException` if it doesn't want to differentiate between cancellation and other errors.

What do you think?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -158,6 +180,14 @@ Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
       }
       yield (chunk.value! as JSUint8Array).toDart;
     }
+  } on DOMException catch (e, st) {
+    isError = true;
+
+    if (e.name == 'AbortError') {
+      Error.throwWithStackTrace(const AbortedRequest(), st);
```
brianquinlan, line 184: If `AbortedRequest` becomes a `ClientException` then you could move this logic to `_rethrowAsClientException`.

pkgs/http/lib/src/io_client.dart
```diff
@@ -123,7 +125,45 @@ class IOClient extends BaseClient {
...
       });
 
-      var response = await stream.pipe(ioRequest) as HttpClientResponse;
+      // We can only abort the actual connection up until the point we obtain
+      // the response.
+      // After that point, the full response bytes are always available.
+      // However, we instead inject an error into the response stream to match
+      // the behaviour of `BrowserClient`.
+
+      StreamSubscription<List<int>>? subscription;
```
brianquinlan, line 134: maybe `ioResponseSubscription`?

pkgs/http/lib/src/io_client.dart
```diff
@@ -123,7 +125,45 @@ class IOClient extends BaseClient {
...
+      final controller = StreamController<List<int>>(sync: true);
```
brianquinlan, line 135: Maybe `responseController`?

brianquinlan (changes requested):
Amazing work. I have a few nits and I'd like to patch this in and try it out myself before we merge it.

pkgs/http_client_conformance_tests/pubspec.yaml
JaffaKetchup:
```suggestion
# TODO(brianquinlan): Remove dependency_overrides when package:http 1.5.0 is released.
dependency_overrides:
  http:
    path: ../http

```

JaffaKetchup:
> I have a few nits

just a few :D

> try it out myself before we merge it.

Sounds good. I have already tested it in flutter_map - the plan is to deprecate the mess we have at the moment of support Dio to do this with a separate 'plugin' package.

---

In terms of `RetryClient` - might need some help here.

Swapping it in place of the `IOClient` does cause two tests to fail. 

> Invalid argument:  <- Instance of '_SyncStreamController<List<int>>' (from dart:async)
 <- Instance of 'StreamedRequest' (from package:http/src/streamed_request.dart)
 <- bodyBytes in Instance of 'Response' (from package:http/src/response.dart)
 <- _List len:1 (from dart:core)
is unsendable object (see restrictions listed at`SendPort.send()` documentation for more information): Instance of 'Future<void>'
dart:isolate  _RemoteRunner._run

I'm not quite sure where isolates are getting involved here. But I suspect it's failing because it's not handling splitting the request body stream well on cancellation.

> TimeoutException after 0:00:30.000000: Test timed out after 30 seconds. See https://pub.dev/packages/test#timeouts
dart:isolate  _RawReceivePort._handleMessage

I assume this is the retry client again not fully listening or closing something, meaning the response stream never fully completes. I guess it's again related to the splitting.

I have no good idea on how to fix these, so any pointers would be appriciated!

brianquinlan:
> > I have a few nits
> 
> just a few :D
> 
> > try it out myself before we merge it.
> 
> Sounds good. I have already tested it in flutter_map - the plan is to deprecate the mess we have at the moment with a seperate 'plugin' package which uses Dio to do this. Would be good to get it merged this week if possible - and maybe a release as well 👀

No promises but I'll try. I was trying to implement request cancellation in the two other clients in this repo. So far I've succeeded with `package:cupertino_http`: https://github.com/dart-lang/http/pull/1779

> In terms of `RetryClient` - might need some help here.
> [snipped]

I'll take a look at this tomorrow.

brianquinlan:
I've updated your PR to fix the analysis errors. It looks like two tests are flaky:

1.

```
test/io/client_conformance_test.dart: abort during request stream (failed)
  TimeoutException after 0:00:30.000000: Test timed out after 30 seconds. See https://pub.dev/packages/test#timeouts
  dart:isolate  _RawReceivePort._handleMessage
```

2.

```
00:02 +60 ~3 -1: test/io/client_test.dart: detachSocket returns a socket from an IOStreamedResponse [E]
  Null check operator used on a null value
  dart:isolate  _RawReceivePort._handleMessage
```

Let's deal with those issues before the `RetryClient` issues.

@JaffaKetchup can you look at those?

brianquinlan:
You can see both failures in this CI run: https://github.com/dart-lang/http/actions/runs/15307371338/job/43064107042?pr=1773

JaffaKetchup:
I'm not sure why it would be working on mine and failing on CI - I guess CI is just 30 seconds slower which somehow causes the issue.

How would we cancel the request stream? Put it through another internal controller, and inject an error into it on abortion, which I think will make `stream.pipe(ioRequest)` throw? I'm not really sure.

brianquinlan:
> I'm not sure why it would be working on mine and failing on CI - I guess CI is just 30 seconds slower which somehow causes the issue.
> 
> How would we cancel the request stream? Put it through another internal controller, and inject an error into it on abortion, which I think will make `stream.pipe(ioRequest)` throw? I'm not really sure.

I haven't looked into how request cancellation in `dart:io`.

JaffaKetchup:
I'm assuming the test is hanging because the input stream is never `done`. Somehow on my environment, it must be done. What happens if that last await is removed?

Unfortunately I've got more limited time coming up over the next couple weeks, so I'm happy to implement but don't really have too much time for experimentation.

brianquinlan:
> I'm assuming the test is hanging because the input stream is never `done`. Somehow on my environment, it must be done. What happens if that last await is removed?
> 
> Unfortunately I've got more limited time coming up over the next couple weeks, so I'm happy to implement but don't really have too much time for experimentation.

I'm on holiday for two weeks starting this Friday so I don't have a lot of time in the short term either.

My intuition was that getting this to work reliably would be difficult, given that we have multiple implementations and cancellation can happen at any point in the request.

Let me make a suggestion on https://github.com/dart-lang/http/issues/424

JaffaKetchup:
I can't be certain, but it looks like that fixed the issue?

pkgs/http_client_conformance_tests/lib/src/abort_tests.dart
```diff
@@ -0,0 +1,168 @@
...
+      final request = AbortableStreamedRequest(
+        'POST',
+        serverUrl,
+        abortTrigger: abortTrigger.future,
+      );
+
+      final response = client.send(request);
+      request.sink.add('Hello World'.codeUnits);
+
+      abortTrigger.complete();
```
brianquinlan, line 74: Closing the sink will break if we change the code to:
```
      expect(
        response,
        throwsA(isA<AbortedRequest>()),
      );

      for (int i = 0; i < 10000; ++i) {
        request.sink.add('Hello World'.codeUnits);
        await Future.delayed(const Duration());
      }
```

```
  Bad state: Cannot add event after closing
```

JaffaKetchup: So my suggestion would be to either:
 - make this a forbidden action (the user shouldn't be trying to pipe in after aborting)
 - make a dummy consumer which silently switches away from the aborted request and drops incoming data

brianquinlan: The docs for `abort` say:

  /// Using the [IOSink] methods (e.g., [write] and [add]) has no effect after
  /// the request has been aborted

JaffaKetchup: Not at my computer right now so can't investigate, but does addStream use that? I suspect it might use a more efficient method on the stream consumer (but I might be wrong).

brianquinlan: Maybe my test should just considered malformed for not closing the sink.

brianquinlan: I think that I changed my mind about this. Since streaming uploads are not supported by `BrowserClient`, it will never cancel the stream subscription. So I don't think that we need to do so in `IOClient`.

brianquinlan:
`detachSocket returns a socket from an IOStreamedResponse` in test/io/client_test.dart seems to be flaky with this PR but not on master. I have no idea why.

brianquinlan:
> `detachSocket returns a socket from an IOStreamedResponse` in test/io/client_test.dart seems to be flaky with this PR but not on master. I have no idea why.

I think that the problem is that we now call `response.listener` even if no one is listening to the returned `StreamedResponse`.

brianquinlan:
This change seems to fix the flake: https://github.com/dart-lang/http/pull/1773/commits/9001a1d7580f8586fef6d2a610083c999dd8a574

pkgs/http/lib/src/abortable.dart
```diff
@@ -0,0 +1,43 @@
...
+  ///
+  /// Some clients may not support abortion, or may not support this trigger.
+  abstract final Future<void>? abortTrigger;
+}
+
+/// Thrown when an HTTP request is aborted.
+///
+/// This exception is triggered when [Abortable.abortTrigger] completes.
+/// property for more info.
+class AbortedRequest extends ClientException {
```
kevmoo, line 41: Has there been discussion on naming here?
Generally we'd follow `AbortedRequestException` or similar naming, right?

See https://api.dart.dev/dart-core/Exception-class.html

kevmoo: `AbortedRequest` sounds like a "flavor" or `Request`, which this certainly is NOT.

brianquinlan: There is not been a discussion on naming. +1 to changing the exception name.

brianquinlan:
I added a (currently failing) test that verifies the behavior when the response stream is not being listened to when the response ends:
https://github.com/dart-lang/http/pull/1773/commits/3d5caa79b1c156df8a1d3c01b3553f12c0fbc074

brianquinlan:
I don't have any more time to work on this today (and I'm not sure about tomorrow). But I think that we have an outline of a solution for the `detachSocket` flakes.

pkgs/http/lib/src/abortable.dart
JaffaKetchup: I can see the potential confusion that it looks like an object that holds information about an aborted request. But the reason I was against calling it an exception (and against extending from client exception initially), is because it doesn't meet the docs on Exception:

> ...information ... about a failure

See also TickerCanceled: https://api.flutter.dev/flutter/scheduler/TickerCanceled-class.html

I'm not against changing the name to something like `RequestAborted` - but I wouldn't necessarily want to include "exception".

brianquinlan:
@JaffaKetchup I'm on holiday until June 23rd.

If you need approval to run the github workflow, maybe ping @natebosch here?

brianquinlan:
I'm back from holiday. IIRC, the last flake that I was was with `detach`. I added a fix to `io_client.dart` that removes the flake but introduces another bug.

@JaffaKetchup do you have time now to continue with this?

JaffaKetchup:
Hey, yes I can continue, but getting to the point where I'm not sure exactly what's causing issues. Happy to take a look though, let me know what's changed :) Got a fairly large amount of time now so can do a little digging.

brianquinlan:
> Hey, yes I can continue, but getting to the point where I'm not sure exactly what's causing issues. Happy to take a look though, let me know what's changed :) Got a fairly large amount of time now so can do a little digging.

I think that looking at the note here and related subscription handling might be a good place to start:
https://github.com/dart-lang/http/pull/1773/files#diff-854200f24ea7a2603c70b046c80763c8d0b103bb29c93f7ef9eb36282261c5f9R150

You might need to add some additional tests to verify more of the behavior.

JaffaKetchup:
@brianquinlan I can't quite wrap my head around what's not quite right here. Does `response` start producing events without a listener? Surely if it requires a listener before it produces events then everything is OK? The implementation looks the same as https://pub.dev/documentation/async/latest/async/StreamExtensions/listenAndBuffer.html, which claims it buffers events.

Although I'm not sure about the order of things here - we're assigning a stream controller to `responseController`, then assigning again a little later.

brianquinlan:
> @brianquinlan I can't quite wrap my head around what's not quite right here. Does `response` start producing events without a listener? Surely if it requires a listener before it produces events then everything is OK? The implementation looks the same as https://pub.dev/documentation/async/latest/async/StreamExtensions/listenAndBuffer.html, which claims it buffers events.

IIRC, your original code was unconditionally listening calling `response.listen`. The problem is that `detachSocket` cannot be called after `response.listen` because the socket will be consumed by the listener.

That's why I added `responseController` - so we could only listen to the response when needed. That fixes the `detachSocket` flake but I thought that it would introduce a problem with error responses. Now I don't think so but I'm not sure. I intended by `StreamController` to be a proof-of-concept that will need to be cleaned up.

JaffaKetchup:
@brianquinlan OK, I've fixed the test that was failing on my system 67132f6. Let me know if that fixes what you meant.

I've also renamed the exception to `RequestAborted`. See https://github.com/dart-lang/http/pull/1773#discussion_r2128254542 for why I think using "exception" in the name isn't a good idea.

JaffaKetchup:
@brianquinlan Not sure what the last tests are failing for on iOS (❌ defaultSessionConfiguration profile disabled abort while streaming response (failed)) but I think this is pretty much ready for a re-review.

brianquinlan:
I'll take another look this afternoon and maybe add some additional tests.

pkgs/http/lib/src/io_client.dart
```diff
@@ -123,7 +120,76 @@ class IOClient extends BaseClient {
...
+              if (!responseController.isClosed) {
+                responseController
+                  ..addError(RequestAborted(request.url))
+                  ..close();
+              }
+              ioResponseSubscription?.cancel();
+            });
+          }
+
+          ioResponseSubscription = response.listen(
```
brianquinlan, line 173: Do you think that it would be worth the reduction in complexity to write this as:

```
ioResponseSubscription = response.listen(...
if (isAborted) {
  responseController
  ..addError(RequestAborted(request.url))
  ..close();
  ioResponseSubscription.cancel();
}

**Review by brianquinlan — COMMENTED** [review]
I fixed all of the `cronet_http` fails and most of `cupertino_http` failures. I think that the `ok_http` failures are unrelated (and the `cupertino_http` failures are probably unrelated too). I'll take a closer look at the `cupertino_http` failures tomorrow.

**Comment by brianquinlan** [comment]
I think that the `package:cupertino_http` tests will be fixed in https://github.com/dart-lang/http/pull/1786 and the `package:cronet_http` test failures are unrelated.

File: pkgs/http/lib/src/io_client.dart (hunk 10 above)
Comment (JaffaKetchup) [reply pkgs/http/lib/src/io_client.dart:174]: I don't necessarily see how this would make anything easier. Surely anything not in the `onListen` callback would require another asynchronous trigger/completer, which is more complex. Still need to wait for a listener & the abort trigger.

I think having the separation of concerns with the two trigger listeners makes sense to me.

File: pkgs/http_client_conformance_tests/lib/src/abort_tests.dart [hunk 11]
```diff
@@ -0,0 +1,256 @@
...
+      expect(
+        response.stream.listen(
+          (data) {
+            i += data.length;
+            if (i >= 1000 && !abortTrigger.isCompleted) abortTrigger.complete();
+          },
+        ).asFuture<void>(),
+        throwsA(isA<RequestAborted>()),
+      );
+      expect(i, lessThan(48890));
```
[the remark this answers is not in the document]
JaffaKetchup: Fixed in a9255fd

JaffaKetchup:
@brianquinlan Not sure where/how to test the retry client effectively. Would just plugging it into the existing client conformance tests for io & html be acceptable? I'm not really sure why it is tested alone in the first place.

```diff
  testAbort(clientFactory(),
      supportsAbort: supportsAbort,
      canStreamRequestBody: canStreamRequestBody,
      canStreamResponseBody: canStreamResponseBody);
+ testAbort(RetryClient(clientFactory()),
+     supportsAbort: supportsAbort,
+     canStreamRequestBody: canStreamRequestBody,
+     canStreamResponseBody: canStreamResponseBody);
```

I think that it makes sense use the conformance tests on `RetryClient` but that test should be in `pkg/http`. May have `test/http_retry_conformance_test.dart`?

But that test would not check that the behavior if the request is aborted after the first request fails. Maybe you needs some tests in `http_retry_test.dart` like:

```dart
  test('', () async {
    var count = 0;
    final client = RetryClient(
        MockClient(expectAsync1((request) async {
          count++;
          if (count < 2) return Response('', 503);
          throw RequestAbortedException(...);
        }, count: 2)),
        delay: (_) => Duration.zero);

    expect(client.get(Uri.http('example.org', ''), throwsA<RequestAbortedException>(...));
    
  });
```

You should probably also add a test for the case where the `Client` does not support being aborted. Do you want to not retry is the request has been imported even if the `Client` does not throw `RequestAborted`? Like:

```dart
  test('', () async {
    var count = 0;
    final client = RetryClient(
        MockClient(expectAsync1((request) async {
          count++;
          if (count == 2) {
             abortCompleter.complete();
          }
          return Response('', 503);
        }, count: 2)),
        delay: (_) => Duration.zero);

    final response = await client.get(Uri.http('example.org', ''));
    expect(client.send(<abortable request with abortCompleter>), throwsA<RequestAbortedException>(...));
  });
```

JaffaKetchup:
@brianquinlan That should be all sorted, ready for re-review.

brianquinlan:
Thanks for fixing the lints!

pkgs/http/lib/src/browser_client.dart
```diff
@@ -49,15 +51,14 @@ external JSPromise<Response> _fetch(
...
-  final _abortController = AbortController();
-
   /// Whether to send credentials such as cookies or authorization headers for
   /// cross-site requests.
   ///
   /// Defaults to `false`.
   bool withCredentials = false;
 
   bool _isClosed = false;
+  final _openRequestAbortControllers = <AbortController>[];
```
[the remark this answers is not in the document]
kevmoo: I really don't like this style. Prefer as-is!

Devon: No offence to your bad ideas. 😛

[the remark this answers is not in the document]
brianquinlan: I agree with @kevmoo here - since the type is clearly indicated on the RHS, adding it explicitly is redundant

brianquinlan (approved):

brianquinlan:
I'm going to merge this. My intent is to release a `1.5.0-beta` version and update the associated issues asking for testing.

JaffaKetchup:
Sounds good to me. Thanks for all the help!

brianquinlan:
> Sounds good to me. Thanks for all the help!

@JaffaKetchup No, thank you for all the great work!

JaffaKetchup:
(Just to note, this is now used by default in the latest version of flutter_map, so hopefully it all goes well :D)

brianquinlan:
> (Just to note, this is now used by default in the latest version of flutter_map, so hopefully it all goes well :D)

Awesome! I was wondering if we would get adequate testing for the beta release but I don't need to worry anymore ;-)

JaffaKetchup:
Hey @brianquinlan, just a bit of feedback from our end - we haven't received any feedback at all in regards to this :D. IMO this should be all good for a more general release.

We're taking that as a good thing, since for users this maybe just appears as a slight speed up in map loading if they notice anything at all. If there were any major issues, we'd expect app devs to relay those to us, but we haven't had anything.

If pub.dev and a bit of guess-work is to be believed, ~16% of package-users (likely to increase to 40%+ fairly quickly, as 17% are on v8.2, while 41% are on v8.x) are likely using this functionality (most FM users will be using this unless they're using a plugin). We don't know what that translates to in actual users ofc, but we think quite a few at least based on apps we know about.

Issue #1438, opened 2024-12-26
Avoid adding `charset` to `Content-Type` header


samrat:
Hello! 👋 

I'm trying to use the [OpenAI Realtime API](https://platform.openai.com/docs/guides/realtime-webrtc#:~:text=%22Content%2DType%22%3A%20%22application/sdp%22) and need to send `Content-Type: application/sdp` as one of the headers.

Here's what my Dart code looks like

```
      final response = await http.post(
        Uri.parse('$baseUrl?model=$model'),
        headers: {
          'Authorization': 'Bearer $ephemeralKey',
          'Content-Type': 'application/sdp',
        },
        body: sdp,
        encoding: null
      );
```

Even though I'm setting the header to `application/sdp` the request sent has the header set to `application/sdp; charset=utf-8`(I tested this with iOS, macOS and Web debug targets)

And looks like the OpenAI endpoint is very strict about the value it accepts for this header:

```
Response body: {"error":{"message":"Unsupported content type. This API method only accepts 'application/sdp' requests, but you specified the header 'Content-Type: application/sdp; charset=utf-8'. Please try again with a supported content type.","type":"invalid_request_error","param":null,"code":"unsupported_content_type"}}
```

Is there any way to prevent the `http` library from modifying headers before the request is made?

seifibrahim32:
@brianquinlan I will be looking into this issue.

brianquinlan:
The [documentation for `encoding`](https://pub.dev/documentation/http/latest/http/Request/encoding.html) says that setting it will add the charset parameter.

seifibrahim32:
@brianquinlan Is it obligatory from http to add charset parameter into the headers?

brianquinlan:
"charset" is not required, see https://datatracker.ietf.org/doc/html/rfc2046#section-4.1.2

@samrat I think that your problem is that you are setting the HTTP payload using `body` and `package:http` is assuming a default encoding when converting it into bytes. Instead, you can set `bodyBytes` and use whatever encoding that you want. See https://pub.dev/documentation/http/latest/http/Request/body.html

@seifibrahim32 Changing `package:http`'s charset behavior would be breaking. We may just need better documentation for the top-level functions e.g. https://pub.dev/documentation/http/latest/http/post.html

seifibrahim32:
@brianquinlan Yes, charset should be with default utf8 or utf16 atmost.

@samrat Have you tried to use `encoding` property to maneuver this request?

[...]

PR #1798, opened 2025-07-24
Only set the charset when the body is set for certain mime types

brianquinlan:
Fixes https://github.com/dart-lang/http/issues/1438

NOTE: The `charset` parameter is not defined for [`application/x-www-form-urlencoded`](https://www.iana.org/assignments/media-types/application/x-www-form-urlencoded) so the changes to those tests should not break anything.

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/request.dart
+++ b/pkgs/http/lib/src/request.dart
@@ -61,8 +61,9 @@ class Request extends BaseRequest {
     _checkFinalized();
     _defaultEncoding = value;
     var contentType = _contentType;
-    if (contentType == null || !contentType.parameters.containsKey('charset'))
+    if (contentType == null || !contentType.parameters.containsKey('charset')) {
       return;
+    }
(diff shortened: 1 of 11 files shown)


pkgs/http/lib/src/request.dart
```diff
@@ -14,6 +14,22 @@ import 'utils.dart';
 
 /// An HTTP request where the entire request body is known in advance.
 class Request extends BaseRequest {
+  /// Whether the given MIME type should have a 'charset' parameter.
+  bool _shouldHaveCharset(MediaType? contentType) =>
```
mraleph, line 18: Should this be a `static` (it does not actually use any `Request` state?) or maybe even global function - rather than being nested in `Request`?

mraleph (approved):
LGTM!

pkgs/http/lib/src/request.dart
brianquinlan: I made it `static`.

Issue #1820, opened 2025-09-02
Fetch client: Cancelling response subscriptions waits for next chunk


simolus3:
Consider an HTTP server that sends a line every 10 seconds:

```dart
import 'dart:convert';
import 'dart:io';

void main() async {
  final server = (await HttpServer.bind('localhost', 0))
    ..listen((request) async {
      await request.drain<void>();
      request.response.headers.set('Access-Control-Allow-Origin', '*');
      request.response.headers.set('Content-Type', 'text/plain');

      request.response.bufferOutput = false;
      request.response.add(utf8.encode('0\n'));
      await request.response.addStream(
        Stream.periodic(const Duration(seconds: 10), (i) => '${i + 1}\n')
            .transform(utf8.encoder),
      );
    });

  print('listening on http://localhost:${server.port}/');
}
```

Now, let's say we have a client that inspects the first line and, after a second of deliberation, decides it's done with the response:

```dart
import 'dart:async';
import 'dart:convert';

import 'package:http/http.dart';

void main() async {
  final client = Client();
  final response =
      await client.send(Request('GET', Uri.parse('http://localhost:<port>')));
  print('has response');
  late StreamSubscription<String> sub;
  sub = response.stream
      .transform(utf8.decoder)
      .transform(const LineSplitter())
      .listen((line) async {
    print('has line: $line');
    await Future<void>.delayed(const Duration(seconds: 1));
    print('cancelling subscription');
    final sw = Stopwatch()..start();
    await sub.cancel();
    sw.stop();
    print('subscription cancelled: ${sw.elapsed}');
  });
}

```

With the `dart:io` implementation, cancelling the stream subscription is fast:
```
has response
listening
has line: 0
cancelling subscription
subscription cancelled: 0:00:00.003603
```

With the fetch client however, cancelling the stream after a line waits for the next one:

```
has response
listening
has line: 0
cancelling subscription
subscription cancelled: 0:00:08.981000
```

The issue is that, because the client is implemented with `async*`, it can't react to cancellation requests while in this line: https://github.com/dart-lang/http/blob/ef05b3744424885d93f88a6a50664fb5b7d5cbdb/pkgs/http/lib/src/browser_client.dart#L172

I think the only solution is to rewrite that logic with stream controllers, and I'm happy to contribute a fix (unless I'm doing something wrong or there's a better solution). I know that I could use `Abortable` for that, but I think we should also be able to cancel responses like any other stream.

brianquinlan:
[...]

PR #1822, opened 2025-09-19
Fix `fetch` response stream cancellation

simolus3:
The `fetch` client turns a `ReadableStreamDefaultReader` into a Dart `Stream` by repeatedly invoking `reader.read()` in an `async*` function.

This generally works well, but a problem with `async*` functions is that they can only respect cancellations when yielding. This results in cancellations taking longer than they need to in the following case:

1. A chunk of data is reported from the reader.
2. This is emitted by the stream, and we go into the next iteration of the `_readBody` loop.
3. Asynchronously, the listener calls `.cancel()` on its subscription.
4. This now has to wait for the next chunk of the source stream instead of aborting the request.

Since we already have an `AbortController` set up, this fixes the issue by adding an `onCancel` callback to abort the `fetch` request when the stream subscription is cancelled.

Closes https://github.com/dart-lang/http/issues/1820.

cc @brianquinlan 

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -239,8 +239,7 @@ extension<T> on Stream<T> {
   Stream<T> doOnCancel(void Function() onCancel) => Stream.multi((listener) {
         // Note: We can't use listener.addStream because cancelling the listener
         // in that state would first await cancelling the upstream subscription,
-        // which is what we're trying to avoid here because we need to cancel
-        // the controller as soon as we receive the cancellation request.
+        // but we want to invoke onCancel first.
         var subscription = listen(listener.addSync,
             onError: listener.addErrorSync, onDone: listener.closeSync);
         listener
--- a/pkgs/http/lib/src/browser_client.dart
+++ b/pkgs/http/lib/src/browser_client.dart
@@ -115,8 +115,15 @@ class BrowserClient extends BaseClient {
         headers[header.toLowerCase()] = value;
       }.toJS);
 
+      var subscriptionWasCancelled = false;
+      void cancelSubscription() {
+        subscriptionWasCancelled = true;
+        abortController.abort();
+      }
+
       return StreamedResponseV2(
-        _readBody(request, response),
+        _readBody(request, response, () => subscriptionWasCancelled)
+            .doOnCancel(cancelSubscription),
         response.status,
         headers: headers,
         request: request,
@@ -144,9 +151,9 @@ class BrowserClient extends BaseClient {
   }
 }
 
-Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
+Object _toClientException(Object e, BaseRequest request) {
   if (e case DOMException(:final name) when name == 'AbortError') {
-    Error.throwWithStackTrace(RequestAbortedException(request.url), st);
+    return RequestAbortedException(request.url);
   }
   if (e is! ClientException) {
     var message = e.toString();
@@ -155,10 +162,27 @@ Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
     }
     e = ClientException(message, request.url);
   }
-  Error.throwWithStackTrace(e, st);
+  return e;
 }
 
-Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
+Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
+  Error.throwWithStackTrace(_toClientException(e, request), st);
+}
+
+Stream<List<int>> _readBody(BaseRequest request, Response response,
+    bool Function() subscriptionWasCancelled) async* {
+  void rethrowAsClientException(Object e, StackTrace st) {
+    final clientException = _toClientException(e, request);
+    if (clientException is RequestAbortedException &&
+        subscriptionWasCancelled()) {
+      // The .read() call was aborted in response to the stream subscription
+      // being cancelled (as opposed to an abortTrigger completing). The abort
+      // error is expected in this case and not an error.
+    } else {
+      Error.throwWithStackTrace(clientException, st);
+    }
+  }
+
   final bodyStreamReader =
       response.body?.getReader() as ReadableStreamDefaultReader?;
 
@@ -178,7 +202,7 @@ Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
     }
   } catch (e, st) {
     isError = true;
-    _rethrowAsClientException(e, st, request);
+    rethrowAsClientException(e, st);
   } finally {
     if (!isDone) {
       try {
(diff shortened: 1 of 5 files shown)


pkgs/http/lib/src/browser_client.dart
```diff
@@ -155,10 +162,27 @@ Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
...
+}
+
+Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
+  Error.throwWithStackTrace(_toClientException(e, request), st);
 }
 
-Stream<List<int>> _readBody(BaseRequest request, Response response) async* {
+Stream<List<int>> _readBody(BaseRequest request, Response response,
+    bool Function() subscriptionWasCancelled) async* {
+  void rethrowAsClientException(Object e, StackTrace st) {
```
brianquinlan, line 174: This could just be a matter of taste, but I find the approach of using request abort to handle stream cancellation problematic. Could I suggest that you try a different approach so that we can compare? That approach would be to create a `StreamController` in this method and then do chained calls to `read`. Pseudocode:

```dart
var Completer? readDoneOrCancelled;
final s = StreamController(
  ...
  onCancel: () { isDone = true; readDoneOrCancelled?.complete() };
);
while (!isDone) {
    readDoneOrCancelled = Completer();
    final chunk = bodyStreamReader.read().toDart.then((chunk) {
      readDoneOrCancelled.complete();
      if (chunk.done) {
        isDone = true;
      }
      s.add(chunk.value! as JSUint8Array).toDart);
    });
  await readDoneOrCancelled.future;
}

```

brianquinlan (commented):
Very high quality PR! Especially the added conformance test (a few of which seem to be failing).

pkgs/http/lib/src/browser_client.dart
simolus3: I like that approach :+1: We also need to handle the stream pausing but that doesn't make this much more complicated either.

simolus3:
It looks like the remaining failures are all infra or appear to be unrelated to my changes. Let me know if there's anything I should look into.

brianquinlan:
> It looks like the remaining failures are all infra or appear to be unrelated to my changes. Let me know if there's anything I should look into.

Yep, these have nothing to do with your change.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -155,50 +155,105 @@ Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
...
-        // If we have already encountered an error swallow the
-        // error from cancel and simply let the original error to be
-        // rethrown.
-        if (!isError) {
-          _rethrowAsClientException(e, st, request);
-        }
+
+      controller.add((chunk.value! as JSUint8Array).toDart);
+      if (controller.isPaused) {
+        final resume = waitingForResume = Completer();
```
lrhn, line 196: (Good choice to only allocate completer if actually needed, and not for each pause/resume!)

pkgs/http/lib/src/browser_client.dart
```diff
@@ -155,50 +155,105 @@ Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
...
+    }
+    ..onResume = markResumed
+    ..onCancel = () async {
+      // Ensure the read loop isn't blocked due to a paused listener.
+      markResumed();
+      cancelCompleter.complete(null);
+      await pipeIntoController;
+    };
+
+  return controller.stream;
```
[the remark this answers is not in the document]
simolus3: Both MDN and (from a quick look) also the streams whatwg standard are super unclear on whether we need to call `.cancel()` on the underlying `ReadableStreamDefaultReader` if the reader emits a done event. Intuitively I'd say it's not necessary, but the old implementation did it and I'm not sure if this may be a leak if we don't do it.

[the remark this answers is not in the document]
simolus3: Hm no, judging by step 2 of [this spec](https://streams.spec.whatwg.org/#readable-stream-cancel), it appears to be unecessary. That information is hard to find though, I'll add a comment somewhere.

simolus3:
Alright, I have adopted most of the suggestions shared by @lrhn. I think they make the control flow a lot more obvious 👍 

I think the correct behavior (which we also had with the old `async*` implementation) is to only call `getReader()` when the stream is actually listened to, so I've moved most of the setup into the `onListen` callback. That also avoids the use of finalizers here.

Despite being a single-use stream, I think using `Stream.multi` may be helpful here to be explicit about adding events synchronously? I've also added a comment on each `addSync` / `addErrorSync` / `closeSync` invocation to explain why it's safe.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -155,49 +156,98 @@ Never _rethrowAsClientException(Object e, StackTrace st, BaseRequest request) {
...
+        // We only cancel the reader when the subscription is cancelled - we
+        // don't need to do that for normal done events because the stream is in
+        // a completed state at that point.
+        await reader.cancel().toDart;
+      } catch (e, s) {
+        if (!hadError) {
+          // If the stream enters an error state, .cancel() will rethrow that
+          // error. We want to ignore that since we would have added it to the
+          // stream already.
+          _rethrowAsClientException(e, s, request);
```
brianquinlan, line 203: This will appear as an asynchronous exception, right? I really hate those.

Also, is it possible for it to throw. The [mdn](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/cancel) says:

> Exceptions
> TypeError
> The source object is not a ReadableStreamDefaultReader, or the stream has no owner.

If you change this, you should also change the comment on line 217.

pkgs/http_client_conformance_tests/lib/src/response_body_streamed_test.dart
```diff
@@ -77,5 +77,63 @@ void testResponseBodyStreamed(Client client,
       });
       await cancelled.future;
     });
+
```
brianquinlan, line 80: [optional] Would you mind adding some tests to check that stream pause and resume works?

brianquinlan (changes requested):

pkgs/http_client_conformance_tests/lib/src/response_body_streamed_test.dart
simolus3: I have added a few tests ensuring pause and resume work. Let me know if you were looking for more. To properly test this, one would probably have to check that pausing a stream on the client eventually also pauses the stream on the server due to flow control. But that seems like a very tricky test to write.

brianquinlan:
It looks like one test is failing on WASM.

simolus3:
Thanks, I've fixed that test.

brianquinlan:
It looks like packages `cupertino_http` and `cronet_http` both fail on "cancel paused stream via abortable request". Let me see if I can figure out why...

brianquinlan:
@simolus3 Thanks for sticking with this and for the great work!

brianquinlan:
The remaining CI failures are unrelated.

brianquinlan (approved):

simolus3:
Thanks for fixing the test and reviewing this!

PR #1831, opened 2025-10-08
feat(ok_http): add user-agent

denysvitali:

---

This adds an `userAgent` configuration field as with other HTTP libraries. This allows more flexibility in migrating from dart:io's HTTP to OkHttp.

### DIFF
--- a/pkgs/ok_http/lib/src/ok_http_client.dart
+++ b/pkgs/ok_http/lib/src/ok_http_client.dart
@@ -91,6 +91,8 @@ class OkHttpClientConfiguration {
   /// This should only be used in a testing environment.
   final bool validateServerCertificates;
 
+  final String? userAgent;
+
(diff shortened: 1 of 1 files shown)


pkgs/ok_http/lib/src/ok_http_client.dart
```diff
@@ -91,6 +91,8 @@ class OkHttpClientConfiguration {
   /// This should only be used in a testing environment.
   final bool validateServerCertificates;
 
+  final String? userAgent;
```
brianquinlan: Could you add dart doc for this?

brianquinlan (changes requested):
Could you add a test here: https://github.com/dart-lang/http/blob/master/pkgs/ok_http/example/integration_test/client_configuration_test.dart

Issue #1842, opened 2025-10-31
[cronet_http] Native JNI crashes in globalEnv_CallObjectMethod/CallVoidMethod with native_dio_adapter


androidanit:
After migrating from the default `IOClient` to `cronet_http` via `native_dio_adapter`, the app is experiencing native JNI crashes reported on Google Play Console. The crashes occur intermittently and are affecting production users, though they are not persistent (users experience them occasionally, not every launch).

The app was previously stable with default `IOClient` and no JNI-related crashes were observed.

### Crash Statistics

Two primary crash patterns account for 78% of all crashes:

**1. Crash Pattern 1 (44% of crashes):**
```
[split_config.arm64_v8a.apk!libdartjni.so] global_jni_env.c - globalEnv_CallObjectMethod
```

**2. Crash Pattern 2 (34% of crashes):**
```
[split_config.arm64_v8a.apk!libdartjni.so] global_jni_env.c - globalEnv_CallVoidMethod
```

### Environment

- **Flutter:** 3.27.1
- **Dart:** 3.6.0
- **cronet_http:** ^1.6.0
- **native_dio_adapter:** ^1.5.0
- **dio:** ^5.9.0
- **Platform:** Android (arm64_v8a)
- **Target Devices:** Production devices with Google Play Services

### Dependencies

```yaml
dependencies:
  flutter:
    sdk: flutter
  dio: ^5.9.0
  cronet_http: ^1.6.0
  native_dio_adapter: ^1.5.0
  dio_smart_retry: ^7.0.1
  # ... other dependencies
```

### Implementation Code

```dart
// Configure Dio instance
Dio dio = Dio();

// Using NON default adapter - iOS specific addition
dio.httpClientAdapter = NativeAdapter(
    createCupertinoConfiguration: () => URLSessionConfiguration.ephemeralSessionConfiguration()
      ..cache = null // Completely disable caching
      ..requestCachePolicy = NSURLRequestCachePolicy.NSURLRequestReloadIgnoringCacheData);
```

#### Crash Pattern 1 - globalEnv_CallObjectMethod (44%)

<details>
<summary>Click to expand full backtrace</summary>

```
#00  pc 0x000000000031562c  /apex/com.android.art/lib64/libart.so (art::JValue art::InvokeVirtualOrInterfaceWithVarArgs<_jmethodID*>(art::ScopedObjectAccessAlreadyRunnable const&, _jobject*, _jmethodID*, std::__va_list)+1640)
[...]

PR #1845, opened 2025-11-07
[cronet_http] Fix: multiple Flutter engines would cause crash

### LINKED ISSUE #1217: Cronet exception: m.ms: Exception in CronetUrlRequest: net::ERR_CONNECTION_CLOSED, ErrorCode=5
### Package

cronet_http

### Version

1.2.1

### Operating-System

Android

### Output of `flutter doctor -v`

```shell
flutter doctor -v
[✓] Flutter (Channel stable, 3.19.6, on macOS 14.4.1 23E224 darwin-arm64, locale en-RU)
    • Flutter version 3.19.6 on channel stable at /Users/dmitry/fvm/versions/3.19.6
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 54e66469a9 (6 weeks ago), 2024-04-17 13:08:03 -0700
    • Engine revision c4cd48e186
    • Dart version 3.3.4
    • DevTools version 2.31.1

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/dmitry/Documents/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/dmitry/Documents/sdk
    • Java binary at: /Users/dmitry/Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.10+0-17.0.10b1087.21-11572160)
    • All Android licenses accepted.

[!] Xcode - develop for iOS and macOS (Xcode 15.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15E204a
    ! CocoaPods 1.12.1 out of date (1.13.0 is recommended).
        CocoaPods is used to retrieve the iOS and macOS platform side's plugin code that responds to your plugin usage on the Dart side.
        Without CocoaPods, plugins will not work on iOS or macOS.
        For more info, see https://flutter.dev/platform-plugins
      To upgrade see https://guides.cocoapods.org/using/getting-started.html#updating-cocoapods for instructions.

[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome --ignore-certificate-errors
    --disable-web-security --user-data-dir=/Users/dmitry/temp)
    ! /Applications/Google Chrome.app/Contents/MacOS/Google Chrome --ignore-certificate-errors --disable-web-security --user-data-dir=/Users/dmitry/temp is not
      executable.

[✓] Android Studio (version 2023.3)
    • Android Studio at /Users/dmitry/Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.10+0-17.0.10b1087.21-11572160)

[✓] IntelliJ IDEA Community Edition (version 2023.3.2)
    • IntelliJ at /Users/dmitry/Applications/IntelliJ IDEA Community Edition.app
    • Flutter plugin version 78.4.2
    • Dart plugin version 233.15123

[✓] VS Code (version 1.87.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.88.0

[✓] Connected device (2 available)
    • iPhone SE (3rd generation) (mobile) • *** • ios          • com.apple.CoreSimulator.SimRuntime.iOS-17-4 (simulator)
    • macOS (desktop)                     • macos                                • darwin-arm64 • macOS 14.4.1 23E224 darwin-arm64

[✓] Network resources
    • All expected network resources are available.

! Doctor found issues in 2 categories.
```

### Dart Version

3.3.4

### Steps to Reproduce

1. Create long polling client (that will make requests one after another for a period of time)
2. Start making requests

### Expected Result

Requests succeed

### Actual Result

After some time requests will fail with `Cronet exception: m.ms: Exception in CronetUrlRequest: net::ERR_CONNECTION_CLOSED, ErrorCode=5, InternalErrorCode=-100, Retryable=true, uri=https://***/v1/polling.getUpdates?ts=***,statusCode: null,cause: DioException [unknown]: null`

Devices distribution (out of 30k events):
image

Full log
```
Cronet exception: m.ms: Exception in CronetUrlRequest: net::ERR_CONNECTION_CLOSED, ErrorCode=5, InternalErrorCode=-100, Retryable=true, uri=https://alive.go-rhino.com/v1/polling.getUpdates?ts=1717011232038) | *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 27540, tid: 27615, name 1.ui
os: android arch: arm64 comp: yes sim: no
build_id: 'b866ace18fa006c0f7dae5b7963c0614'
isolate_dso_base: 797e36f000, vm_dso_base: 797e36f000
isolate_instructions: 797e5b5580, vm_instructions: 797e59f000
    #00 abs 000000797ead0d0b virt 0000000000761d0b _kDartIsolateSnapshotInstructions+0x51b78b
    #01 abs 000000797ead0073 virt 0000000000761073 _kDartIsolateSnapshotInstructions+0x51aaf3
    #02 abs 000000797e71adcb virt 00000000003abdcb _kDartIsolateSnapshotInstructions+0x16584b
    #03 abs 000000797eb2fcfb virt 00000000007c0cfb _kDartIsolateSnapshotInstructions+0x57a77b
    #04 abs 000000797eb2fafb virt 00000000007c0afb _kDartIsolateSnapshotInstructions+0x57a57b
<asynchronous suspension>
```
image

### LINKED ISSUE #1720: [cronet_http] Potential problem when having multiple Flutter engines
There is an issue with the context getter of package:jni that it will crash if used again after creating a second engine: https://github.com/dart-lang/native/issues/1865

This is because the second engine sets it to another context invalid for the first one. Can we check if `cronet_http` is susceptible to this problem? (or maybe we store the context and it won't cause a crash/cronet doesn't really need the context so it doesn't cause a crash/etc.)

If yes, then we would want to add a warning to `cronet_http`'s readme about the fact that it's not safe to use it when using multiple flutter engines. 

The problem can be solved either if we find a different way of getting the global context, which might work well at least for cronet. Or by exposing an "engine ID", so we can keep a map of contexts per Flutter engine instead of a singleton: https://github.com/flutter/flutter/issues/163430

HosseinYousefi:
Upgraded jni and jnigen to 0.15.0. This fixes #1217 and closes #1720.


HosseinYousefi:
@knopp There seems to be another issue where `engineId` is `null` even though `IntegrationTestWidgetsFlutterBinding.ensureInitialized();` is called in the CI. 

It fails when trying to test a post inside `Isolate.run`: 

https://github.com/dart-lang/http/blob/c8355d1fbf26f53eb45472252514ae1233b5ff00/pkgs/http_client_conformance_tests/lib/src/isolate_test.dart#L16-L17

Do you have any idea why that might be?

--

Update: I guess the `engineId` is attached to the main isolate? I wonder if we could attach it to the isolate group instead, this way it would be non-null from all of the isolates

HosseinYousefi:
@brianquinlan If I change the conformance test so that it creates the client itself in the main isolate and just does the POST request in another isolate, then it would work:

```diff
Future<void> _testPost(Client Function() clientFactory, String host) async {
+ final client = clientFactory();
  await Isolate.run(
      () => 
-         clientFactory()
+         client
           .post(Uri.http(host, ''), body: 'Hello World!'));
}
```

But I'm not sure if you want this behavior for cronet_http.

knopp:
Can we intercept background isolate creation somehow in the engine to set the engine id? Right now it is only set for the main isolate.

brianquinlan:
> @brianquinlan If I change the conformance test so that it creates the client itself in the main isolate and just does the POST request in another isolate, then it would work:
> 
> ```diff
> Future<void> _testPost(Client Function() clientFactory, String host) async {
> + final client = clientFactory();
>   await Isolate.run(
>       () => 
> -         clientFactory()
> +         client
>            .post(Uri.http(host, ''), body: 'Hello World!'));
> }
> ```
> 
> But I'm not sure if you want this behavior for cronet_http.

At the time that I wrote this test, native pointers could not be shared between isolates. Now that they can be, I think that we should test both scenarios, i.e.:
1. create the `Client` in one isolate, use it in another
2. create the `Client` in the non-main isolate and use it there

Performing HTTP requests in a background Isolate is definitely something that people do.

HosseinYousefi:
> Can we intercept background isolate creation somehow in the engine to set the engine id? Right now it is only set for the main isolate.

If we keep a map from isolate group ids to the engine ids, then we could fetch the right engine id using the isolate group id. 

Moreover, what if we just have engine id be exactly equal to the isolate group id? I assume that each engine is attached to a single isolate group anyway so this reduces the number of maps we are keeping. This way there is no need to set any engine id, instead we can just call `Dart_CurrentIsolateGroup` from any isolate and get it. 

cc @mkustermann

mkustermann:
Reading through the code, it seems the implementation does roughtly this:

* `package:jni` is a real flutter plugin pub package. It's Java code will be invoked (by flutter engine) when the plugin is loaded. That Java code will call into `package:jni` C code to set a bunch of things, including cache some java context objects (e.g. application context) via jni global refs.
* ??? If using multiple flutter engines, it may do the above multiple times with different context objects
* The bindings of cronet will use `package:jni` to get application contexts which it needs due to cronet java API needing it (and it gets the context from C code of `package:jni`)

Now looking through the cronet API it [says](https://developer.android.com/develop/connectivity/cronet/reference/org/chromium/net/CronetEngine.Builder.html#Builder(android.content.Context)):
"""
`context` Android `Context` to retrieve the application context. A reference to only the application context will be kept, so as to avoid extending the lifetime of context unnecessarily.
"""
And googling around it seemingly suggests that "android application context" has same lifetime as the OS process itself. If this is the case, then I don't see an issue caching the "android application context" in `package:jni` C code and using across isolates or isolate groups. If this isn't the case (it somehow sounds  like it, @HosseinYousefi ?) and an OS process can live longer and  multiple "android application context"s can be created inside that process than the current caching of the global ref in C code seems problematic.

Now from a Dart perspective: The OS process can have multiple isolate groups, each of which can have multiple isolates. Flutter may start an isolate group and one particular UI isolate per flutter engine. Any isolate can spawn more isolates at runtime. When the android lifecycle system tells flutter it shuts down the current activity or UI, flutter may stop the UI isolate, but not necessarily background isolates.

So isolates in a group can outlive the UI isolate, so they shouldn't depend on resources only available to the UI isolate (such as an activity context).

So we'd want to avoid sharing a cronet client across isolates in an isolate group if shutting the UI isolate down will make this cronet client defunct/invalid.

> If we keep a map from isolate group ids to the engine ids, then we could fetch the right engine id using the isolate group id.

@HosseinYousefi Could you explain why "engine id" matters here? Isn't this purely a question about lifecycle of android context (and other?) objects and from which isolates/threads those objects with lifecycle can be used?

knopp:
> > Can we intercept background isolate creation somehow in the engine to set the engine id? Right now it is only set for the main isolate.
> 
> If we keep a map from isolate group ids to the engine ids, then we could fetch the right engine id using the isolate group id.
> 
> Moreover, what if we just have engine id be exactly equal to the isolate group id? I assume that each engine is attached to a single isolate group anyway so this reduces the number of maps we are keeping. This way there is no need to set any engine id, instead we can just call `Dart_CurrentIsolateGroup` from any isolate and get it.
> 
> cc @mkustermann

On most platforms the engineId is simply a casted pointer to engine instance.

HosseinYousefi:
> that "android application context" has same lifetime as the OS process itself. If this is the case, then I don't see an issue caching the "android application context" in package:jni C code and using across isolates or isolate groups.

We're getting the application context from `onAttachedToEngine` callback of `package:jni`s FlutterPlugin and it seems that when another `onAttachedToEngine` gets called and therefore overrides the cached "global android application context", we get crashes suggesting that we should be using one context **per flutter engine**.

> @HosseinYousefi Could you explain why "engine id" matters here? Isn't this purely a question about lifecycle of android context (and other?) objects and from which isolates/threads those objects with lifecycle can be used?

Because we store a map from "engine id" to "android application context" to be able to retrieve one context per flutter engine.

https://github.com/dart-lang/native/blob/28b7c5fcfc1492d1821af0cdcb3c277c372110b1/pkgs/jni/android/src/main/java/com/github/dart_lang/jni/JniPlugin.java#L18

> So isolates in a group can outlive the UI isolate, so they shouldn't depend on resources only available to the UI isolate (such as an activity context).

Yes and we should remove the stored context from the map when `onDetachedFromEngine` gets called. 

https://github.com/dart-lang/native/blob/28b7c5fcfc1492d1821af0cdcb3c277c372110b1/pkgs/jni/android/src/main/java/com/github/dart_lang/jni/JniPlugin.java#L41-L46

That's not a problem for fetching the short-lived activity because we explicitly document that it should not be used from the other threads:

https://github.com/dart-lang/native/blob/28b7c5fcfc1492d1821af0cdcb3c277c372110b1/pkgs/jni/lib/src/jni.dart#L228-L287

This is a general `package:jni` problem, regardless of whether cronet specifically is being affected because of it: https://github.com/dart-lang/native/issues/1865

mkustermann:
(Apologies I was accidentally looking at the old code from the archived https://github.com/dart-lang/jnigen)

> We're getting the application context from onAttachedToEngine callback of package:jnis FlutterPlugin and it seems that when another onAttachedToEngine gets called and therefore overrides the cached "global android application context", we get crashes suggesting that we should be using one context per flutter engine.

Well, specifically for the "global android application context" it seems there's only one per process so why not just a static global for it, initialize it once, on any second attachment call check that it's the same reference (which it should if there's only one per process) and never clear it out?

See [docs](https://developer.android.com/reference/android/content/Context#getApplicationContext()) which says "Return the context of the single, global Application object of the current process. "

> we get crashes suggesting that we should be using one context per flutter engine.

What kind of crash?

HosseinYousefi:
> Well, specifically for the "global android application context" it seems there's only one per process so why not just a static global for it, initialize it once, on any second attachment call check that it's the same reference (which it should if there's only one per process) and never clear it out?

Somehow it seems that it's not the case. Maybe the context we get from `onAttachedToEngine` is not the global android application context but something tied to the Flutter engine. I think we need this kind of context as well specially for UI, but maybe then we can afford to limit the context getter to the platform thread as well.

I could try getting the application context via `ActivityThread.currentApplication()` and see if things would be fixed that way for cronet or not. This API is annotated with [`@UnsupportedAppUsage`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/ActivityThread.java#2918) so I'm not sure how stable it will be.

Edit: yeah,`context.getApplicationContext()` should theoretically provide something that doesn't change.

> What kind of crash?

There are two specified in https://github.com/dart-lang/http/issues/1217 and https://github.com/dart-lang/http/issues/1842. For the second one I asked if they still face the same problem if they remove any background tasks via creating new flutter engines and they don't. I should try to reproduce this in a small example myself to make sure I'm solving the right problem here.

mkustermann:
> Somehow it seems that it's not the case. Maybe the context we get from onAttachedToEngine is not the global android application context but something tied to the Flutter engine.

If so, can you get the global one by calling [Context.getApplicationContext()](https://developer.android.com/reference/android/content/Context#getApplicationContext()) on the one you get from flutter? The real application context probably satisfies `context == context.GetApplicationContext()` condition.

HosseinYousefi:
> The real application context probably satisfies `context == context.GetApplicationContext()` condition.

`FlutterPluginBinding` itself actually has a [`getApplicationContext`](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/plugins/FlutterPlugin.FlutterPluginBinding.html#getApplicationContext()) which we're using and the context of that indeed satisfies `context == context.getApplicationContext() == ActivityThread.currentApplication().getApplicationContext()`.

So the problem is something else.

brianquinlan:
Is this change obsolete?

HosseinYousefi:
> Is this change obsolete?

Yes. I will close this PR.

PR #1858, opened 2025-12-09
test(cupertino_http): Retry cancellation tests.

brianquinlan:

<details>

</details>

### DIFF
--- a/pkgs/cupertino_http/example/integration_test/url_session_task_test.dart
+++ b/pkgs/cupertino_http/example/integration_test/url_session_task_test.dart
@@ -239,9 +239,6 @@ void testURLSessionTaskCommon(
       'cancel',
       () {
         task.cancel();
-        while (true) {
-          print(task.toString());
-        }
         if (suspendedAfterCancel) {
           expect(
             task.state,
@@ -256,7 +253,7 @@ void testURLSessionTaskCommon(
         expect(task.response, null);
         task.toString(); // Just verify that there is no crash.
       },
-      // If the task completes before cancelling then the task state will be
+      // After the request cancellation is complete, the task state will be
       // `NSURLSessionTaskStateCompleted`. So run the test a few times to allow
       // the timing to work out (which still isn't a great approach but I can't
       // think of a better way).


pkgs/cupertino_http/example/integration_test/url_session_task_test.dart
```diff
@@ -213,6 +213,9 @@ void testURLSessionTaskCommon(
           await request.drain<void>();
           request.response.headers.set('Content-Type', 'text/plain');
           request.response.write('Hello World');
+          // Add a small delay to allow cancellation to run before the request
+          // is completed.
```
natebosch: Is it feasible to coordinate the sequencing without an arbitrary delay? Should we wait on a `Completer` here and have the test body complete it after the cancellation runs? It will require edits in existing tests, but much better to be explicit about the interleaving.

pkgs/cupertino_http/example/integration_test/url_session_task_test.dart
```diff
@@ -235,22 +235,30 @@ void testURLSessionTaskCommon(
...
+          expect(
+            task.state,
+            NSURLSessionTaskState.NSURLSessionTaskStateCanceling,
+          );
+        }
+        expect(task.response, null);
+        task.toString(); // Just verify that there is no crash.
+      },
+      // After the request cancellation is complete, the task state will be
+      // `NSURLSessionTaskStateCompleted`. So run the test a few times to allow
```
natebosch: Why do we care about testing this specific behavior if it's so hard to observe it in usage? Should we allow the completed case as a successful test?

```dart
        expect(
          task.state,
          anyOf(
            NSURLSessionTaskState.NSURLSessionTaskStateSuspended,
            NSURLSessionTaskState.NSURLSessionTaskStateCompleted,
          ),
        );
```

brianquinlan: > Why do we care about testing this specific behavior if it's so hard to observe it in usage?

At some point my `.toString()` would throw for certain states so I created a test to catch that. But it might not be worth it.

> Should we allow the completed case as a successful test?
> 
> ```dart
>         expect(
>           task.state,
>           anyOf(
>             NSURLSessionTaskState.NSURLSessionTaskStateSuspended,
>             NSURLSessionTaskState.NSURLSessionTaskStateCompleted,
>           ),
>         );
> ```

That would work but it would miss the intent of the test i.e. to verify that a cancelled test is actually cancelled instead of running to completion (and to verify that `toString` works correctly ;-))

natebosch:
I think the retry is an OK solution. If it continues to be flaky we can reconsider dropping it.

PR #1862, opened 2025-12-09
[cronet_http] Fix background bugs

HosseinYousefi:
By making the callbacks all async, we basically make them all fire-and-forget so a long running http request cannot callback to a potentially non-existent Dart callback synchronously and fail. Instead it will just be ignored.

Also release the JNI references eagerly to prevent reference overflow issues that could be created by doing too many requests in one go (and GC not picking up the references in the meantime)

These changes should overall fix the bugs people are facing in https://github.com/dart-lang/http/issues/1217 and https://github.com/dart-lang/http/issues/1842.

### DIFF
--- a/pkgs/cronet_http/lib/src/cronet_client.dart
+++ b/pkgs/cronet_http/lib/src/cronet_client.dart
@@ -19,34 +19,16 @@ class _StreamedResponseWithUrl extends StreamedResponse
   @override
   final Uri url;
 
-  _StreamedResponseWithUrl(super.stream, super.statusCode,
-      {required this.url,
-      super.contentLength,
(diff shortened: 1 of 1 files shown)


pkgs/cronet_http/lib/src/cronet_client.dart
```diff
@@ -28,20 +28,39 @@ class _StreamedResponseWithUrl extends StreamedResponse
       super.reasonPhrase});
 }
 
+class UrlResponseInfo {
```
brianquinlan: I think that this should be private. Also, would it make sense to just add these add fields to `CronetStreamedResponse`?

PR #1874, opened 2026-01-31
feat: add isOk convenience getter to BaseResponse

Turskyi:
This PR adds a convenience getter isOk to the BaseResponse class.

What is changing? I have added a boolean getter isOk to BaseResponse that returns true if the statusCode is exactly 200. By placing it in the base class, this convenience is now available for all response types, including Response and StreamedResponse.
I have also added unit tests to verify this behavior across both response types and various status codes (200, 201, 404, 500).

Why? Checking for a 200 OK response is the most frequent operation when handling HTTP requests. Currently, developers must either:

1. Use magic numbers (e.g., if (response.statusCode == 200)), which is less expressive.
2. Import dart:io to use HttpStatus.ok (e.g., if (response.statusCode == HttpStatus.ok)).

This addition makes the API more intuitive and readable while remaining platform-agnostic:

```
final response = await http.get(url);
if (response.isOk) {
  // Handle success
}
```

•
[x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
<details> <summary>Contribution guidelines:</summary><br>
•
See our contributor guide for general expectations for PRs.
•
Larger or significant changes should be discussed in an issue before creating a PR.
•
Contributions to our repos should follow the Dart style guide and use dart format.
•
Most changes should add an entry to the changelog and may need to rev the pubspec package version.
•
Changes to packages require corresponding tests.

### DIFF
--- a/pkgs/http/lib/src/base_response.dart
+++ b/pkgs/http/lib/src/base_response.dart
@@ -84,6 +84,9 @@ abstract class BaseResponse {
       throw ArgumentError('Invalid content length $contentLength.');
     }
   }
+
+  /// Whether the status code is 200 (OK).
+  bool get isOk => statusCode == 200;
 }
 
 /// A [BaseResponse] with a [url] field.
--- a/pkgs/http/lib/src/response.dart
+++ b/pkgs/http/lib/src/response.dart
@@ -64,9 +64,6 @@ class Response extends BaseResponse {
         persistentConnection: response.persistentConnection,
         reasonPhrase: response.reasonPhrase);
   }
-
-  /// Returns true if the status code is 200 (OK).
-  bool get isOk => statusCode == 200;
 }
 
 /// Returns the encoding to use for a response with the given headers.
--- a/pkgs/http/test/response_test.dart
+++ b/pkgs/http/test/response_test.dart
@@ -184,25 +184,43 @@ void main() {
     });
   });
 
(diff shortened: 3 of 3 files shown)


Reprevise:
Personally, I'm more in favor of something like an `isSuccess` that checks for `>=200 && <400`

Turskyi:
> Personally, I'm more in favor of something like an `isSuccess` that checks for `>=200 && <400`

Thanks for the suggestion - that makes sense as a broader convenience 👍

Right now this PR adds `isOk` specifically for `statusCode == 200`. The intention here was:
	•	to keep the change small and focused,
	•	and to match the most common check people do when they care about exactly 200.

A broader `isSuccess` getter (e.g. `statusCode >= 200 && statusCode < 400`) could definitely be useful, but that’s a slightly different API/design choice. I think it might deserve its own discussion or a follow-up PR if maintainers want that.

Happy to update this PR based on maintainer feedback!

brianquinlan:
I don't think that we can merge this as-is because there could be instances of `implements BaseResponse`, which would be broken by this implementation.

Instead, we could implement this as an extension method.

I'd suggest one check per [response type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status), e.g. `isSuccessful`, and having `isOk` for exactly 200.

WDYT?

PR #1876, opened 2026-02-04
feat(cupertino_http): shared session

### LINKED ISSUE #1002: Sharing CupertinoClient's URLSession
I have a use-case where I want to:
1. Use URLSession directly to run efficient large-file downloadTask's.
2. Make API calls using the simplified interface.

I would like to use the same URLSession for both of these. One way would be to allow a CupertinoClient to be constructed from a URLSession, another to make the internal URLSession public.

mertalev:
This PR enables sharing of a single URLSession with platform and across isolates. The existing implementation's delegate approach tightly couples a session to a single isolate. As a solution, it adds an implementation based on the `bytes(for:)` API, which does not require delegate-based callbacks. This allows users to bring their own session configured with their own delegate. It enables my goal to use a single shared session in my app; this is optimal for multiplexing and caching and means a single place to configure authentication like client certificates.

I've tested these changes in the app and made sure the conformance tests are passing with this implementation. A few changes were needed in the tests because `bytes(for:)` does its own chunking internally before yielding bytes. I have not tested its performance. I'd guess it performs better since the implementation is more contained in Swift with less Dart trampolining/bookkeeping, but it'd be good to profile it.

From a maintenance standpoint, I think this implementation is simpler than the delegate-based approach and might be able to replace it at some point, perhaps when the minimum iOS version is bumped to 15? It's relatively self-contained as it is, though, and should have no effect on existing users of cupertino_http.

Fixes #1002

---

<details>

</details>

### DIFF
--- a/pkgs/cupertino_http/lib/src/cupertino_api.dart
+++ b/pkgs/cupertino_http/lib/src/cupertino_api.dart
@@ -30,6 +30,7 @@ import 'dart:async';
 import 'dart:ffi' as ffi;
 import 'dart:typed_data';
 
+import 'package:http_profile/http_profile.dart';
 import 'package:objective_c/objective_c.dart' as objc;
 
 import 'native_cupertino_bindings.dart' as ncb;
@@ -1348,6 +1349,7 @@ class StreamingTask {
     required URLRequest request,
     required int maxRedirects,
     required Object Function(objc.NSError error, URLRequest request) mapError,
+    HttpClientRequestProfile? profile,
   }) {
     final completer = Completer<URLResponse>();
     late final StreamingTask task;
@@ -1405,6 +1407,16 @@ class StreamingTask {
           task.numRedirects += 1;
           if (task.numRedirects <= maxRedirects) {
             task.lastUrl = _nsurlToUri(nsRequest.URL!);
+            if (profile != null) {
+              final newRequest = URLRequest._(nsRequest);
+              profile.responseData.addRedirect(
+                HttpProfileRedirectData(
+                  statusCode: nsResponse.statusCode,
+                  method: newRequest.httpMethod,
+                  location: newRequest.url!.toString(),
+                ),
+              );
+            }
             nsRequestCompleter.call(nsRequest);
           } else {
             nsRequestCompleter.call(null);
--- a/pkgs/cupertino_http/lib/src/cupertino_client.dart
+++ b/pkgs/cupertino_http/lib/src/cupertino_client.dart
@@ -337,6 +337,7 @@ class CupertinoClient extends BaseClient {
       request: urlRequest,
       maxRedirects: request.followRedirects ? request.maxRedirects : 0,
       mapError: _mapError,
+      profile: profile,
     )..start();
 
     if (request case Abortable(:final abortTrigger?)) {


mertalev:
I just looked into per-task delegates and I think they'd also work for this. Any hooks the task delegate doesn't override go to the session-level delegate, so things like client certificates would still work. It'd avoid the limitations with websockets and redirect tracking, so maybe that's the way to go. I'll give it a try later.

brianquinlan:
Hi @mertalev this is very high quality but I'll need a while to review it.

> From a maintenance standpoint, I think this implementation is simpler than the delegate-based approach and might be able to replace it at some point, perhaps when the minimum iOS version is bumped to 15? It's relatively self-contained as it is, though, and should have no effect on existing users of cupertino_http.

> ...avoid the limitations with websockets and redirect tracking, so maybe that's the way to go. I'll give it a try later.

If I understand correctly, the current approach can't be made to honor max-redirects, right? But will the redirect handler attached to the session be called?

Also, I guess that you are doing something like:

 ```dart
 final sharedSession = URLSession.sessionWithConfiguration(
  ..., onFinishedDownloading: onFinished);

void foo() {
  sharedSession.downloadTaskWithRequest(...);
}  

void bar() {  // Possibly in another isolate
  final client = CupertinoClient.fromSharedSession(sharedSession);
  client.
}
```

The `startWithLegacyFallback` code path could be implemented in Dart using `dataTaskWithCompletionHandler`, right?

@liamappelbe is there a way to generate Dart bindings for Swift-only methods like these: https://developer.apple.com/documentation/foundation/urlsession/bytes(from:delegate:)?language=objc

**Comment by liamappelbe** [comment]
> @liamappelbe is there a way to generate Dart bindings for Swift-only methods like these: https://developer.apple.com/documentation/foundation/urlsession/bytes(from:delegate:)?language=objc

In theory you could run this through swiftgen, and it would take care of generating an `@objc` compatibility layer for you. But it looks like it returns a tuple, which [isn't supported yet](https://github.com/dart-lang/native/issues/1783). I think that's the only blocker to this particular method.

**Comment by mertalev** [comment]
> If I understand correctly, the current approach can't be made to honor max-redirects, right? But will the redirect handler attached to the session be called?

Yup, the `maxRedirects` / `followRedirect` settings couldn't be respected without a delegate. If the external session's delegate happens to handle redirects, this would still be respected, though. I just changed it to a task-level delegate approach that should respect those settings. As a bonus, it seems to have a modest performance improvement in a quick benchmark.

Dart delegate vs `bytes(for:)`:
```
flutter: ╔══════════════════╤════════════╤══════════╤══════════╤══════════╤══════════════╤════════╗
flutter: ║ Label            │ Payload    │  p50(µs) │  p90(µs) │  p99(µs) │ Throughput   │ Chunks ║
flutter: ╠══════════════════╪════════════╪══════════╪══════════╪══════════╪══════════════╪════════╣
flutter: ║ owned-session    │ 1.0 KB     │      487 │      792 │     1084 │     2.0 MB/s │      1 ║
flutter: ║ shared-session   │ 1.0 KB     │      490 │      763 │     2156 │     2.0 MB/s │      1 ║
flutter: ║ owned-session    │ 64.0 KB    │     1018 │     1312 │     2264 │    61.4 MB/s │      1 ║
flutter: ║ shared-session   │ 64.0 KB    │     1064 │     1400 │     2248 │    58.7 MB/s │      1 ║
flutter: ║ owned-session    │ 1.0 MB     │     5886 │     6464 │     8677 │   169.9 MB/s │      3 ║
flutter: ║ shared-session   │ 1.0 MB     │     6325 │     6809 │     8960 │   158.1 MB/s │     16 ║
flutter: ║ owned-session    │ 16.0 MB    │    88930 │    91404 │   106781 │   179.9 MB/s │    168 ║
flutter: ║ shared-session   │ 16.0 MB    │    87860 │    90202 │    98826 │   182.1 MB/s │    256 ║
```

Dart delegate vs task-level delegate:
```
flutter: ╔══════════════════╤════════════╤══════════╤══════════╤══════════╤══════════════╤════════╗
flutter: ║ Label            │ Payload    │  p50(µs) │  p90(µs) │  p99(µs) │ Throughput   │ Chunks ║
flutter: ╠══════════════════╪════════════╪══════════╪══════════╪══════════╪══════════════╪════════╣
flutter: ║ owned-session    │ 1.0 KB     │      461 │      647 │     1183 │     2.1 MB/s │      1 ║
flutter: ║ shared-session   │ 1.0 KB     │      399 │      562 │     1736 │     2.4 MB/s │      1 ║
flutter: ║ owned-session    │ 64.0 KB    │      846 │     1054 │     1695 │    73.9 MB/s │      1 ║
flutter: ║ shared-session   │ 64.0 KB    │      843 │     1051 │     1806 │    74.1 MB/s │      1 ║
flutter: ║ owned-session    │ 1.0 MB     │     5733 │     6199 │     8526 │   174.4 MB/s │      3 ║
flutter: ║ shared-session   │ 1.0 MB     │     5641 │     5979 │     8730 │   177.3 MB/s │     16 ║
flutter: ║ owned-session    │ 16.0 MB    │    85992 │    88690 │    97948 │   186.1 MB/s │    168 ║
flutter: ║ shared-session   │ 16.0 MB    │    82941 │    84355 │   102873 │   192.9 MB/s │    256 ║
flutter: ╚══════════════════╧════════════╧══════════╧══════════╧══════════╧══════════════╧════════╝
```

As for usage, I create a session in `AppDelegate.swift` at bootstrap and have each isolate (including the main isolate) call `fromSharedSession` to use it.

> The startWithLegacyFallback code path could be implemented in Dart using dataTaskWithCompletionHandler, right?

Probably yes!

mertalev:
I moved the legacy path to use Dart with `dataTaskWithCompletionHandler` and did some refactoring to simplify the code. The biggest change is to unify to using the task-level delegate approach - there's actually not much point to having the owned vs streamed paths since the new implementation has feature parity and is slightly more efficient. I can revert that if you'd rather have both versions, though.

brianquinlan:
If we were willing to limit support to iOS 15+ and macOS 12.0+, could we accomplish the shared session using a per-task delegate implemented in Dart?

I was experimenting with that approach and got stuck when I hit this issue: https://github.com/dart-lang/native/issues/3119

mertalev:
The Dart binding approach simplified the code quite a bit! Swift ended up just being extra indirection.

brianquinlan:
Hey @mertalev 

I'd really like to land this.

It seems like this would be a lot simpler if we just dropped support for iOS <15 and macOS <12.

I checked and, internally at Google, we have no products that use `package:cupertino_http` that don't require iOS 17+.

I don't know if there is a better source than [this](https://telemetrydeck.com/survey/apple/iOS/majorSystemVersions/) but it looks like iOS version <15 is used by 0.08% of iOS devices.

We could release the new version of `cupertino_http` as 3.x so that we won't break people who rely on iOS <15 support.

That do you think? If we did that then we could try to break this PR up a bit:
1. add support for `NSURLSessionTask.delegate` to cupertino_api
2. modify cupertino_client to use `NSURLSessionTask.delegate`
3. add the ability to pass in an existing `NSURLSession` to `CuperinoClient` (with some caveats about how that's sessions handlers may be invoked)

mertalev:
I'd definitely prefer dropping <iOS 15 as well - those versions aren't really relevant at this point today. Especially native apps using Swift UI tend to target iOS 16 or 17 to be able to use modern APIs, so I think iOS 15 is perfectly fine.

Also agree this PR does a lot and could be split up cleanly. I'll make the separate PRs so it's easier to review and land.

brianquinlan:
Thanks @mertalev !

mertalev:
Closing in favor of separate PRs.

Issue #1894, opened 2026-03-17
cupertino_http 2.4.0 crashes on iOS on wrapListenerBlock


orestesgaolin:
Typically user leaves the app and it crashes with the following stacktrace:

```
SIGABRT: ZTUM
  libsystem_kernel    0x1d79981dc  __pthread_kill
  libsystem_pthread   0x210fe3c18  pthread_kill
  libsystem_c         0x18eb23050  abort
  Flutter             0x10ae56c9c  dart::Assert::Fail (assert.cc:54)
  Flutter             0x10af104dc  DLRT_GetFfiCallbackMetadata (runtime_entry.cc:5144)
  0x114945d10  <redacted>
  cupertino_http      0x10982c34c  ___NativeCupertinoHttp_wrapListenerBlock_xx612k_block_invoke (native_cupertino_bindings.m:534)
  cupertino_http      0x10982c34c  ___NativeCupertinoHttp_wrapListenerBlock_xx612k_block_invoke (native_cupertino_bindings.m:534)
  cupertino_http      0x10982c758  _NativeCupertinoHttp_protocolTrampoline_xx612k (native_cupertino_bindings.m:555)
  CFNetwork           0x18809b2e8  __81-[__NSCFURLSessionDelegateWrapper dataTask:didReceiveResponse:completionHandler:]_block_invoke
  Foundation          0x18576543c  __NSINDEXSET_IS_CALLING_OUT_TO_A_BOOL_BLOCK__
  Foundation          0x18579be2c  -[NSBlockOperation main]
  Foundation          0x18579b570  __NSOPERATION_IS_INVOKING_MAIN__
  Foundation          0x18579b7f8  -[NSOperation start]
  Foundation          0x18579bbac  __NSOPERATIONQUEUE_IS_STARTING_AN_OPERATION__
  Foundation          0x18579d1c8  __NSOQSchedule_f
  libdispatch         0x18ea772b4  _dispatch_block_async_invoke2
  libdispatch         0x18ea81580  _dispatch_client_callout
  libdispatch         0x18ea6c55c  _dispatch_continuation_pop
  libdispatch         0x18ea6bbd0  _dispatch_async_redirect_invoke
  libdispatch         0x18ea79dac  _dispatch_root_queue_drain
  libdispatch         0x18ea7a548  _dispatch_worker_thread2
  libsystem_pthread   0x210fdc9cc  _pthread_wqthread

  libsystem_kernel    0x1d79981dc  __pthread_kill
[...]

PR #1910, opened 2026-04-27
feat(cupertino): throw on close with active requests

brianquinlan:
Makes https://github.com/dart-lang/http/issues/1894 more diagnosable by changing `CupertinoClient.close` to throw if there are pending requests. So long as the developer remembers to call `close` before the isolate exists, they will be warned (through the exception) that they did not wait for all of requests to complete.

Also makes it more clear that `Client.close` **must** be called.

---

<details>

</details>

### DIFF
--- a/pkgs/cupertino_http/lib/src/cupertino_client.dart
+++ b/pkgs/cupertino_http/lib/src/cupertino_client.dart
@@ -78,6 +78,16 @@ class _StreamedResponseWithUrl extends StreamedResponse
 /// }
 /// ```
 class CupertinoClient extends BaseClient {
+  /// A map of [URLSessionTask.taskIdentifier] to a [Completer] that will be
+  /// completed when the task's `onComplete` callback is invoked (i.e. the task
+  /// is complete and will no longer invoke callbacks).
+  ///
+  /// The [Completer.future] is `await`ed to ensure that there is no possibility
+  /// of further callbacks being invoked after:
+  /// - [send] throws (e.g. too many redirects).
+  /// - the response stream subscription is cancelled.
+  final _tasks = <int, Completer<void>>{};
+
   URLSession? _urlSession;
 
   CupertinoClient._(this._urlSession);
@@ -185,6 +195,9 @@ class CupertinoClient extends BaseClient {
 
   @override
   void close() {
+    if (_tasks.isNotEmpty) {
+      throw StateError('cannot close with running requests');
+    }
     _urlSession?.finishTasksAndInvalidate();
     _urlSession = null;
   }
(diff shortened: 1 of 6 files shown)


natebosch:
> Makes #1894 more diagnosable.

Could you elaborate on this a bit? How does it help with the diagnosis?

pkgs/cupertino_http/lib/src/cupertino_client.dart
```diff
@@ -349,64 +364,70 @@ class CupertinoClient extends BaseClient {
...
+        isRedirect: isRedirect,
+        headers: responseHeaders,
+      );
+    } catch (e) {
+      await _tasks[task.taskIdentifier]?.future;
+      rethrow;
+    }
   }
 
   Map<String, String> _getResponseHeaders(
```
natebosch: [optional] since the pr already touches the single usage point, consider renaming to `_responseHeaders`

https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

natebosch (approved):
I think it's OK to do this either way, but is there precedence for clients that throw if they are closed while requests are ongoing? I can't recall the behavior of our existing clients.

liamappelbe (approved):

brianquinlan:
> > Makes #1894 more diagnosable.
> 
> Could you elaborate on this a bit? How does it help with the diagnosis?

I changed the wording a bit - is it better?

brianquinlan:
> I think it's OK to do this either way, but is there precedence for clients that throw if they are closed while requests are ongoing? I can't recall the behavior of our existing clients.

`package:cronet_http` has the behavior because the underlying cronet library has this behavior - if you call `shutdown` while there pending requests, you get an exception.

natebosch:
> is it better?

Yes thanks

pkgs/cupertino_http/lib/src/cupertino_client.dart
brianquinlan: Done.

Jjagg:
This change caused an issue in our code. Reading through why this change was made I'm not sure if that was intentional. 

Essentially, we do an HTTP call for which we do not consume the response:
```dart
// Constructs a cupertino_http client on iOS when `!kDebugMode`
final client = _createHttpClient();

try {
  /* {...} Construct the request */
  // We only care about the status code, so ignore the response content.
  final StreamedResponse response = await client.send(request);

  /* {...} Status code validation */

  return const Success(null);
} catch (error, stackTrace) {
  /* {...} Error handling */
  return const Failure(null);
} finally {
  // This call fails because the response is not consumed.
  client.close();
}
```

Unfortunately, because we use the `dart:io` client for debug builds including in automated testing, we didn't catch this issue and it landed in prod.

We use the `cronet_http` client on Android and this code runs fine there.

If it is intentional that an exception is thrown in this scenario, could the changelog entry be marked as a breaking change?

brianquinlan:
Hey @Jjagg

This change was intentional - without the explicit `close`, the client may leak resources or crash.

I should have made it more clear that this was a breaking change :-(

PR #1918, opened 2026-05-14
Preserve header case in `IOClient`

### LINKED ISSUE #609: Add support for `preserveHeaderCase`

natebosch:
Closes #609

Some non-compliant servers misbehave with the default handling of
headers in the `dart:io` HTTP implementation. Well behaved servers
should not be impacted, but other use cases which require a particular
formatting can be handled by always preserving header case.

### DIFF
--- a/pkgs/http/CHANGELOG.md
+++ b/pkgs/http/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Clarified the behavior of response headers in API documentation comments.
 * Replace references to `dart:web` with `package:web` dartdoc.
+* Preserve header cases in `IOClient`.
 
 ## 1.6.0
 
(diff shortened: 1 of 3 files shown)


brianquinlan (approved):
Is it worth adding a test for this?

natebosch:
> Is it worth adding a test for this?

Can't hurt. Done.

PR #1930, opened 2026-05-29
fix(cupertino_http): Build hook code assets check

### LINKED ISSUE #1929: Build hook crashes with null check exception when building for Android
When building a Flutter application for Android that includes cupertino_http as a dependency, the native assets build step fails with an unhandled exception in the package's build.dart hook.

**Package version:** cupertino_http: 3.0.1

**Flutter version:** 3.44.0 (via FVM)

**Platform:** Building for Android (target device: Samsung SM-G955F)

**Error:**
Unhandled exception:
```
Null check operator used on a null value
#0      new CodeConfig._fromJson (package:code_assets/src/code_assets/config.dart:53:67)
#1      HookConfigCodeConfig.code (package:code_assets/src/code_assets/config.dart:22:37)
#2      main.<anonymous closure> (file:///.../cupertino_http-3.0.1/hook/build.dart:12:22)
#3      build (package:hooks/src/api/build_and_link.dart:191:18)
#4      main (file:///.../cupertino_http-3.0.1/hook/build.dart:11:9)
```

**What happened:**

The build.dart hook unconditionally accesses config.code (line 12), which internally calls CodeConfig._fromJson. On an Android build, the code assets configuration appears to be null, causing the null-assertion operator (!) to throw at config.dart:53:67.

**Expected behavior:**

Since cupertino_http is an Apple-platform library (iOS/macOS), the build hook should gracefully skip native asset compilation when the target platform is not iOS or macOS, rather than crashing with an unhandled exception.

**Reproduction steps:**

Add cupertino_http: ^3.0.1 to a Flutter project's [pubspec.yaml](vscode-file://vscode-app/c:/Users/jasper.wauters/AppData/Local/Programs/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Run flutter build apk or flutter run targeting an Android device
Observe the build failure during "Building assets for package:cupertino_http"

**Workaround:**

None found. The package cannot currently be included in a project that also targets Android, even if cupertino_http is only used in iOS/macOS code paths.

**Additional context:**

The hook should likely guard against non-Apple target platforms before attempting to resolve config.code, similar to:
```
if (config.targetOS != OS.iOS && config.targetOS != OS.macOS) {
  return; // No native assets needed on non-Apple platforms
}
```

dcharkes:
Closes: https://github.com/dart-lang/http/issues/1929

### DIFF
--- a/pkgs/cupertino_http/CHANGELOG.md
+++ b/pkgs/cupertino_http/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.0.2
+
+* Fix a [bug](https://github.com/dart-lang/http/issues/1929) where code
+  assets were built on non-code-assets hook invocations.
+
 ## 3.0.1
 
 * Fix a [bug](https://github.com/dart-lang/http/issues/1925) where native
--- a/pkgs/cupertino_http/hook/build.dart
+++ b/pkgs/cupertino_http/hook/build.dart
@@ -9,6 +9,9 @@ import 'package:native_toolchain_c/native_toolchain_c.dart';
 
 void main(List<String> args) async {
   await build(args, (input, output) async {
+    if (!input.config.buildCodeAssets) {
+      return;
+    }
     if (input.config.code.targetOS != OS.iOS &&
         input.config.code.targetOS != OS.macOS) {
       return;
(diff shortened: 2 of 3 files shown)


brianquinlan (approved):
Thanks @dcharkes

PR #1937, opened 2026-06-02
Support for configuring the browser `fetch` credentials mode in `BrowserClient`.

### LINKED ISSUE #1936: Expose the browser `fetch` credentials mode in `BrowserClient`
### Summary
`BrowserClient` currently maps its `withCredentials` option to only two browser `fetch` credentials modes:

```dart
credentials: withCredentials ? 'include' : 'same-origin',
```

This makes it possible to use:

- `include`
- `same-origin`

However, the third valid `fetch` credentials mode, `omit`, is currently not configurable through the public `BrowserClient` API.

It would be useful if `BrowserClient` allowed callers to explicitly configure the credentials mode used for browser requests.

### Use case
In some Flutter Web applications, requests must be sent without browser credentials, even for same-origin requests. This means cookies, client certificates, and HTTP authentication information should not be included by the browser.

The desired browser `fetch` configuration is:

```dart
credentials: 'omit',
```

At the moment, the only way to achieve this behavior with `package:http` appears to be forking or copying the `BrowserClient` implementation and changing the request initialization manually.

### Current behavior
When `withCredentials` is `false`, `BrowserClient` uses:

```dart
credentials: 'same-origin',
```

When `withCredentials` is `true`, `BrowserClient` uses:

```dart
credentials: 'include',
```

There is no public API to select:

```dart
credentials: 'omit',
```

### Expected behavior
`BrowserClient` should expose a way to configure all credentials modes supported by the browser `fetch` API:

- `omit`
- `same-origin`
- `include`

This would avoid the need for package forks when applications require requests to be sent without browser credentials.

### Possible API
One possible approach would be to introduce an explicit credentials mode enum:

```dart
enum BrowserCredentialsMode {
  omit,
  sameOrigin,
  include,
}
```

And allow it to be configured on `BrowserClient`:

```dart
final client = BrowserClient(
  credentialsMode: BrowserCredentialsMode.omit,
);
```

For backward compatibility, the default could remain equivalent to the current behavior:

```dart
BrowserCredentialsMode.sameOrigin
```

The existing `withCredentials` API could either be kept as a convenience option or deprecated in favor of the more explicit credentials mode.

### Why `withCredentials` is not expressive enough
The current boolean API can only represent two states:

```text
false -> same-origin
true  -> include
```

But the browser supports three distinct credentials modes:

```text
omit
same-origin
include
```

Therefore, a boolean does not fully represent the underlying browser behavior.

### Workaround
The current workaround is to fork or copy `BrowserClient` and change the `RequestInit` configuration manually:

```dart
credentials: 'omit',
```

This works technically, but it is difficult to maintain because the fork has to track upstream changes in `package:http`.

### Benefits
Adding this option would:

- expose the full browser `fetch` credentials behavior
- remove the need for forks for this use case
- make credentials handling explicit at the call site
- improve compatibility with applications that have strict cookie or authentication requirements
- keep the existing default behavior unchanged

### Proposed outcome
Please consider adding support for configuring the browser `fetch` credentials mode in `BrowserClient`, including support for `omit`.

mauriceraguseinit:
This PR adds support for configuring the browser `fetch` credentials mode in `BrowserClient`.

Currently `BrowserClient` only exposes a boolean `withCredentials` option, which maps to either `same-origin` (when `false`) or `include` (when `true`). The browser `fetch` API, however, supports a third credentials mode, `omit`, which cannot be expressed through the existing boolean API. The only way to use `omit` today is to fork or copy `BrowserClient` and change the `RequestInit` manually.

Closes #1936.

### What this changes

- Adds a new `BrowserCredentialsMode` enum with the values `omit`, `sameOrigin`, and `include`, matching the three credentials modes defined by the [[Fetch standard](https://fetch.spec.whatwg.org/#requestcredentials)](https://fetch.spec.whatwg.org/#requestcredentials).
- Adds a `credentialsMode` property and constructor parameter to `BrowserClient`, defaulting to `BrowserCredentialsMode.sameOrigin` to preserve the existing default behavior.
- The `RequestInit.credentials` value is now derived from `credentialsMode`.

### Backward compatibility

- The existing `withCredentials` property is kept and continues to work, but is now deprecated because a boolean cannot represent all three credentials modes.
- It is implemented as a compatibility adapter on top of `credentialsMode`:
  - the getter returns `true` only when `credentialsMode` is `include`;
  - the setter maps `true` to `include` and `false` to `sameOrigin`.
- Existing code such as `BrowserClient()..withCredentials = true` keeps producing the same `fetch` behavior as before.

### New usage

```dart
final client = BrowserClient(
  credentialsMode: BrowserCredentialsMode.omit,
);
```

---


pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +82,49 @@ external JSPromise<Response> _fetch(
...
+  /// matches the previous default behavior when [withCredentials] was `false`.
+  BrowserClient({
+    this.credentialsMode = BrowserCredentialsMode.sameOrigin,
+  });
+
+  /// The browser `fetch` credentials mode used for requests.
+  ///
+  /// Defaults to [BrowserCredentialsMode.sameOrigin], which matches the
+  /// previous behavior when [withCredentials] was `false`.
+  BrowserCredentialsMode credentialsMode;
```
brianquinlan: Can we make this private? I think that mutating it after construction is error-prone in any case.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -35,9 +35,39 @@ BaseClient createClient() {
...
+    ]);
+
+/// The browser `fetch` credentials mode used by [BrowserClient].
+///
+/// Controls whether the browser sends credentials such as cookies, TLS client
+/// certificates, or authorization headers with a request.
+///
+/// See also:
+/// - https://fetch.spec.whatwg.org/#requestcredentials
+enum BrowserCredentialsMode {
```
brianquinlan: Do we want to call this RequestCredentials to match the spec?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -35,9 +35,39 @@ BaseClient createClient() {
...
+  /// Never send credentials with the request and never include credentials in
+  /// the response.
+  ///
+  /// This corresponds to the browser `fetch` credentials mode `omit`.
+  omit('omit'),
+
+  /// Send credentials for same-origin requests only.
```
brianquinlan: and only include credentials in same-origin replies.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -35,9 +35,39 @@ BaseClient createClient() {
...
+  ///
+  /// This corresponds to the browser `fetch` credentials mode `same-origin`.
+  sameOrigin('same-origin'),
+
+  /// Always send credentials, even for cross-origin requests.
```
brianquinlan: ...and include them in all responses.

brianquinlan (changes requested):
It it possible to construct tests for this? Also update CHANGELOG.md.

pkgs/http/lib/src/browser_client.dart
mauriceraguseinit: I agree that mutating this after construction is error-prone. However, withCredentials was a mutable public field in the original code. Making requestCredentials final and removing or disabling the withCredentials setter would break backwards compatibility for users who configure the client after instantiation.
How do we want to deal with this?

mauriceraguseinit:
I have taken the comments regarding the implementation into account and adjusted the code.

pkgs/http/lib/src/browser_client.dart
brianquinlan: How about we make `credentialsMode` non-`final` but `private`. The `withCredentials` setter can mutate it. When `withCredentials` is removed, we can make `credentialsMode` final. WDYT?

pkgs/http/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.6.2-wip
```
brianquinlan: 1.6.1-wip (the `-wip` means Work-In-Progress) has not yet been released so you don't need to add a new version - just add your note to the 1.6.1-wip section.

pkgs/http/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Add `BrowserCredentialsMode` and `BrowserClient.credentialsMode` to support the `omit` browser fetch credentials mode. Deprecate `withCredentials`.
```
brianquinlan: 80 columns please

pkgs/http/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: http
-version: 1.6.1-wip
+version: 1.6.2-wip
```
brianquinlan: You can revert this.

pkgs/http/test/browser_client_test.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```
brianquinlan: I was imagining that these tests could actually verify the client/server interaction by running an HTTP server. But now I'm not sure if that is possible without actually server the Dart-compiled Javascript from the same host/port as the test server. If it is not possible, then I don't think that these tests are valuable and you can delete them. Sorry for asking you to do this unnecessary work.

pkgs/http/README.md
```diff
@@ -266,6 +266,9 @@ Some well-supported implementations are:
 | [`package:cronet_http`][cronethttp] — [`CronetClient`][cronetclient] | Android | Flutter | ✅︎ | ✅︎ | ― |
 | [`package:fetch_client`][fetch] — [`FetchClient`][fetchclient] | Web | Dart, Flutter | ✅︎ | ✅︎ | ✅︎ |
 
+> [!NOTE]
+> [`BrowserClient`][browserclient] allows you to configure the browser `fetch` credentials mode (e.g., `omit`, `same-origin`, `include`) by passing the `requestCredentials` parameter to its constructor.
```
brianquinlan: 80 columns

But this actually feels out of place to mention a client's feature here. Maybe add this to the dartdoc for `BrowserClient`?

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +84,50 @@ external JSPromise<Response> _fetch(
 /// Responses are streamed but requests are not. A request will only be sent
 /// once all the data is available.
 class BrowserClient extends BaseClient {
+  /// Create a [BrowserClient].
+  ///
+  /// By default, credentials are sent for same-origin requests only, which
```
brianquinlan: I would remove. Most people will not be aware of the past behavior.

pkgs/http/lib/src/browser_client.dart
mauriceraguseinit: Sounds perfect! That's exactly how I've structured it in the latest push.

The internal field is now private (_requestCredentials) and mutable via the deprecated withCredentials setter to preserve backwards compatibility. I've also exposed a public read-only getter (requestCredentials) so new users can read the mode without being able to mutate it after construction. When withCredentials is eventually removed, we can easily make the underlying field final.

I think keeping the old setter mutable for now is a great call, as some users might rely on HTTP client generators (like OpenAPI Generator) that instantiate the client first and configure properties like credentials dynamically right after. This way, we don't break their existing workflows.

mauriceraguseinit:
All items updated and cleaned up:
- Reverted README.md changes; the feature is now documented exclusively in the class/constructor Dartdoc.
- Cut down the constructor documentation to remove references to historical behavior.
- Cleaned up the changelog entry by moving the note under the existing `1.6.1-wip` section and reverted the pubspec version bump.
- Removed the unnecessary browser unit/integration tests.

Ready for another look! Thanks for the guidance, Brian.

pkgs/http/lib/src/browser_client.dart
mauriceraguseinit: Done, removed the reference to the past behavior to keep the documentation focused and clean.

pkgs/http/README.md
mauriceraguseinit: Good point. I've removed the note from the README.md entirely.

The documentation is already fully covered in the Dartdoc for BrowserClient and its constructor in browser_client.dart (keeping the 80-column limit in mind).

pkgs/http/test/browser_client_test.dart
mauriceraguseinit: No worries at all,..

I've gone ahead and removed the test file as requested. The PR should now be clean and ready.

pkgs/http/pubspec.yaml
mauriceraguseinit: ok

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +84,49 @@ external JSPromise<Response> _fetch(
...
+      : _requestCredentials = requestCredentials;
+
+  /// The internal browser `fetch` credentials mode used for requests.
+  RequestCredentials _requestCredentials;
+
+  /// The browser `fetch` credentials mode used for requests.
+  ///
+  /// Defaults to [RequestCredentials.sameOrigin], which matches the
+  /// previous behavior when [withCredentials] was `false`.
+  RequestCredentials get requestCredentials => _requestCredentials;
```
brianquinlan: I meant that this getter shouldn't exist at all - it should be purely private.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +84,49 @@ external JSPromise<Response> _fetch(
...
 class BrowserClient extends BaseClient {
+  /// Create a [BrowserClient].
+  ///
+  /// By default, credentials are sent for same-origin requests only.
+  BrowserClient(
+      {RequestCredentials requestCredentials = RequestCredentials.sameOrigin})
+      : _requestCredentials = requestCredentials;
+
+  /// The internal browser `fetch` credentials mode used for requests.
+  RequestCredentials _requestCredentials;
```
brianquinlan: Move this before the constructor, please.

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +84,49 @@ external JSPromise<Response> _fetch(
...
+
   /// Whether to send credentials such as cookies or authorization headers for
   /// cross-site requests.
   ///
   /// Defaults to `false`.
-  bool withCredentials = false;
+  ///
+  /// This property is deprecated because it can only represent two of the three
+  /// browser `fetch` credentials modes (`same-origin` and `include`). Use
+  /// [_requestCredentials] instead to also support [RequestCredentials.omit].
```
brianquinlan: Don't reference a private variable in dartdoc (I suspect that you used AI to generate this PR - you need to supervise it ;-))

pkgs/http/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 1.6.1-wip
 
+* Add `BrowserCredentialsMode` and `BrowserClient.credentialsMode` to support
```
brianquinlan:
```suggestion
* Add `BrowserCredentialsMode` and to support
```

mauriceraguseinit:
Alright, everything is sorted out! 

- Moved the `_requestCredentials` field declaration above the constructor.
- Removed the public getter entirely.
- Fixed the remaining internal field reference in the deprecated `withCredentials` getter Dartdoc to point to the constructor parameter instead.

Fingers crossed for the final review! 😉

pkgs/http/lib/src/browser_client.dart
```diff
@@ -52,11 +84,43 @@ external JSPromise<Response> _fetch(
...
+
   /// Whether to send credentials such as cookies or authorization headers for
   /// cross-site requests.
   ///
   /// Defaults to `false`.
-  bool withCredentials = false;
+  ///
+  /// This property is deprecated because it can only represent two of the three
+  /// browser `fetch` credentials modes (`same-origin` and `include`). Use
+  /// the constructor parameter instead to also support [RequestCredentials.omit].
```
brianquinlan: This line is too long, otherwise looks good!

brianquinlan (changes requested):

pkgs/http/lib/src/browser_client.dart
mauriceraguseinit: I've broken down that last line so it fits well within the limit.

brianquinlan (approved):
Thank you!

PR #1948, opened 2026-07-04
[http] Fix incorrect IOClient documentation for non-2xx HTTP responses

### LINKED ISSUE #1947: [http] IOClient doc comment incorrectly states that non-2xx responses throw ClientException
This [doc comment](https://github.com/dart-lang/http/blob/5d94ef52582867e077bf41c3fa20fb8b1d1d834e/pkgs/http/lib/src/io_client.dart#L80) suggests that `IOClient` throws `ClientException` in case of a non-2xx HTTP status code (such as 404).

However, that's not the case with [http](https://pub.dev/packages/http) and most `BaseClient` implementations including [cupertino_http](https://pub.dev/packages/cupertino_http), [ok_http](https://pub.dev/packages/ok_http) and `BrowserClient`.

Try running this (when using using top-level functions):

```dart

void main() async {
  final response = await http.get(Uri.https('dart.dev', '_404'));
  print(response.statusCode);
}
```

Or when using a `Client` instance:

```dart
import 'package:http/io_client.dart';

void main(List<String> args) async {
  final http.Client client = IOClient();

  try {
    final response = await client.get(Uri.https('dart.dev', '_404'));
    print(response.statusCode);
  } finally {
    client.close();
  }
}
```

In either case, `statusCode` is `404` and no exception was thrown (assuming `dart.dev` is reachable).

<details><summary>More details</summary>
<p>

it seems that `IOClient` throws `_ClientException` that extends `ClientException` and implements `SocketException` ([code](https://github.com/dart-lang/http/blob/6656f15e88e68f6cafa2a7bbffa37fd6ac2dd33a/pkgs/http/lib/src/io_client.dart#L22-L27)) to avoid breaking existing consumers that handle `SocketException`.

 `ClientException` hides `OSError` (from `dart:io`) and other platform-specific details from `SocketException`.

`_ClientException` implements `SocketException` in a compatible manner, allowing consumers to handle these details (e.g., when using conditional imports to support web) when needed.

[NSErrorClientException](https://github.com/dart-lang/http/blob/master/pkgs/cupertino_http/lib/src/cupertino_client.dart#L19-L20) extends `ClientException`.

In most cases, consumers of the `http` should consider handling `ClientException` unless they need internal `SocketException` details (does not work on the web). Most apps do not need to.

```dart
try {
  data = await client.read(Uri.https('example.com', ''));
} on http.ClientException catch (e) {
  // Exception is transport-related.
  //
  // If platform-specific socket details are needed, catch `SocketException`
  // before `ClientException` (or without `ClientException`) and inspect `osError` for additional information.
  //
  // If the handler for `SocketException` were removed, then all exceptions
  // would be caught by this handler.
}
```

This would keep the handling implementation independent (`BrowserClient`, `IOClient`, `CupertinoClient`, `OkHttpClient`, `WinHttpClient`, `FetchClient`, `RhttpCompatibleClient`), which works on the web as well.

Implementation-specific exceptions can still be handled when internal details are needed (e.g., `OSError`, `NSError`).

</p>
</details>

EchoEllet:
Fixes #1947

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -68,16 +68,16 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
 /// callers to get more detailed exception information for socket-level
 /// failures, if desired.
 ///
-/// For example:
+/// Example:
 /// ```dart
-/// final client = http.Client();
+/// final BaseClient client = IOClient();
 /// late String data;
 /// try {
 ///   data = await client.read(Uri.https('example.com', ''));
-/// } on SocketException catch (e) {
-///   // Exception is transport-related, check `e.osError` for more details.
 /// } on http.ClientException catch (e) {
-///   // Exception is HTTP-related (e.g. the server returned a 404 status code).
+///   // Exception is transport-related.
+///   // If platform-specific socket details are needed, catch `SocketException`
+///   // before `ClientException` (or without `ClientException`) and inspect `osError` for more details.
 ///   // If the handler for `SocketException` were removed then all exceptions
 ///   // would be caught by this handler.
 /// }
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -70,7 +70,7 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
 ///
 /// Example:
 /// ```dart
-/// final BaseClient client = IOClient();
+/// final client = IOClient();
 /// late String data;
 /// try {
 ///   data = await client.read(Uri.https('example.com', ''));


pkgs/http/lib/src/io_client.dart
```diff
@@ -67,17 +67,19 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
 /// [ClientException] that also implements [SocketException]. This allows
 /// callers to get more detailed exception information for socket-level
 /// failures, if desired.
+////
+/// See also: [_ClientSocketException]
```
brianquinlan: Can you remove this reference? It is to a private class.

pkgs/http/lib/src/io_client.dart
```diff
@@ -67,17 +67,19 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
...
 ///
-/// For example:
+/// Example:
 /// ```dart
-/// final client = http.Client();
+/// final BaseClient client = IOClient();
```
brianquinlan:
```suggestion
/// final client = IOClient();
```

pkgs/http/lib/src/io_client.dart
```diff
@@ -67,17 +67,19 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
...
 /// late String data;
 /// try {
 ///   data = await client.read(Uri.https('example.com', ''));
-/// } on SocketException catch (e) {
-///   // Exception is transport-related, check `e.osError` for more details.
 /// } on http.ClientException catch (e) {
-///   // Exception is HTTP-related (e.g. the server returned a 404 status code).
+///   // Exception is transport-related.
```
brianquinlan, line 78: The previous comment was false but I don't think that the new comment is correct either. I think that the exception could be transport-related but it could also be HTTP-related, for example a failure to parse the response message. Maybe restore the previous `SocketException` `on`-clause and change this to.

```dart
// Exception is HTTP-related (e.g. the client could not parse the server's response).
```

Maybe there is a better (or other) example.

brianquinlan (changes requested):
Thanks for doing this!

pkgs/http/lib/src/io_client.dart
EchoEllet: Sure. 

>  It is to a private class.

I added it since it may potentially help consumers understand the difference between `SocketException` and `ClientException` and why `on ClientException` is often sufficient even for `IOClient` implementation.

pkgs/http/lib/src/io_client.dart
EchoEllet: > I don't think that the new comment is correct either
> I think that the exception could be transport-related but it could also be HTTP-related

Indeed. Some `BaseClient` implementations may throw `ClientException` in the event of a TLS handshake error, whereas `IOClient` (or `HttpClient` from `dart:io`) throws `TlsException`.


> // Exception is HTTP-related (e.g. the client could not parse the server's response).

I propose:

```dart
// Exception is transport-related (e.g., no internet connection or server is unreachable)
// or HTTP protocol-related (e.g., redirect processing failure, such as a missing
// Location header).
```

Since both `io.SocketException` and `io.HttpException` (and by extension `io.RedirectException`) are mapped to `ClientException` from the `http` package.


> Maybe restore the previous SocketException on-clause and change this to.

Makes sense. However, I suggest adding a note indicating remove if not needed, since this is already covered by `http.ClientException`. Many developers assume they need to handle both when seeing the code.

Places that may throw `ClientException` in `IOClient`:

1. When sending a request after closing the client

```dart
  @override
  Future<IOStreamedResponse> send(BaseRequest request) async {
    if (_inner == null) {
      throw ClientException(
          'HTTP request failed. Client is already closed.', request.url);
    }
}
```

Maybe it should throw a Dart error (e.g., `StateError`) instead of `ClientException`, since this is typically considered a programming bug? 

When I sent a PR to flutter/packages, the preferred approach was to use `Error`s for programming bugs (e.g., https://github.com/flutter/packages/pull/8079/#discussion_r1902075152)


2. When `HttpClient` from `dart:io` throws

```dart
} on SocketException catch (error) {
      throw _ClientSocketException(error, request.url);
    } on HttpException catch (error) {
      throw ClientException(error.message, error.uri);
    }
```

I assume `HttpException` means something went wrong at the HTTP protocol level.

There is also `RedirectException` from `dart:io`:

```dart
class RedirectException implements HttpException {
  final String message;
  final List<RedirectInfo> redirects;

  const RedirectException(this.message, this.redirects);
  // ...
}
```

Which suggests that an HTTP response was at least received?
I'm not sure if `HttpException` indicates that a response was received.

In either case, this does not suggest `statusCode >= 300` (not a successful response or non-2xx).

brianquinlan:
A bit of context...

I introduced `_ClientSocketException` because `IOClient.send` was not catching `SocketException` from `dart:io` so people where writing code like in the example:

```dart
 try {
   data = await client.read(Uri.https('example.com', ''));
 } on SocketException catch (e) {
 } on http.ClientException catch (e) {
 }
```

But, that is not correct for any client other than `IOClient`. So `_ClientSocketException` was designed as a way to be backwards-compatible with existing code (since it `implements SocketException`) while allowing people to only catch `ClientException` (since it `extends ClientException`) if the wanted to support `Client`'s other than `IOClient`.

My expectation is that most applications should just catch `ClientException` and you should only catch `SocketException` if:
1. you know that you are only working with `IOClient`
2. you need to extract `osError` 

Throwing `ClientException` if the client is already closed is probably the wrong thing (that change predates me working on `package:http`) but fixing it would be breaking and it probably isn't worth it.

For

```
// Exception is transport-related (e.g., no internet connection or server is unreachable)
// or HTTP protocol-related (e.g., redirect processing failure, such as a missing
// Location header).
```

Which exception does that apply to? Internet connection issues should be `SocketException` while HTTP protocol-related issues should be `ClientException`, right?

EchoEllet:
Thank you for the context.

> just catch ClientException and you should only catch SocketException if:
>
> you know that you are only working with IOClient
> you need to extract osError

I agree.

> My expectation is that most applications should just catch

<details><summary>There is an exception</summary>

As far as I have checked, catching `ClientException` is not sufficient to cover `TlsException`.

At least in my case, I workarouned this with [a conditional import](https://github.com/LibreLabApp/librelab/blob/main/librelab_flutter/lib/common/network/tls_exception/tls_exception_web.dart#L26) (see also [this](https://github.com/LibreLabApp/librelab/blob/main/librelab_flutter/lib/common/network/tls_exception/tls_exception_io.dart#L3)) to cover TLS failures on IO implementation. Does not appear to be an issue with `BrowserClient` or `OkHttpClient` since it always throws a `ClientException` in case of a TLS failure.

<p>

</p>
</details> 

> Throwing ClientException if the client is already closed is probably the wrong thing (that change predates me working on package:http) but fixing it would be breaking, and it probably isn't worth it.

I would still consider it or add a comment indicating the reason why this is not a Dart error (e.g., `StateError`). Application developers should probably attempt to fix this programming bug rather than causing it to display an error message to the end user as if it were an HTTP protocol-related or transport error.

Many applications keep the client open for the lifetime of their application, so it may not even close.

> Which exception does that apply to? Internet connection issues should be SocketException while HTTP protocol-related issues should be ClientException, right?

Yes, but `ClientException` should cover both cases unless I'm missing anything.

brianquinlan:
We should probably catch `TlsException` and rethrow it (possible as a new exception type, in the same pattern as `_ _ClientSocketException`. Why don't you make the changes as you want and I'll review?

EchoEllet:
Sure. Maybe in a separate PR since this is only focused on clarifying the comment.

pkgs/http/lib/src/io_client.dart
```diff
@@ -68,16 +68,16 @@ class _IOStreamedResponseV2 extends IOStreamedResponse
...
-/// final client = http.Client();
+/// final client = IOClient();
 /// late String data;
 /// try {
 ///   data = await client.read(Uri.https('example.com', ''));
-/// } on SocketException catch (e) {
-///   // Exception is transport-related, check `e.osError` for more details.
 /// } on http.ClientException catch (e) {
-///   // Exception is HTTP-related (e.g. the server returned a 404 status code).
+///   // Exception is transport-related.
```
brianquinlan, line 78: Is this always transport-related? How about just remove this line.

And then wrap the next two lines to 80 columns and I'll merge!

PR #1949, opened 2026-07-05
docs(readme): add close method override in UserAgentClient example

EchoEllet:
It seems that `close()` is a no-op without an override in this case:

```dart
import 'package:http/io.dart';

// Copied from https://pub.dev/packages/http#using
class UserAgentClient(final String userAgent, final http.Client _inner) extends http.BaseClient {

  @override
  Future<http.StreamedResponse> send(http.BaseRequest request) {
    request.headers['user-agent'] = userAgent;
    return _inner.send(request);
  }
}

void main() {
  final http.BaseClient client = UserAgentClient('user-agent', IOClient());
  client.close(); // No-op. Does not call IOClient.close()
}
```

I'm not sure if I misunderstood something or am consuming the API in the wrong way.

If the intended way to close the HTTP client is:

```dart
void main() {
  final http.BaseClient innerClient = IOClient();
  
  final http.BaseClient client = UserAgentClient('user-agent', innerClient);
   // Send requests using "client"...

  innerClient.close();
}
```

Then the current example code is as intended. But even then, I would throw a Dart error when calling `UserAgentClient.close()`, to avoid silent programming bugs.

---

<details>

</details>

### DIFF
--- a/pkgs/http/README.md
+++ b/pkgs/http/README.md
@@ -80,6 +80,9 @@ class UserAgentClient extends http.BaseClient {
     request.headers['user-agent'] = userAgent;
     return _inner.send(request);
   }
+
+  @override
+  void close() => _inner.close();
 }
 ```
 ## Testing


EchoEllet:
> If the intended way to close the HTTP client is:

It is worth noting that `RetryClient` (from the `http` package) calls `_inner.close()` ([source](https://github.com/dart-lang/http/blob/master/pkgs/http/lib/retry.dart#L178-L179)). So I don't think the current example is as intended.

brianquinlan (approved):
Thank you so much!

EchoEllet:
Nit: `BrowserClient` uses `fetch()`, which automatically adds `User-Agent`. So this `UserAgentClient` example is only applicable in non-web platforms. It cannot be overridden in many browsers including Google Chrome.

So I would add an optional note indicating that this should be used only for non-browser clients.

Please let know whther it should be added.

brianquinlan:
> Nit: `BrowserClient` uses `fetch()`, which automatically adds `User-Agent`. So this `UserAgentClient` example is only applicable in non-web platforms. It cannot be overridden in many browsers including Google Chrome.
> 
> So I would add an optional note indicating that this should be used only for non-browser clients.
> 
> Please let know whther it should be added.

It's up to you. This is only meant as an example on how you can compose clients, not code that anyone should actually use. So I don't think it is important to document this aspect.

brianquinlan (approved):
Thank you!

PR #1951, opened 2026-07-10
Fix RFC 850 two-digit year parsing

### LINKED ISSUE #1939: [http_parser] parseHttpDate does not apply RFC 850 two-digit year rule
Package: `http_parser`

## Description

`parseHttpDate()` supports obsolete RFC 850 HTTP-date values, but it interprets the two-digit year as `1900 + yy` unconditionally.

That does not match the HTTP-date recipient rule for `rfc850-date`. RFC 9110 Section 5.6.7 says:

> Recipients of a timestamp value in rfc850-date format, which uses a two-digit year, MUST interpret a timestamp that appears to be more than 50 years in the future as representing the most recent year in the past that had the same last two digits.

Reference: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7

## Reproduction

With `http_parser` 4.1.2:

```dart
import 'package:http_parser/http_parser.dart';

void main() {
  print(parseHttpDate('Sunday, 21-Jun-26 07:28:00 GMT').toUtc());
}
```

## Actual behavior

The value is parsed as year 1926, because the implementation currently does:

```dart
final year = 1900 + _parseInt(scanner, 2);
```

Current source on `master`:
https://github.com/dart-lang/http/blob/master/pkgs/http_parser/lib/src/http_date.dart#L71

## Expected behavior

When parsed in 2026, `Sunday, 21-Jun-26 07:28:00 GMT` should resolve to `2026-06-21T07:28:00Z`, not 1926.

More generally, `rfc850-date` two-digit years need to be resolved using the RFC 9110 50-year rule rather than a fixed `1900 + yy` mapping.

## Impact

Downstream HTTP clients that rely on `parseHttpDate()` for fields such as `Retry-After` can treat a future RFC 850 timestamp as a past timestamp and retry immediately.

medz:
## Summary

- resolve RFC 850 two-digit years relative to the current UTC year
- apply RFC 9110's 50-year rollover rule
- add a regression test for the reported 2026 date
- update the API documentation and changelog

## Root cause

`parseHttpDate()` always added 1900 to RFC 850's two-digit year, so `26` was parsed as 1926 regardless of the current year.

## Impact

RFC 850 timestamps such as those used in `Retry-After` headers now resolve to the correct century instead of potentially being treated as expired.

Closes #1939

## Validation

- `dart format --output=none --set-exit-if-changed .`
- `dart analyze --fatal-infos`
- `dart test --platform vm`
- `dart test --platform chrome`
- `dart test --platform chrome --compiler dart2wasm`

### DIFF
--- a/pkgs/http_parser/CHANGELOG.md
+++ b/pkgs/http_parser/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 4.1.3-wip
 
+* Apply the RFC 9110 50-year rule when parsing RFC 850 dates.
 * Replace reference to `dart:web` with `package:web` in README.md.
 
 ## 4.1.2
--- a/pkgs/http_parser/CHANGELOG.md
+++ b/pkgs/http_parser/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 4.1.3-wip
 
-* Apply the RFC 9110 50-year rule when parsing RFC 850 dates.
+* **BREAKING:** Apply the RFC 9110 50-year rule when parsing RFC 850 dates.
+  Previously, two-digit years were always interpreted as years between 1900
+  and 1999.
 * Replace reference to `dart:web` with `package:web` in README.md.
 
 ## 4.1.2
--- a/pkgs/http_parser/lib/src/http_date.dart
+++ b/pkgs/http_parser/lib/src/http_date.dart
@@ -68,6 +68,11 @@ DateTime parseHttpDate(String date) =>
         scanner.expect('-');
         final month = _parseMonth(scanner);
         scanner.expect('-');
+        // RFC 9110, Section 5.6.7 says:
+        // Recipients of a timestamp value in rfc850-date format, which uses a
+        // two-digit year, MUST interpret a timestamp that appears to be more
+        // than 50 years in the future as representing the most recent year in the
+        // past that had the same last two digits.
         final currentYear = DateTime.now().toUtc().year;
         var year = currentYear - currentYear % 100 + _parseInt(scanner, 2);
         if (year > currentYear + 50) year -= 100;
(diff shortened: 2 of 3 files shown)


pkgs/http_parser/lib/src/http_date.dart
```diff
@@ -68,7 +68,9 @@ DateTime parseHttpDate(String date) =>
         scanner.expect('-');
         final month = _parseMonth(scanner);
         scanner.expect('-');
-        final year = 1900 + _parseInt(scanner, 2);
+        final currentYear = DateTime.now().toUtc().year;
```
brianquinlan, line 76: Could you add a note explaining this? Like:


```suggestion
        // RFC 9110, Section 5.6.7 says:
        // Recipients of a timestamp value in rfc850-date format, which uses a
        // two-digit year, MUST interpret a timestamp that appears to be more
        // than 50 years in the future as representing the most recent year in the
        // past that had the same last two digits.
        final currentYear = DateTime.now().toUtc().year;
```

pkgs/http_parser/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 4.1.3-wip
 
+* Apply the RFC 9110 50-year rule when parsing RFC 850 dates.
```
brianquinlan: This is technically breaking. I'm not sure if we need to call this 5.0.0 or if we can just note the break in the CHANGELOG like:

* *BREAKING*: Apply ... Previously RFC-850 dates were always interpreted as ....

medz: Thank you for your comment, I've added it.

medz: I don't think a 5.0 release is necessary because I personally believe it's a bug. Previously, the 1900 number was always added, and it seems nobody used it to parse the Retry-After header. I only discovered this issue while developing a request library that respects the retry header.

brianquinlan:
Could you fix the lint so that I can merge this?

medz:
> Could you fix the lint so that I can merge this?

fixed

brianquinlan (approved):

Issue #1827, opened 2025-09-30
Thousands of requests with no TLS versioning


codyfilatov:
I apologize if this issue is incorrectly defined. My app supports over 100k MAU and I am in a bit of a panic.

At the moment about 0.25% or so of my users are experiencing WRONG_VERSION_NUMBER, this has periodically sprung up in the past but a massive increase in reports are seen. My backend security policy supports TLS 1.0 - 1.3 on AWS Load Balancer... but I am finding that requests at the load balancer have empty TLS version numbers. In practically all cases this is happening on Wifi connections, and them changing to 5G or using a VPN suddenly it works. If they then switch back to their Wifi it fails again consistently.

**Example of not working connection log:**
2025-09-30T02:50:01.421565Z 180.75.XX.XX 60836 443 - - - "-" - - - TID_695e9abde1345a408da00999afd3f5c5

**Example of working connection log:**
2025-09-30T02:50:19.240004Z 172.56.XX.XX 31601 443 TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 0.120 "-" - - Success TID_8d3f296ac60a7f49bcee2408908f6ef4

I am not doing anything fancy here, my requests are called via _response = await get(Uri.parse(requestUrl));_ without a persistent client. The endpoint is just a standard _https://xxxxx.com/ping_

What I don't understand is, why are requests hitting my backend without a TLS version leading to the failure, and is there anything specific on the app client I need to do to prevent this from happening? Again it works fine for 99.75% of users, but the remaining just straight up see constant failure.

codyfilatov:
What we found was, for thee 0.25% of users. If we just change the subdomain name from "api.xx.io" to "fallback,xx,io" the issue is resolved for all of those users. I still don't understand why though, both subdomains point to the same CNAME place, have the same wildcard certificate, and end up on the same load balancer.

kevmoo:
Are you using this package via the VM? What version of Dart?

aphasingnirvana:
## Request to reopen: independently observed near-identical production incident

We are seeing a current production incident in a separate Flutter mobile application that closely matches this report, including the Wi-Fi-only concentration and recovery after changing the API hostname.

### Client environment

- Affected recent releases: multiple app builds, including releases built with **Flutter 3.44.1 / Dart 3.12.1**
- HTTP layer: **Dio 5.10.0 using its native IO adapter** (ultimately the Dart IO networking stack)
- Platforms: iOS/iPadOS and Android
[...]

PR #1952, opened 2026-07-10
Add no-response workflow

mosuem:
Copy no-response (stale) workflow from i18n repo to close inactive issues.


brianquinlan:
Don't we already have this? For example:
https://github.com/dart-lang/http/issues/1827

mosuem:
Right, thanks! Misconfigured grep on my side.

PR #1953, opened 2026-07-14
Switch to feature version bump

natebosch:
Use a feature bump for the new API surface area with the
`requestCredentials` optional argument to the `BrowserClient`
constructor.


brianquinlan (approved):
Oops, thanks for catching that!

PR #1954, opened 2026-07-21
doc(http_profile): fix broken code examples in header setter docs

meliharik:
The four header-setter doc comments in `package:http_profile` contain examples that don't compile, plus one wrong receiver. All verified against the package's own tests, which use the correct form.

### Wrong receiver in the response class

`http_profile_response_data.dart:98` documented `HttpProfileResponseData.headersListValues` with `profile?.requestData...` — copied from the request file. Now `responseData`.

### Setters invoked with call syntax

All four examples called the setters like methods:

```dart
profile?.requestData.headersListValues({'Foo': ['Bar', 'Baz']});
```

`headersListValues` and `headersCommaValues` are setters, so this doesn't compile. The package's tests use assignment — `http_profile_response_data_test.dart:105`:

```dart
profile.responseData.headersListValues = {
```

The examples now match.

### Bracket typos

- `{'Foo': 'Bar, Baz']});` (request:120, response:118) — a `]` closing a `{`.
- The `headers` getter docs showed the bare map as `{'Foo': ['Bar', 'Baz']});` (request:137, response:135) — a stray `);` on what the prose describes as just a map.

Each fixed snippet was checked with `dart format --output=none`: fails to parse before, parses after.

Added a bullet to the existing `0.1.1-wip` CHANGELOG section; no version bump.

### DIFF
--- a/pkgs/http_profile/CHANGELOG.md
+++ b/pkgs/http_profile/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 0.1.1-wip
 
 * Fixed unintended HTML tags in doc comments. 
+* Fixed the code examples in the `headersListValues` and `headersCommaValues`
+  doc comments.
 
 ## 0.1.0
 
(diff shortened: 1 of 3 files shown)


brianquinlan:
Thank you so much for fixing this!

brianquinlan (approved):

PR #1955, opened 2026-07-24
Migrate to the prototype visitor API

liamappelbe:
Context: https://github.com/dart-lang/native/pull/3482

I'm not planning to merge this, just sharing what it would look like under the new API.


pkgs/cupertino_http/tool/generate_code.dart
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:ffigen/ffigen.dart';
+import 'package:logging/logging.dart';
+
+class NativeCupertinoHttpVisitor extends Visitor {
+  static const includedInterfaces = {
```
dcharkes, line 9: I'd probably inline this in the one method where it is used.

pkgs/cupertino_http/tool/generate_code.dart
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  const NativeCupertinoHttpVisitor();
+
+  @override
+  void visitObjCInterface(ObjCInterface node) {
+    if (includedInterfaces.contains(node.originalName)) {
+      node.isExcluded = false;
+      for (final method in node.methods) {
+        if (node.originalName == 'NSURLResponse' &&
```
dcharkes, line 49: Maybe I'd try to make const maps with strings for the renames as well.

```dart
const methodRenames = {
  'NSURLResponse' : {
    'initWithURL:MIMEType:expectedContentLength:textEncodingName:':
       'initWithUrlAndMIMEType',
  },
  'NSHTTPURLResponse': // ...
}
```

and ditto for the param renames

That might be easier to read than the for loops.

dcharkes: Side note, we don't visit the methods and params and look up at the parents? That would avoid us having to have the for loops here. E.g. then we'd just reuse the visitParam and visitMethod that already traverse the whole ast.

PR #1956, opened 2026-07-30
feat(http2): add a pooled, multiplexed HTTP/2 http.Client

### LINKED ISSUE #1385: Add a `http.Client` compatible client
With the introduction of the cupertino_http and cronet_http only Windows and Linux Dart and Flutter apps can't use the benefits of HTTP2 while staying compatible with he abstract `http.Client` interface.
So it would be nice if this was added to this package.

demolaf:
Closes #1385

Adds `Http2Client`, a pooled, multiplexed `http.Client` backed by HTTP/2 connections, plus the `ClientPool` it's built on — fixes `dart:io`'s `HttpClient` opening one connection per concurrent request. Ported and generalized from a downstream implementation built for firebase/firebase-admin-dart#305.

- `ClientPool`: most-full-first packing, idle GC, failure retirement, and a lease API so a slot can outlive the call that claimed it
- `Http2Client`: pools per `host:port`, caps concurrent handshakes globally, multi-host safe (e.g. as `googleapis_auth`'s `baseClient`)
- Never opens more concurrent streams on a connection than the server's `SETTINGS_MAX_CONCURRENT_STREAMS` allows, and doesn't use a connection at all until that frame has arrived — see `Http2Client.settingsTimeout`
- A request holds its pool slot until the response body ends or is cancelled, so the pool's view matches the number of streams actually open
- Retries once when a pooled connection was closed by the peer before reuse
- Adds `ClientTransportConnection.peerMaxConcurrentStreams`. This is a new member on an implementable class, hence the minor version bump


pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+  /// Sends [request] as a single HTTP/2 stream on [transport], translating
+  /// between `BaseRequest`/`StreamedResponse` and http2's frames.
+  Future<StreamedResponse> _sendOverHttp2(
+    ClientTransportConnection transport,
+    BaseRequest request,
+    List<int> bodyBytes,
+  ) async {
+    if (!transport.isOpen) throw const _ConnectionClosedByPeer();
+
+    final path =
```
mosuem, line 178: AI tells me the path should default to `/` when empty.

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+            : request.url.path;
+
+    final stream = transport.makeRequest([
+      Header.ascii(':method', request.method),
+      Header.ascii(':scheme', 'https'),
+      Header.ascii(':authority', request.url.host),
+      Header.ascii(':path', path),
+      // HTTP/2 requires lowercase header names (RFC 7540 8.1.2).
+      for (final entry in request.headers.entries)
+        Header.ascii(entry.key.toLowerCase(), entry.value),
```
mosuem: Again an AI comment:

> HTTP/1.x connection-specific headers (connection, keep-alive, proxy-connection, transfer-encoding, upgrade) are forbidden in HTTP/2 and can cause strict servers to reject the stream with a  protocol error. Since Http2Client may wrap requests originating from general HTTP/1.1-oriented code, consider stripping any connection-specific headers (as well as host, since :authority is already sent).

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,136 @@
...
+  final Future<T> Function() _create;
+  final Future<void> Function(T resource) _destroy;
+  final int maxConcurrentOperations;
+  final int maxIdleResources;
+
+  final _resources = <_PooledResource<T>>[];
+  var _terminated = false;
+  Completer<void>? _drained;
+
+  /// The number of resources currently in the pool. For testing.
```
mosuem: Then mark it as testing only

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,136 @@
...
+  int get size => _resources.length;
+
+  /// The number of in-flight operations across every resource. For testing.
+  int get opCount =>
```
mosuem: Also testing only annotation then

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,136 @@
...
+    try {
+      await _destroy(await resource.future);
+    } catch (_) {
+      // Best-effort: a failure here must not shadow the caller's own
+      // request error, since this runs inside run()'s finally block.
+    }
+  }
+
+  bool get _hasExcessIdleCapacity {
+    final idleCapacity = _resources.fold(
```
mosuem:
```
final idleCapacity = _resources.fold(
  0,
  (total, resource) =>
      total + (resource.failed ? 0 : maxConcurrentOperations - resource.inFlight),
);
```

To not count failed resources, which technically have no idle capacity, into the sum

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,136 @@
...
+      0,
+      (total, resource) =>
+          total + (maxConcurrentOperations - resource.inFlight),
+    );
+    return idleCapacity > maxIdleResources * maxConcurrentOperations;
+  }
+
+  void _maybeCompleteDrain() {
+    final drained = _drained;
```
mosuem: Is this just for not having to use a `!`? Then rather use a `case final` maybe?

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /// Waits for in-flight requests to finish, then closes every connection.
+  ///
+  /// Unlike [close] (constrained by `http.Client`'s synchronous signature),
+  /// this can be awaited by callers who hold a concrete [Http2Client].
+  Future<void> terminate() async {
+    for (final pool in _pools.values) {
+      await pool.terminate();
+    }
```
mosuem, line 313: AI: 

```
After awaiting pool.terminate() for all entries in _pools, consider calling _pools.clear();. If someone calls client.terminate() and subsequently sends a new request, clearing _pools allows _poolFor() to spin up a fresh pool rather than throwing a StateError('This pool has already been  terminated.') from the old pool. Alternatively, if Http2Client is intended to be permanently unusable after termination, adding a _closed = true flag at the top of send() would give a clearer error message.
```

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+          );
+        }
+        if (!bodyController.isClosed) bodyController.close();
+      },
+      onError: (Object error, StackTrace stackTrace) {
+        if (!statusCompleter.isCompleted) {
+          statusCompleter.completeError(error, stackTrace);
+        }
+        bodyController.addError(error, stackTrace);
+        if (!bodyController.isClosed) bodyController.close();
```
mosuem: Nit: Always use parentheses with `if`s.

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+      bodyController.stream,
+      statusCode,
+      headers: responseHeaders,
+      request: request,
+    );
+  }
+
+  /// The number of connections currently pooled, across every host.
+  int get connectionCount =>
+      _pools.values.fold(0, (total, pool) => total + pool.size);
```
mosuem: Nit: use `.sum` from https://pub.dev/documentation/collection/latest/collection/IterableIntegerExtension/sum.html here and elsewhere.

pkgs/http2/test/http2_client_test.dart
```diff
@@ -0,0 +1,215 @@
...
+
+SecurityContext _serverContext() =>
+    SecurityContext()
+      ..useCertificateChain('test/certificates/server_chain.pem')
+      ..usePrivateKey('test/certificates/server_key.pem', password: 'dartdart');
+
+Future<MultiProtocolHttpServer> _bind() =>
+    MultiProtocolHttpServer.bind('localhost', 0, _serverContext());
+
+// test/certificates/server_chain.pem is long expired and self-signed, same
```
mosuem: Hm, then we should probably regenerate it?

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,218 @@
...
+/// token exchange) targets different hosts than the API calls that follow.
+/// A connection dialed for one host is never reused for another.
+///
+/// `onBadCertificate` is forwarded as-is to `SecureSocket.connect`: returning
+/// `true` accepts a certificate that failed normal verification (expired,
+/// self-signed, wrong host, ...). It exists for tests and trusted private
+/// networks - do not use it to accept arbitrary certificates in production.
+class Http2Client extends BaseClient {
+  Http2Client({
+    this.maxStreamsPerConnection = 100,
```
brianquinlan, line 39: Should this default to `SETTINGS_MAX_CONCURRENT_STREAMS` send by the server?

brianquinlan (commented):
I'll have more review feedback tomorrow. Also, this PR adds (preliminary) conformance tests: https://github.com/dart-lang/http/pull/1960

You might want to add them to this PR.

pkgs/http2/lib/src/http2_client.dart
demolaf: yes, this should read `SETTINGS_MAX_CONCURRENT_STREAMS` returned from the server. i'll make that change now

pkgs/http2/test/http2_client_test.dart
demolaf: no the comment was misleading, I've removed that now — same bypass at https://github.com/dart-lang/http/blob/master/pkgs/http2/test/multiprotocol_server_test.dart#L63

pkgs/http2/lib/src/http2_client.dart
demolaf: `_settingsHandler.peerSettings.maxConcurrentStreams` isn't exposed in the public API as of now. So I'll need to add a getter e.g. `peerMaxConcurrentStreams` on `ClientTransportConnection` which can then be used in `Http2Client`.

demolaf: Kept the getter on `ClientConnection` rather than `ClientTransportConnection`

demolaf:
@brianquinlan @mosuem I'm deferring the TODOs in the conformance tests to a later PR since this is already quite a lot, sounds good?

pkgs/http2/example/main.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```
brianquinlan, line 1: Let's make this example more useful for the user. Maybe copy the example from `package:cupertino_http` or something.

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,421 @@
...
+/// for example as the `baseClient` passed to `googleapis_auth`'s client
+/// helpers, whose credential negotiation (OAuth2 token endpoint, WIF/OIDC
+/// token exchange) targets different hosts than the API calls that follow.
+/// A connection dialed for one host is never reused for another.
+///
+/// `onBadCertificate` is forwarded as-is to `SecureSocket.connect`: returning
+/// `true` accepts a certificate that failed normal verification (expired,
+/// self-signed, wrong host, ...). It exists for tests and trusted private
+/// networks - do not use it to accept arbitrary certificates in production.
+class Http2Client extends BaseClient {
```
brianquinlan, line 37: Maybe marks this experimental for now:
https://api.flutter.dev/flutter/meta/experimental-constant.html

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,421 @@
...
+    if (request.url.scheme != 'https') {
+      throw ClientException(
+        'Http2Client only supports https (got "${request.url.scheme}").',
+        request.url,
+      );
+    }
+
+    List<int>? bodyBytes;
+
+    Future<StreamedResponse> attempt() async {
```
brianquinlan, line 298: This is awkward to read. Do we need to define this function, which is only used once?

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,240 @@
...
+/// A pool of resources of type [T].
+///
+/// Packs load onto the most-full resource under its capacity (rather than
+/// spreading evenly across resources), opens a new resource once existing
+/// ones are full, stops routing new work to a resource once an operation on
+/// it throws, and garbage-collects idle resources past [maxIdleResources].
+///
+/// Capacity is [maxConcurrentOperations], lowered to whatever limit a
+/// resource reports for itself via `concurrencyLimitOf`.
+class ClientPool<T> {
```
brianquinlan: Would it be more clear if these were not generic? Isn't the type always going to be `ClientConnection` for all of these `T` arguments.

Oh, did you do this to make the tests easier to write?

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,240 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
brianquinlan, line 4: Maybe provide an example on how to use this library. But I like the approach.

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+
+  // Synchronous (no `await`), so concurrent calls can't race each other
+  // into both creating a resource before either sees the other's.
+  _PooledResource<T> _acquire() {
+    _PooledResource<T>? selected;
+    for (final resource in _resources) {
+      if (resource.failed) continue;
+      if (resource.inFlight < _capacityOf(resource) &&
```
brianquinlan:
```suggestion
      // Pick the available connection with the most inflight requests. This makes it more likely that
      // fewer active connections need to be maintained.
      if (resource.inFlight < _capacityOf(resource) &&
```

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,240 @@
...
+
+import 'dart:async';
+import 'dart:math';
+
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+class _PooledResource<T> {
+  _PooledResource(this.future);
+  final Future<T> future;
```
brianquinlan: I think that these need to be better named/documented:
`future` refers to the creation of the resource, right?
`inFlight` refers to the number of concurrent requests of the resource, right?
`failed` indicates that creating the resource failed, right?

Maybe:
`create`
`inFlightCount`
`createFailed`

And a comment? 

This would be more clear if `T` were `ClientConnection` but maybe the tests would then be too hard to write.

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,240 @@
...
+  /// drains. Prefer [run] unless the slot has to outlive the future that
+  /// produced whatever the caller is returning.
+  Future<PoolLease<T>> acquire() async {
+    if (_terminated) {
+      throw StateError('This pool has already been terminated.');
+    }
+
+    while (true) {
+      final pooled = _acquire();
+      pooled.inFlight++;
```
brianquinlan: Do you need to decrement this if you call `_release` without using it?

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,421 @@
...
+
+  /// Waits for in-flight requests to finish, then closes every connection.
+  ///
+  /// Unlike [close] (constrained by `http.Client`'s synchronous signature),
+  /// this can be awaited by callers who hold a concrete [Http2Client].
+  ///
+  /// A request counts as in-flight until its response body ends or is
+  /// cancelled, so a caller holding a response it never reads will hold this
+  /// up. [close] does not await this, so it can never block on that.
+  Future<void> terminate() async {
```
brianquinlan: Can we just make this part of `close` for now? I think that we need a general solution for async close?

pkgs/http2/lib/src/connection.dart
```diff
@@ -513,6 +513,16 @@ class ClientConnection extends Connection implements ClientTransportConnection {
   bool get isOpen =>
       !_state.isFinishing && !_state.isTerminated && _streams.canOpenStream;
 
+  /// The maximum number of concurrent streams the peer currently allows, per
+  /// its most recent SETTINGS_MAX_CONCURRENT_STREAMS (RFC 7540 6.5.2), or
+  /// `null` if it hasn't advertised a limit.
+  ///
+  /// Deliberately not on [ClientTransportConnection]: that class can only be
```
brianquinlan: @mosuem maybe we should add it there anyway - there are no "implements ClientTransportConnection"  on GitHub but we could also bump semver.

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,421 @@
...
+    );
+    if (socket.selectedProtocol != 'h2') {
+      socket.destroy();
+      throw StateError(
+        'Server did not negotiate HTTP/2 (got ${socket.selectedProtocol})',
+      );
+    }
+
+    final died = Completer<void>();
+    final incoming = socket.transform(
```
brianquinlan, line 122: Could you add a comment explaining why we need this?

pkgs/http2/lib/src/connection.dart
mosuem: Yes, just add it there and bump semver.

pkgs/http2/example/main.dart
demolaf: okay, i'll look into this

pkgs/http2/lib/src/http2_client.dart
demolaf: that makes sense to do since we still have more pending feature work to do per the conformance tests

pkgs/http2/lib/src/http2_client.dart
demolaf: yeah i agree, it's actually used twice, the initial attempt and then in the `.catchError` block.

pkgs/http2/lib/src/client_pool.dart
demolaf: yes that's correct, we could also just use `MockClientConnection` through mockito.

pkgs/http2/lib/src/client_pool.dart
demolaf: yeah i agree it's not clear — it's not a double decrement, `_release()` does that itself so it pairs with the inFlight++ above. i'll rework it.

pkgs/http2/lib/src/connection.dart
demolaf: done — moved it to `ClientTransportConnection` and bumped to `3.1.0-wip`

pkgs/http2/example/main.dart
demolaf: I added an example showing 100 concurrent requests but it's not really a great example since `http/1.1` can handle that faster (if uncapped). The only notable difference would be memory usage since `http/1.1 `would open TCP+TLS connections per request.

pkgs/http2/lib/src/client_pool.dart
demolaf: done, renamed all three. also dropped the generic — `ClientPool` is `ClientConnection`-specific now, and the pool tests use a generated `MockClientConnection`.

pkgs/http2/lib/src/http2_client.dart
demolaf: I've reverted back to `ClientTransportConnection. peerMaxConcurrentStreams` and a version bump.

pkgs/http2/example/main.dart
brianquinlan: Why don't you just show a simple usage example like the other clients (I'm working on harmonizing them now: https://github.com/dart-lang/http/pull/1961):

```dart
// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file

import 'dart:convert' as convert;

import 'package:http2/client.dart';

void main(List<String> arguments) async {
  // This example uses the pub.dev API to fetch details about
  // package:cupertino_http.
  // https://pub.dev/help/api
  final client = Http2Client();
  final url = Uri.https('pub.dev', '/api/packages/http2/score');

  // Await the http get response, then decode the json-formatted response.
  final response = await client.get(url);
  if (response.statusCode == 200) {
    var jsonResponse =
        convert.jsonDecode(response.body) as Map<String, dynamic>;
    var likes = jsonResponse['likeCount'];
    var downloads = jsonResponse['downloadCount30Days'];
    print('Information about package:cupertino_http:');
    print('- Likes: $likes');
    print('- 30-day downloads: $downloads');
  } else {
    print('Request failed with status: ${response.statusCode}.');
  }
  client.close();
}
```

pkgs/http2/lib/src/client_pool.dart
```diff
@@ -0,0 +1,296 @@
...
+/// carry:
+///
+/// ```dart
+/// final pool = ClientPool(
+///   () => dial(host, port),
+///   maxConcurrentStreams: 100,
+/// );
+/// ```
+///
+/// Use [ClientPool.run] when the work finishes with the call:
```
brianquinlan: What do you mean by "when the work finishes`?

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,434 @@
...
+/// A pooled, multiplexed `http.Client` backed by HTTP/2 connections.
+///
+/// Every request is sent as its own HTTP/2 stream on a shared connection -
+/// see [ClientPool] - dialed per `host:port` via
+/// `SecureSocket.connect(..., supportedProtocols: ['h2'])`. Once a connection
+/// is carrying as many concurrent streams as it may - the lower of
+/// [maxStreamsPerConnection] and the server's own advertised limit - a new
+/// connection is dialed rather than queuing behind the existing one.
+///
+/// Being multi-host makes this safe to use as a general-purpose transport -
```
brianquinlan: I think that you can delete this. I actually think that you should add a note that this client does not fall back to HTTP 1.1 if the server does not support HTTP 2.

pkgs/http2/lib/src/http2_client.dart
brianquinlan: Did you add the annotation?

pkgs/http2/lib/src/http2_client.dart
```diff
@@ -0,0 +1,438 @@
...
+import 'package:pool/pool.dart';
+
+import '../transport.dart';
+import 'client_pool.dart';
+import 'connection.dart';
+
+/// A pooled, multiplexed `http.Client` backed by HTTP/2 connections.
+///
+/// Every request is sent as its own HTTP/2 stream on a shared connection -
+/// see [ClientPool] - dialed per `host:port` via
```
brianquinlan: Isn't [ClientPool] an implementation detail?

pkgs/http2/lib/src/client_pool.dart
demolaf: `run` is for when the connection is free again as soon as the returned future completes, `acquire` for when it's still in use after that.

pkgs/http2/README.md
```diff
@@ -53,3 +53,29 @@ Future<void> main() async {
...
+  print(response.body);
+  client.close();
+}
+```
+
+A connection is dialed per `host:port` as needed, so a single `Http2Client`
+is safe to reuse across requests to different hosts. Note that it speaks only
+HTTP/2 and does not fall back to HTTP/1.1: a server that does not negotiate
+`h2` is treated as an error. See the example
+[here](example/pooled_client.dart).
```
brianquinlan: Could you rename the example "main.dart" and update this link? Otherwise I think that we are good to go!

brianquinlan (approved):

PR #1958, opened 2026-08-02
fix(http2): bound peer-controlled streams and headers

DubaiChewyCookie:
SETTINGS_MAX_CONCURRENT_STREAMS was only consulted when this endpoint created local streams. A peer could ignore the advertised setting and open unbounded streams, allocating stream state, queues, windows, and controllers.

Enforce the configured peer-initiated stream limit locally before creating Http2StreamImpl. Track active peer streams and locally bound reserved PUSH_PROMISE streams as well. Reject excess streams with RST_STREAM REFUSED_STREAM without publishing them through incomingStreams or peerPushes.

Terminate the connection with GOAWAY ENHANCE_YOUR_CALM after the configured number of consecutive violations. Disabled server push is also enforced on receipt instead of relying on peer compliance or SETTINGS acknowledgement.

Inbound HEADERS and CONTINUATION fragments were accumulated without a size or time bound. Each continuation copied the preceding block again, making fragmented field-block assembly quadratic. HPACK also retained the complete decoded header list before application-level limits could run.

Add the following secure defaults per connection:

- 100 peer-initiated streams
- 16 KiB compressed field block
- 8 KiB decoded field section
- 16 CONTINUATION frames per field block
- 10 second absolute field-block timeout
- 8 consecutive post-acknowledgement stream-limit violations

Start the field-block timer when the initial HEADERS or PUSH_PROMISE frame header is parsed, before its payload is complete. Retain fragments as chunks and combine them once, enforcing compressed-size and CONTINUATION-count limits while the block is received.

Enforce the decoded field-section limit during HPACK processing using name length + value length + 32 bytes per field. Once the limit is crossed, release retained application-visible headers and continue decoding in discard mode so dynamic-table updates remain synchronized.

Reject a completely decoded oversized field section with RST_STREAM ENHANCE_YOUR_CALM and do not publish its headers. Compressed-size, CONTINUATION-count, and timeout violations terminate the connection with GOAWAY ENHANCE_YOUR_CALM because the shared HPACK context cannot safely continue without complete decompression. Invalid HPACK encoding terminates the connection with COMPRESSION_ERROR.

Advertise SETTINGS_MAX_HEADER_LIST_SIZE while enforcing the configured limit locally because the setting is advisory. This also resolves the unbounded buffering TODO introduced by fdaeafb.

Tests cover ACK-independent stream admission, pre-ACK rejection without abuse counting, post-ACK repeated abuse, exact N and N+1 limits, active and reserved stream recovery, disabled push, repeated abuse, compressed and decoded size boundaries, many continuations, slow initial payloads, Huffman and indexed amplification, HPACK dynamic-table synchronization, HEADERS, trailers, and PUSH_PROMISE.

RFC 9113 sections 4.3, 5.1.2, 6.5.2, and 10.5:
https://www.rfc-editor.org/rfc/rfc9113.html

CERT/CC VU#421644:
https://kb.cert.org/vuls/id/421644

### DIFF
--- a/pkgs/http2/lib/src/frames/frame_defragmenter.dart
+++ b/pkgs/http2/lib/src/frames/frame_defragmenter.dart
@@ -2,94 +2,161 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import '../sync_errors.dart';
+import 'dart:typed_data';
 
+import '../sync_errors.dart';
 import 'frames.dart';
 
-/// Class used for defragmenting [HeadersFrame]s and [PushPromiseFrame]s.
-// TODO: Somehow emit an error if too many continuation frames have been sent
-// (since we're buffering all of them).
+/// Defragments field blocks from [HeadersFrame]s and [PushPromiseFrame]s with
+/// bounded buffering.
 class FrameDefragmenter {
-  /// The current incomplete [HeadersFrame] fragment.
-  HeadersFrame? _headersFrame;
+  final int? maxHeaderBlockSize;
+  final int? maxContinuationFrames;
 
-  /// The current incomplete [PushPromiseFrame] fragment.
-  PushPromiseFrame? _pushPromiseFrame;
+  Frame? _initialFrame;
+  BytesBuilder? _fragments;
+  int _fieldBlockSize = 0;
+  int _combinedPayloadLength = 0;
+  int _continuationFrames = 0;
+
+  FrameDefragmenter({this.maxHeaderBlockSize, this.maxContinuationFrames}) {
+    if (maxHeaderBlockSize case final limit? when limit < 0) {
+      throw ArgumentError.value(limit, 'maxHeaderBlockSize', 'must be >= 0');
+    }
+    if (maxContinuationFrames case final limit? when limit < 0) {
+      throw ArgumentError.value(limit, 'maxContinuationFrames', 'must be >= 0');
+    }
+  }
+
+  /// Whether an incomplete field block is currently buffered.
+  bool get isDefragmenting => _initialFrame != null;
 
   /// Tries to defragment [frame].
   ///
-  /// If the given [frame] is a [HeadersFrame] or a [PushPromiseFrame] which
-  /// needs de-fragmentation, it will be saved and `null` will be returned.
-  ///
-  /// If there is currently an incomplete [HeadersFrame] or [PushPromiseFrame]
-  /// saved, [frame] needs to be a [ContinuationFrame]. It will be added to the
-  /// saved frame. In case the defragmentation is complete, the defragmented
-  /// [HeadersFrame] or [PushPromiseFrame] will be returned.
-  ///
-  /// All other [Frame] types will be returned.
-  // TODO: Consider handling continuation frames without preceding
-  // headers/push-promise frame here instead of the call site?
+  /// Incomplete field blocks return `null`. A completed field block is returned
+  /// as one [HeadersFrame] or [PushPromiseFrame]. Fragment bytes are retained
+  /// as chunks and combined only once, avoiding repeated whole-block copies.
   Frame? tryDefragmentFrame(Frame? frame) {
-    if (_headersFrame != null) {
-      if (frame is ContinuationFrame) {
-        if (_headersFrame!.header.streamId != frame.header.streamId) {
-          throw ProtocolException(
-            'Defragmentation: frames have different stream ids.',
-          );
-        }
-        _headersFrame = _headersFrame!.addBlockContinuation(frame);
-
-        if (frame.hasEndHeadersFlag) {
-          var frame = _headersFrame;
-          _headersFrame = null;
-          return frame;
-        } else {
-          return null;
-        }
-      } else {
-        throw ProtocolException(
+    final initialFrame = _initialFrame;
+    if (initialFrame != null) {
+      if (frame is! ContinuationFrame) {
+        _failProtocol(
           'Defragmentation: Incomplete frame must be followed by '
           'continuation frame.',
         );
       }
-    } else if (_pushPromiseFrame != null) {
-      if (frame is ContinuationFrame) {
-        if (_pushPromiseFrame!.header.streamId != frame.header.streamId) {
-          throw ProtocolException(
-            'Defragmentation: frames have different stream ids.',
-          );
-        }
-        _pushPromiseFrame = _pushPromiseFrame!.addBlockContinuation(frame);
+      if (initialFrame.header.streamId != frame.header.streamId) {
+        _failProtocol('Defragmentation: frames have different stream ids.');
+      }
 
-        if (frame.hasEndHeadersFlag) {
-          var frame = _pushPromiseFrame;
-          _pushPromiseFrame = null;
-          return frame;
-        } else {
-          return null;
-        }
-      } else {
-        throw ProtocolException(
-          'Defragmentation: Incomplete frame must be followed by '
-          'continuation frame.',
+      _continuationFrames++;
+      final continuationLimit = maxContinuationFrames;
+      if (continuationLimit != null &&
+          _continuationFrames > continuationLimit) {
+        _failResourceLimit(
+          'Inbound field block used $_continuationFrames CONTINUATION frames, '
+          'exceeding the limit of $continuationLimit.',
         );
       }
-    } else {
-      if (frame is HeadersFrame) {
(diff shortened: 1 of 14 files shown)


mosuem:
Can you link an issue this fixes, or provide context on this PR?

DubaiChewyCookie:
> Can you link an issue this fixes, or provide context on this PR?

There is no public issue to link because I originally reported these findings privately through the channel specified by the [Dart security policy](https://dart.dev/security#reporting-vulnerabilities), using [g.co/vulnz](https://g.co/vulnz). The team reviewing the report confirmed that I could proceed with a public PR directly.

The PR addresses two related per-connection resource-exhaustion issues in `package:http2`:

1. A remote peer can open more peer-initiated streams than the configured `SETTINGS_MAX_CONCURRENT_STREAMS` limit, causing unbounded allocation of stream and application state.

   **Risk tags:** `remote DoS` · `unbounded stream state` · `per-stream memory amplification` · `application-work amplification` · `unenforced peer limits`

2. Inbound HEADERS/CONTINUATION blocks have no aggregate size, fragment-count, or receive-time bounds. Their assembly is quadratic, and HPACK decoding can retain a decoded field section much larger than its compressed input.

   **Risk tags:** `remote DoS` · `unbounded field-block buffering` · `quadratic fragment assembly` · `HPACK memory amplification` · `indexed-header amplification`

Both issues are remotely triggerable before application-level limits can reliably reject the work. They are included in the same PR because the fixes share the connection-level admission and header-processing paths.

mosuem:
Thanks, I'll take a look!

pkgs/http2/lib/src/hpack/hpack.dart
```diff
@@ -121,6 +149,22 @@ class HPackDecoder {
 
     try {
       var headers = <Header>[];
+      var headerListSize = 0;
+      var headerListSizeExceeded = false;
+
+      void processHeader(Header header) {
```
mosuem, line 155: Nit:
```
      void processHeader(Header header) {
        headerListSize += header.name.length + header.value.length + 32;
        if (headerListSizeExceeded) return;
        if (maxHeaderListSize != null && headerListSize > maxHeaderListSize) {
          headers.clear();
          headerListSizeExceeded = true;
        } else {
          headers.add(header);
        }
      }
```

pkgs/http2/lib/transport.dart
```diff
@@ -13,22 +13,95 @@ export 'src/hpack/hpack.dart' show Header;
 
 typedef ActiveStateHandler = void Function(bool isActive);
 
+/// Default maximum number of peer-initiated streams per connection.
```
mosuem, line 16: It would be great if the doc comments for these default could contain the rationale behind choosing that value - something like a link or quote from the spec, or links to other reference implementations.

pkgs/http2/lib/src/frames/frame_defragmenter.dart
```diff
@@ -2,94 +2,160 @@
...
-  ///
-  /// If there is currently an incomplete [HeadersFrame] or [PushPromiseFrame]
-  /// saved, [frame] needs to be a [ContinuationFrame]. It will be added to the
-  /// saved frame. In case the defragmentation is complete, the defragmented
-  /// [HeadersFrame] or [PushPromiseFrame] will be returned.
-  ///
-  /// All other [Frame] types will be returned.
-  // TODO: Consider handling continuation frames without preceding
-  // headers/push-promise frame here instead of the call site?
+  /// Incomplete field blocks return `null`. A completed field block is returned
```
mosuem, line 36: I like the previous code references instead of, for example, `PUSH_PROMISE`.

mosuem (changes requested):

pkgs/http2/lib/transport.dart
DubaiChewyCookie: I also compared the proposed defaults against several major HTTP/2 stacks to sanity-check the values and trade-offs.

| Limit | [Netty 4.1](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java) | [Jetty 12.1](https://github.com/jetty/jetty.project/blob/jetty-12.1.x/jetty-core/jetty-http2/jetty-http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java) | [Apache HttpCore 5](https://github.com/apache/httpcomponents-core/blob/master/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/config/H2Config.java) | [Go `x/net/http2`](https://github.com/golang/net/blob/master/http2/server.go) | [Node.js / nghttp2](https://nodejs.org/api/http2.html) | Decision |
|---|---:|---:|---:|---:|---:|---|
| Peer-initiated concurrent streams | 100 | 128 server-side | 250 | 250 server-side | Local SETTINGS default is `2^32-1` | 100 |
| Decoded header-list size | 8 KiB | 8 KiB request headers | 16,777,215 bytes | About 1 MiB server-side; 10 MiB client-side | 65,535 bytes, plus a default 128-field limit | 8 KiB |
| Compressed header-block size | About 10 KiB before GOAWAY | About 8 KiB server parser capacity | No separate fixed limit identified | No fixed aggregate limit; fragments are checked against the remaining decoded budget | No separate public inbound byte limit | 16 KiB |
| Absolute field-block timeout | No dedicated field-block default identified | No dedicated field-block default identified | No dedicated field-block default identified | No dedicated field-block default identified | No dedicated field-block default identified | 10 seconds |
| CONTINUATION frames per block | 16 small, non-final fragments | No dedicated count identified | 100 | No fixed count | 8 in [nghttp2](https://github.com/nghttp2/nghttp2/blob/master/lib/includes/nghttp2/nghttp2.h) | 16 |
| Consecutive stream-limit violations before GOAWAY | No directly equivalent consecutive threshold identified | No directly equivalent threshold identified | No directly equivalent threshold identified | Excess streams are refused; no equivalent cumulative threshold identified | Similar `maxSessionRejectedStreams` default of 100 | 8, but count only after SETTINGS ACK |

A few details are important when reading the table:

- These values are not always directly equivalent. Some are advertised HTTP/2 settings, while others are local memory, parser, rate, or abuse limits.
- Node's `peerMaxConcurrentStreams = 100` is not equivalent to the inbound limit here. It acts as a pre-SETTINGS assumption about the remote endpoint's advertised limit and therefore applies in the opposite direction.
- Netty's limit of 16 counts only small, non-final CONTINUATION fragments. The Dart limit counts every CONTINUATION frame, regardless of size.
- “No dedicated limit identified” does not mean no protection. Those implementations may rely on general read or idle timeouts, parser byte limits, rate controls, or session memory budgets.
- Go deliberately uses substantially different server and client header-list budgets.
- The Jetty compressed-block figure reflects the request-header limit supplied to its HTTP/2 server parser rather than a separately named compressed-block setting.
- For TypeScript applications, Node's native `node:http2` implementation and its nghttp2 dependency are the relevant protocol stack.

pkgs/http2/lib/src/hpack/hpack.dart
DubaiChewyCookie: Applied the suggested simplification.

pkgs/http2/lib/transport.dart
DubaiChewyCookie: I updated the doc comments with the rationale for all six defaults and the relevant RFC/CERT references.

pkgs/http2/lib/src/frames/frame_defragmenter.dart
DubaiChewyCookie: Restored in both the class-level and method-level documentation.

pkgs/http2/lib/transport.dart
mosuem: Thanks!

mosuem:
LGTM - Any comments @brianquinlan ?

brianquinlan:
I think that this partially overlaps the merged https://github.com/dart-lang/http/pull/1965
It also partially overlaps the proposed https://github.com/dart-lang/http/pull/1978 

@DubaiChewyCookie could you take a look at those two PRs and update your PR to make sure that you capture their functionality.

@mosuem I'm not an expert in this area and am nervous about reviewing code ;-)

DubaiChewyCookie:
Currently updated and rebased.

brianquinlan (approved):

brianquinlan:
@mosuem this looks OK to me, shall we merge?

mosuem (approved):

PR #1959, opened 2026-08-03
Add ecosystem groundskeeper workflow

mosuem:
Adds the ecosystem groundskeeper workflow to run dart format and dart fix across packages in the repository.


brianquinlan (approved):
Does this work with flutter packages? Also do you need "dart format" to be using the "stable" SDK version in all workflows for this to make sense? I think that package:http2 is using "dev".

PR #1962, opened 2026-08-05
Undeprecate the withCredentials field

natebosch:
Work around some restrictive CI by holding off on the deprecation for
now.


natebosch:
Filed https://github.com/dart-lang/sdk/issues/63975 so we can avoid this in the future, but without a replacement feedback mechanism outside of hard-breaking usage there might be pushback

brianquinlan (approved):
Could you add a (possibly vague) note on why the deprecation is commented out long with a link to the issue?

PR #1974, opened 2026-08-12
Fix #1972: bypass body-reading for HEAD requests in io_client

### LINKED ISSUE #1972: HEAD call results in ClientException: Content-Length is larger than actual response length.
It looks like the HTTP implementation is receiving the Content-Length and then performing its usual body-length consistency check without correctly accounting for the fact that the request was HEAD.

The HEAD call itself is fine and gives an 200 OK, but trying to do anything around a try/catch setup gets messed up by this unexpected ClientException.

amanmaurya92:
Fixes #1972

This PR resolves an issue where `HEAD` requests unexpectedly throw a `ClientException` due to a mismatch between the `Content-Length` header and the actual response body length (which is naturally empty for `HEAD` requests).

Because `dart:io`'s `HttpClient` will sometimes fail when verifying the `Content-Length` of an empty `HEAD` response body, we now intercept `HEAD` requests in `io_client.dart`. Instead of attempting to read the body stream, we asynchronously drain the underlying connection (safely swallowing any spurious `dart:io` length-mismatch exceptions) and immediately return the headers with a safe, empty stream. 

---

<details>

</details>

### DIFF
--- a/pkgs/http/lib/src/io_client.dart
+++ b/pkgs/http/lib/src/io_client.dart
@@ -157,6 +157,34 @@ class IOClient extends BaseClient {
       final response = await stream.pipe(ioRequest) as HttpClientResponse;
       hasResponse = true;
 
+      if (request.method.toUpperCase() == 'HEAD') {
+        unawaited(response.drain<void>().catchError((_) {}));
+
+        var headers = <String, String>{};
+        response.headers.forEach((key, values) {
+          // TODO: Remove trimRight() when
+          // https://github.com/dart-lang/sdk/issues/53005 is resolved and the
+          // package:http SDK constraint requires that version or later.
+          headers[key] = values.map((value) => value.trimRight()).join(',');
+        });
+
+        return _IOStreamedResponseV2(
+          const Stream<List<int>>.empty(),
+          response.statusCode,
+          contentLength:
+              response.contentLength == -1 ? null : response.contentLength,
+          request: request,
+          headers: headers,
+          isRedirect: response.isRedirect,
+          url: response.redirects.isNotEmpty
+              ? response.redirects.last.location
+              : request.url,
+          persistentConnection: response.persistentConnection,
+          reasonPhrase: response.reasonPhrase,
+          inner: response,
+        );
+      }
+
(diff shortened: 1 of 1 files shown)


amanmaurya92:
While making this change, I also noticed a `TODO` in `io_client.dart` regarding `dart-lang/sdk#53005` (removing the `.trimRight()` call when parsing headers). 

Since that SDK issue appears to be resolved, we could probably clean this up now. Should I make a separate PR to refactor and remove the `.trimRight()` workaround, or is someone already planning to tackle that alongside a minimum SDK version bump?

Keetz:
I have no experience with the Dart Lang codebase, but out of curiosity I checked your code (I created the issue).
Why not find the place where the content length and response length check is made, and omit it from HEAD calls, instead of this approach of "swallowing" the exception. What if it is another exception? That doesn't seem to be handled correctly, it would just ignore any exceptions here right?

brianquinlan:
I added a comment to https://github.com/dart-lang/http/issues/1972 and https://github.com/dart-lang/http/pull/1979 attempts to reproduce the problem without success.

Do you have a better reproduction?

amanmaurya92:
Hi @brianquinlan,

Thanks for looking into this and creating #1979.

I do not have a reproduction either. I tested this against `master` (both with your conformance test from #1979 and with a local test server returning `Content-Length: 100` with an empty body on a `HEAD` request), and `IOClient` handles it without throwing any `ClientException` (since `dart:io`'s HTTP parser already sets `isHead = true` and `_noMessageBody = true`).

This PR was opened under the assumption from #1972 that `IOClient` was failing on HEAD responses with a `Content-Length` header, but it appears `IOClient` is not the culprit here. The issue reporter (@Keetz) might be using a different client (e.g. browser/Fetch, a proxy, or a third-party wrapper), or experiencing an edge case with a specific server behavior.

Given that, I'm happy to close this PR for now until @Keetz provides a minimal reproduction and specifies their client and environment on #1972. If it turns out to affect a different client in this repo, we can address it there.

brianquinlan:
Thanks @amanmaurya92 !

PR #1980, opened 2026-09-01
[cronet_http] Migrate to JNIgen v1 and use the Dart config API

liamappelbe:
This is ready for review, but I won't try to land it until JNIgen v1 publishes.


brianquinlan (approved):
So long as you remove the `git` reference from `pubspec.yaml` before merging.

